PR #1, opened 2015-02-25
Expose a /collections/ endpoint to deal with collections.

ametaireau:
@leplatrem r?

### DIFF
--- a/config/kinto.ini
+++ b/config/kinto.ini
@@ -10,7 +10,7 @@ cliquet.retry_after = 30
 cliquet.eos =
 
 pyramid.debug_notfound = true
-# cliquet.basic_auth_backdoor = true
+cliquet.basic_auth_backdoor = true
 # cliquet.backoff = 10
 cliquet.userid_hmac_secret = b4c96a8692291d88fe5a97dd91846eb4
 # cliquet.batch_max_requests = 25
--- a/config/kinto.ini
+++ b/config/kinto.ini
@@ -10,7 +10,7 @@ cliquet.retry_after = 30
 cliquet.eos =
 
 pyramid.debug_notfound = true
-cliquet.basic_auth_backdoor = true
+# cliquet.basic_auth_backdoor = true
 # cliquet.backoff = 10
 cliquet.userid_hmac_secret = b4c96a8692291d88fe5a97dd91846eb4
 # cliquet.batch_max_requests = 25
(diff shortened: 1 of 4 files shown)


kinto/tests/test_views_collections.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_individual_collections_can_be_deleted(self):
+        self.app.post('/collections/barley/records',
+                      headers=self.headers)
+
+        self.app.delete('/collections/barley/records',
+                        headers=self.headers)
+
+    def test_items_can_be_added_to_collections(self):
+        response = self.app.post('/collections/barley/records',
```
ametaireau: I'm updating the tests to use a non-empty record. 

I'm actually not sure we should support empty records, but I don't see any good reason to not support them.

config/kinto.ini
```diff
@@ -10,7 +10,7 @@ cliquet.retry_after = 30
 cliquet.eos =
 
 pyramid.debug_notfound = true
-# cliquet.basic_auth_backdoor = true
+cliquet.basic_auth_backdoor = true
```
leplatrem: commit this ?

Natim: I think we should rename it to `basic_auth_activated` since it is not really a backdoor anymore but an Auth feature.

kinto/views/collection.py
```diff
@@ -0,0 +1,10 @@
+from cliquet.resource import crud, BaseResource
+
+
+@crud(path="/collections/{collection_id}/records/{id}",
+      collection_path="/collections/{collection_id}/records")
+class Collection(BaseResource):
+
+    @property
+    def name(self):
+        return self.request.matchdict['collection_id']
```
leplatrem: Hard to believe : this is cloud storage :)

config/kinto.ini
leplatrem: yes, ok, seen the tests

leplatrem: @Natim : fully agree!

ametaireau: Actually I should remove it from here and put it only in the tests.

ametaireau: Created https://github.com/mozilla-services/cliquet/issues/31

kinto/views/collection.py
ametaireau: Yeah, cliquet rocks!

We will need a bit more code to handle collections management, plus currently it seems I can only add empty items to the collections (not very handy ;))

leplatrem: Ah yes, that's because of https://github.com/mozilla-services/cliquet/issues/30

collections management is definitely not a requirement :)

ametaireau: Neat, I know what's next in line then :)

PR #10, opened 2015-03-02
Remove loadtest target

ametaireau:
@Natim r?


ametaireau:
Fixed already. Closing.

PR #11, opened 2015-03-03
added Docker files

tarekziade:


docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+web:
+  build: .
+  links:
+   - redis
+  ports:
+   - "8000:8000"
+redis:
+  image: redis
```
leplatrem: that makes Redis available on localhost:6379 ? No change needed in app conf ?

tarekziade: Yes that's the default port

ametaireau:
Ah okay!

Cool that we have that. We've changed to use postgres as a default to store everything. So if we want to have docker use postgres, we probably want to pull postgres in addition to redis and create the default table.

ametaireau:
(I've rebased this PR on top of master)

PR #13, opened 2015-03-04
Add option to enable **FULL** database flushing

leplatrem:


kinto/tests/support.py
```diff
@@ -28,3 +30,9 @@ def get_app_settings(self):
             'fxa-oauth.oauth_uri': 'https://oauth-stable.dev.lcip.org',
             'fxa-oauth.scope': 'profile'
         }
+
+
+def get_user_headers(user):
```
ametaireau: This could even go into cliquet actually.

kinto/tests/test_views_flush.py
```diff
@@ -0,0 +1,35 @@
...
+        headers = self.headers.copy()
+        headers.update(**get_user_headers('bob'))
+        self.app.post('/collections/barley/records',
+                      headers=headers, status=201)
+        headers.update(**get_user_headers('alice'))
+        self.app.post('/collections/chocolate/records',
+                      headers=headers, status=201)
+
+    def test_returns_405_if_not_enabled_in_configuration(self):
+        self.app.post('/__flush__', headers=self.headers, status=405)
```
ametaireau: Can we also assert that the configuration has the right value?

kinto/views/flush.py
```diff
@@ -0,0 +1,19 @@
+from cliquet.utils import native_value
+from cornice import Service
```
ametaireau: from cliquet import Service?

kinto/views/flush.py
```diff
@@ -0,0 +1,19 @@
...
+
+flush = Service(name='flush',
+                description='Clear database content',
+                path='/__flush__')
+@flush.post(permission=NO_PERMISSION_REQUIRED)
+def flush_post(request):
+    settings = request.registry.settings
+    flush_enabled = settings.get('kinto.flush_endpoint_enabled', False)
+
+    if not native_value(flush_enabled):
```
ametaireau: Which makes me think we should use konfig in our projects, it's already doing this work for us.

ametaireau:
We need to add documentation on this as well.

PR #20, opened 2015-03-12
[do not merge] redis dev environment

tarekziade:
This provides a full redis environment for development purpose.

### DIFF
--- a/config/kinto-dev.ini
+++ b/config/kinto-dev.ini
@@ -1,0 +1,53 @@
+[app:main]
+use = egg:kinto
+
+cliquet.project_name = cloud storage
+cliquet.project_docs = https://kinto.rtfd.org/
+cliquet.session_backend = cliquet.session.redis
+cliquet.storage_backend = cliquet.storage.redis
+cliquet.http_scheme = http
+cliquet.retry_after = 30
+cliquet.eos =
+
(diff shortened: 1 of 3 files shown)


tarekziade:
Turns out there are deeper problems with this.

We need to:
1- allow running and launching tests in cliquet w/o postgres.
2- allow running and launching tests in kinto w/o postgres.

change the Tox file so it tries both environement. Then offer the two configs like in my initial PR

ametaireau:
To run the tests, you should be able to run postgres. I think this needs to stay as a requirement since it's the default database we're using.

config/kinto-dev.ini
```diff
@@ -0,0 +1,53 @@
+[app:main]
```
ametaireau: We should name this kinto-redis

ametaireau:
IMO the default env for devs is to use postgres, since it's the target database.

Installing it with docker is a matter of minutes. In case we really want to go down this path, then agree, we should change this, but I don't think that's the path we want to take.

leplatrem:
https://github.com/mozilla-services/cliquet/issues/124

Natim:
For kinto the default backends are postgresql.
The documentation explains how this can be changed.

PR #27, opened 2015-04-01
Upgrade cliquet and add init command to make serve

leplatrem:

PR #28, opened 2015-04-24
Import code from RL

leplatrem:


leplatrem:
Dropped in favor for #27

PR #36, opened 2015-05-05
Add documentation about Buckets and permissions

Natim:


docs/buckets.rst
```diff
@@ -0,0 +1,62 @@
+Buckets
+#######
+
+Buckets are a way to group users around some data.
```
ametaireau: This definition doesn't say much. "Buckets are a group of collections, which can be shared accross users" seems more close to what I have in mind.

docs/buckets.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Basically has got an and a list of principals that can administrate
+it.
```
ametaireau: Sorry, what?

ametaireau: I believe we should have the same set of permissions we have for other objects (read, create, update, delete)

docs/buckets.rst
```diff
@@ -0,0 +1,62 @@
...
+A bucket can be created using a PUT on the bucket URI:
+
+.. code-block:: http
+
+   > PUT /buckets/servicedenuages HTTP/1.1
+   < 201 Created
+
+    {
+      "id": "servicedenuages",
+      "principals": ["email:natim@example.com"]
```
ametaireau: Having "principals" returned here doesn't make a lot of sense to me, we should probably have a list of permissions and eventually a group "owners"

docs/buckets.rst
```diff
@@ -0,0 +1,62 @@
...
+
+ - collections
+ - groups
+
+
+Collections
+===========
+
+Creating a collection inside a bucket enable all buckets principals to
+administrate the collection.
```
ametaireau: I feel this should be handled by more fine grained permissions

docs/buckets.rst
```diff
@@ -0,0 +1,62 @@
...
+    {
+      "id": "servicedenuages",
+      "principals": ["email:natim@example.com"]
+    }
+
+
+There are two kinds of data linked to a bucket:
+
+ - collections
+ - groups
```
leplatrem: I believe it makes more sense to talk about "roles" instead of "groups"

docs/buckets.rst
```diff
@@ -0,0 +1,62 @@
...
+
+
+Collections
+===========
+
+Creating a collection inside a bucket enable all buckets principals to
```
leplatrem: Principals is weird word. We probably don't want to use it.

docs/buckets.rst
Natim: I don't think we want fined permissions on the bucket itself, do you have a usecase in mind that needs it?

docs/buckets.rst
Natim: Well having Daybed experience in mind I don't think it helps :) groups carries less meaning than roles. Is ``musicians`` a role?

ametaireau: I don't understand what a role is, but I do with a group. Why are you suggesting we could use "role"?

docs/buckets.rst
```diff
@@ -0,0 +1,99 @@
+Buckets
+#######
+
+Buckets are a group of collections, shared between users, with fined
```
ametaireau: fine-grained

docs/buckets.rst
```diff
@@ -0,0 +1,99 @@
...
+permissions on the data stored inside.
+
+Basically a bucket have got an id and a list of people identifier that
+can administrate it.
```
ametaireau: I still believe we shouldn't go with this solution. We should have permissions associated.

docs/buckets.rst
```diff
@@ -0,0 +1,99 @@
...
+
+ - collections
+ - groups
+
+
+Collections
+===========
+
+Creating a collection inside a bucket enable all buckets owners to
+administrate the collection.
```
ametaireau: administrate doesn't convey a lot of meaning. "To have all permissions on the collection items" seems a better description to me.

docs/buckets.rst
```diff
@@ -0,0 +1,99 @@
...
+           ^
+           |
+           |
+    +---------------+
+    | Records       |
+    +---------------+
+    |  - id         |
+    |  - data       |
+    |  - members    |
+    +---------------+
```
ametaireau: I agree with the schema!

docs/buckets.rst
Natim: A role is a category of people that can do specific actions (administrators, moderators, employees, managers, etc.)

docs/buckets.rst
Natim: Yes also since I didn't identify any permissions for bucket it does not means a lot of thing to me. Please elaborate on your proposal with permissions. I did not get what you wanted to do.

Natim: Does: ``To have all permissions on all bucket collections and bucket collections items.`` works for you?

ametaireau: +1

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+
+As soon as there is a need to let people work on the same records set
+or to work with the same collection, the solution is to create a :ref:`bucket <buckets>`.
+
+:ref:`Buckets <buckets>` gives a public unique name to the collection.
+
+The data is not anymore linked to a specific user that only has access
+to her private data but is linked to the bucket and managed by
+bucket's owners (those who have the ``write_bucket`` permission on the
+bucket.
```
ametaireau: From this description, I cannot tell how a bucket is different from a collection. My understanding (not when reading this text) is that buckets are a container of collections, but this text kinda says it's *like* a collection, but open for others.

So, what's the right definition? :)

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+
+**Requires authentication**
+
+This endpoint creates a new bucket with a generated unique id.
+
+By default the user id is used as the only owner.
+
+**Optional parameters**
+
+- ``permissions``: A mapping object that defines the list of user for
```
ametaireau: the list of principals.

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+  each permission
+ - ``write_bucket``: The list of users principals that have
```
ametaireau: `admins` finally looks like a better option then (yes, I know).

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+   administration permissions on the bucket, the creator is
+   automatically added to the owner list.
+ - ``create_groups``: Permission to create new groups
```
ametaireau: I think only admins should be allowed to do that.

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+
+
+Permissions
+===========
+
+They are two kind of permissions on an object:
+
+- **read**: It means that the given user or group of users have
+  got read only access to the object
+- **write**: It means that the given user or group of users have
```
ametaireau: I would actually differenciate between create and update (which seems to be both in the "write" option here).

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+Collections
+===========
+
+A collection have got a few permissions:
+
+- **read_collection**: It is a read access to the collection
+  attributes (schema and permissions)
+- **write_collection**: It is a write access to the collection
+  attributes (schema and permissions)
+- **create_records**: It is a permission that let one create new records
```
ametaireau: oh ok it's actually here.

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
+Working with buckets
+====================
+
+:ref:`Buckets <buckets>` enable the creation of collections handled by
+a group of users.
+
+As soon as there is a need to let people work on the same records set
+or to work with the same collection, the solution is to create a :ref:`bucket <buckets>`.
+
+:ref:`Buckets <buckets>` gives a public unique name to the collection.
```
leplatrem: With the bucket name as prefix, the collection name becomes unique.  (*?*)

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+
+
+PUT /buckets/<bucket_id>
+''''''''''''''''''''''''
+
+**Requires authentication**
+
+This endpoint creates a new bucket with a chosen id.
+
+If the id is already taken, then a ``409 Conflict`` http error will be returned.
```
leplatrem: Why not replace ?

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+-----------------
+
+PATCH /buckets/<bucket_id>
+''''''''''''''''''''''''''
+
+**Requires authentication**
+
+This endpoint lets you update an existing bucket.
+
+If you are not owner of the bucket you will get a ``403 Forbidden`` http error.
```
leplatrem: error response.

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+        "id": "servicedenuages",
+        "permissions": {
+            "write_bucket": ["uid:basicauth_5d127220922673e346c0ebee46c23e6739dfa756",
+                             "email:alexis@example.com"],
+            "create_groups": ["email:natim@example.com"],
+            "create_collections": [],
+        }
+    }
+
+The PATCH endpoint let you add or remove users principals from
```
leplatrem: allows to add or remove

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+permissions sets. In case you want to override the set, you can use
+the PUT endpoint.
```
leplatrem: PUT can be used to override the set.

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
+Working with buckets
+====================
+
+:ref:`Buckets <buckets>` enable the creation of collections handled by
+a group of users.
+
+As soon as there is a need to let people work on the same records set
```
leplatrem: same set of records

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+or to work with the same collection, the solution is to create a :ref:`bucket <buckets>`.
+
+:ref:`Buckets <buckets>` gives a public unique name to the collection.
+
+The data is not anymore linked to a specific user that only has access
```
leplatrem: anymore?

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+to her private data but is linked to the bucket and managed by
+bucket's owners (those who have the ``write_bucket`` permission on the
```
leplatrem: Here I would just say the *write permission* in plain english

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+bucket.
+
+Access to buckets, groups, collections and records are granted using
```
leplatrem: first apparition of *groups* ?

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+
+
+Creating a new bucket
+---------------------
+
+``/buckets/<bucket_id>``
+
+This endpoint defines the bucket resource.
+
+* Handle the owner list as well as the buckets permissions
```
leplatrem:
```
This endpoint defines the bucket resource:
* its owner
* its permissions
```

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+
+**Optional parameters**
+
+- ``permissions``: A mapping object that defines the list of user for
+  each permission
+ - ``write_bucket``: The list of users principals that have
+   administration permissions on the bucket, the creator is
+   automatically added to the owner list.
+ - ``create_groups``: Permission to create new groups
+ - ``create_collections``: Permission to create new collections
```
leplatrem: What if we only keep ``write`` and ``read`` ?

If in the future we have specific use cases, we'll extend them! For example, ``collections:write``, ``groups:write``... considering ``write`` means all of them for retro-compat'...

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+
+The PATCH endpoint let you add or remove users principals from
+permissions sets. In case you want to override the set, you can use
+the PUT endpoint.
+
+You can use ``+principal`` to add one and ``-principal`` to remove one.
+
+
+Getting bucket informations
+---------------------------
```
leplatrem: information

docs/api/buckets.rst
```diff
@@ -0,0 +1,199 @@
...
+
+PATCH /buckets/<bucket_id>
```
leplatrem: Typo GET

docs/buckets.rst
```diff
@@ -0,0 +1,104 @@
...
+
+.. _buckets:
+
+Buckets are a group of collections, shared between users, with fined-grained
+permissions on the data stored inside.
+
+Basically a bucket have got an id and a list of people identifier that
+can administrate it.
+
+A bucket can be created using a PUT on the bucket URI:
```
leplatrem: endpoint

docs/buckets.rst
```diff
@@ -0,0 +1,104 @@
...
+
+
+Collections
+===========
+
+Creating a collection inside a bucket enable all buckets users with
+the ``write_bucket`` permission to have all permissions on all bucket
+collections and bucket collections records.
+
+The collection is not linked to a user anymore but to the bucket.
```
leplatrem: anymore ?

docs/buckets.rst
```diff
@@ -0,0 +1,104 @@
...
+There are two kinds of data linked to a bucket:
+
+ - collections
+ - groups
+
+
+Collections
+===========
+
+Creating a collection inside a bucket enable all buckets users with
```
leplatrem: enables

docs/buckets.rst
```diff
@@ -0,0 +1,104 @@
...
+.. code-block:: http
+
+    > PUT /buckets/servicedenuages/collections/mushrooms HTTP/1.1
+    < 201 Created
+
+
+Groups
+======
+
+Creating a group inside a bucket ease user permission management.
```
leplatrem: eases

docs/buckets.rst
```diff
@@ -0,0 +1,104 @@
...
+    { "members": ["email:alexis@example.com"] }
+
+    < 201 Created
+
+    {
+      "id": "seekers",
+      "members": ["email:alexis@example.com"]
+    }
+
+It is now possible to use the ``groups:seekers`` principal to describe
```
leplatrem: ``group:seekers`` ← singular

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
+Permissions
+###########
+
+.. _permissions:
+
+Objects
+=======
+
+There are four kinds of objects that you can have rights on:
```
leplatrem: *rights* ? → permissions

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+- **Buckets**
+- **Groups**
+- **Collections**
+- **Records**
+
+
+Permissions
+===========
+
+They are two kind of permissions on an object:
```
leplatrem: kinds

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+A **read** access let the user read all the attributes of the object.
+
+A **write** access let the user read, update and delete any attributes
+of the object.
+
+
+Buckets
+=======
+
+Permissions to create a bucket are defined in the kinto configuration.
```
leplatrem: ``*Kinto*``

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+By default **Authenticated users** can create one.
```
leplatrem: any authenticated user can create buckets.

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+        "permissions": {
+            "read_collection": ["Authenticated"]
+        }
+    }
+
+Records access
+''''''''''''''
+
+Finally to give user and sellingapp access to the records they need,
+we will add the following permissions on each records:
```
leplatrem: record

leplatrem:
Some feedback :)

Documentation and flow

* There is a lot of vocabulary
* We need to always use the same words in order to make it easier to read
  * owners, administrators, creators, bucket users → admins
  * users, principals, user identifiers → principals (?)
* In the example part, it is hard to follow because it is not clear where those JSON are located (bucket, collection, records ?)
* We need a list of principals "types" somewhere (e.g. ``email:``, ``uid:``, ``scope:``, ...)

Features

* We should expose the problem we had in a few lines somewhere.
* Maybe the notion of group should not be mixed up with the rest, and should be introduced as a "second-line feature"

Design

* I would definitely would reduce everything to ``read`` and ``write``
```js
"permissions": {
   "groups": {
      "write": ["uid:<user>"]
   },
   "collections": {
      "read": [ ... ]
   }
}
```
* I would suggest to get rid of ``Authenticated`` and ``Everyone`` and replace them with a specific principal type  (e.g. ``status:all``, ``status:authenticated``)

docs/buckets.rst
```diff
@@ -0,0 +1,104 @@
...
+permissions inside the ``servicedenuages`` bucket.
+
+
+Schema
+======
+
+To understand objects imbrication and properties we have the following:
```
leplatrem: we have → consider the following diagram

docs/api/buckets.rst
ametaireau: We're currently changing all the layout of the docs and doing another pass on this proposal. The idea is to have a default bucket for each user.

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+  attributes (schema and permissions)
+- **create_records**: It is a permission that let one create new records
+- **read_records**: It is a read access to any collection record
+- **write_records**: It is a permission that let one update any collection record
+
+
+Records
+=======
+
+A record have got two permissions:
```
leplatrem: has got.

Where is it stored ?

docs/api/buckets.rst
ametaireau: It could be possible to replace, but only if the have the permission to do so ;)

docs/permissions.rst
```diff
@@ -0,0 +1,407 @@
...
+
+A collection have got a few permissions:
+
+- **read_collection**: It is a read access to the collection
+  attributes (schema and permissions)
+- **write_collection**: It is a write access to the collection
+  attributes (schema and permissions)
+- **create_records**: It is a permission that let one create new records
+- **read_records**: It is a read access to any collection record
+- **write_records**: It is a permission that let one update any collection record
```
leplatrem: Where is it stored ?

docs/api/buckets.rst
ametaireau: We want to introduce the concept of principals, and a group is just one principal.

docs/api/buckets.rst
ametaireau: We probably want to have all the combination of object / permission, and probably remove the ones which don't make sense. This would be more future proof than having write / read and then not being able to be more fine grained than that.

docs/buckets.rst
ametaireau: Well, it's the group of "seekers".

ametaireau:
> ```js
"permissions": {
   "groups": {
      "write": ["uid:<user>"]
   },
   "collections": {
      "read": [ ... ]
   }
}
```

I agree with that but we should add the "create" permission.

File: docs/api/buckets.rst (hunk 18 above)
Comment (leplatrem) [inline docs/api/buckets.rst]: But, should it be 409 when permission to replace is denied ?

File: docs/buckets.rst (hunk 34 above)
Comment (leplatrem) [inline docs/buckets.rst]: yes, ``group`` is singular

**Comment by leplatrem** [comment]
Ok then, let's go for Create/Read/Update/Delete permissions. But we must find relevant use-cases for them.

We could also add : 
* Private todo list
* Poll (create,read/update/delete own records)

**Comment by ametaireau** [comment]
I've updated. We probably should upate the APIs as well, but I think it's wiser to wait for feedback first.

File: docs/buckets.rst [hunk 43]
```diff
@@ -0,0 +1,141 @@
+Buckets
+#######
+
+.. _buckets:
+
+:ref:`Buckets <buckets>` enables the creation of collections handled by
```
leplatrem: enable

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+a group of ref:`User identifiers <user-identifiers>`.
+
+All collections are created in a bucket. By default, it uses the connected
+user's bucket (bucket only the connected user has access to).
```
leplatrem: By default, the bucket of the current user is used.

(the second part between parenthesis could not be processed by my brain at this time of the day)

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+
+:ref:`Buckets <buckets>` can be seen as namespaces: you can have different
+collections using the same name, but stored in different buckets, so their
```
leplatrem: collections can have the same name across different buckets

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+names don't collide.
+
+Data (everything stored in a bucket: collections, groups and records) is
```
leplatrem: ``groups`` appear out of the blue here

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+not anymore linked to a specific user that only has access to her private data
```
leplatrem: ~~anymore~~

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+but is linked to the bucket and managed by bucket's owners (those who have the
+``write_bucket`` permission on the bucket.
```
leplatrem: Do mention that notion of permission here. It comes after.

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+=================
+
+A bucket can be created using a PUT on the desired bucket URI, optionally
+specifying the list of attached permissions.
+
+Arguments:
+
+- ``permissions``: A mapping object that defines the list of users for each of
+  the following permissions defined in the table below.  In any case (even if
+  not specified), the current logged-in user will get access to the bucket.
```
leplatrem: The part after *In any case* is not clear at all. It should be put above, after the *optionnally specifying...*.

docs/api/buckets.rst
Natim: Good point, it should be 403

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Here is the list of possible permissions on a bucket:
+
++------------------------+---------------------------------+
+| Permission             | Description                     |
++========================+=================================+
+| ``write_bucket``       | The list of users principals    |
```
leplatrem: user principals come out of the blue here

docs/api/buckets.rst
Natim: Actually there is no owner list anymore.

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+|                        | that have administration        |
```
leplatrem: administration permissions  come out of the blue here (it was mentioned as *owner* before)

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+    $ http PUT http://localhost:8000/v1/buckets/{bucket_id} --auth "admin:"
+
+    POST /v1/buckets{bucket_id} HTTP/1.1
+    Authorization: Basic YWRtaW46
+
+    HTTP/1.1 201 Created
+    Content-Type: application/json; charset=UTF-8
+
+    {
+        "id": "{bucket_id}",
```
leplatrem: here those braces bring confusion. Better put some random UUID

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+        "permissions": {
+            "write_bucket": ["uid:basicauth_5d127220922673e346c0ebee46c23e6739dfa756"],
+            "create_groups": [],
+            "create_collections": [],
+        }
```
leplatrem: I would have prefered something like that : 

```

        "permissions": {
            "admins": ["uid:basicauth_5d127220922673e346c0ebee46c23e6739dfa756"],
            "groups": {
                "create": []
            },
            "collections": {
                "create": []
            },
        }

```

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+There are two kinds of data linked to a bucket: **collections** and **groups**.
```
leplatrem: This is useless here, better put it above in the introduction

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+
+
+Collections
+===========
+
+Creating a collection inside a bucket enables all buckets users with
+the ``write_bucket`` permission to have all permissions on all bucket's
+collections and associated records.
```
leplatrem: Creating a collection inside a bucket enables [..] on all buckets collections ? → not clear :)

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+
+.. code-block:: http
+
+    > PUT /buckets/servicedenuages/collections/mushrooms HTTP/1.1
+    < 201 Created
```
leplatrem: what is this ?

docs/buckets.rst
```diff
@@ -0,0 +1,141 @@
...
+    { "members": ["email:alexis@example.com"] }
+
+    < 201 Created
+
+    {
+      "id": "moderators",
+      "members": ["email:alexis@example.com"]
+    }
+
+It is now possible to use the ``groups:moderators`` principal to describe
```
leplatrem: ``group:moderators`` ?

docs/permissions.rst
```diff
@@ -0,0 +1,431 @@
...
+Users shouldn't be able to write receipts themselves, sellers and users should
+only be able to read their owns.
+
+In this case, the ``payments`` bucket will be created with the following ACLs:
+
+.. code-block:: json
+
+    {
+        "id": "payments",
+        "acls": {
```
leplatrem: acls is not suitable IMO :)

``permissions`` is a lot better, and consistent with buckets

docs/permissions.rst
```diff
@@ -0,0 +1,431 @@
...
+
+.. code-block::
+
+    {
+        "id": "<record_id>",
+        "data": {"records": "data"},
+        "permissions": {
+            "read_record": ["userid:<buyer-id>", "appid:<seller-appid>"]
+        }
+    }
```
leplatrem: Same here, I would have liked this layout:

```
"permissions": {
   "records": {
       "read": ["userid:<buyer-id>", "appid:<seller-appid>"]
   }
}
```

docs/permissions.rst
```diff
@@ -0,0 +1,431 @@
...
+.. code-block:: json
+
+    {
+        "id": "<blogname>",
+        "acls": {
+            "write_bucket": ["email:mathieu@example.com", "email:alexis@example.com"]
+        }
+    }
+
+Moderators are special persons with special rights. Rather than adding
```
leplatrem: rights --> permissions

docs/permissions.rst
```diff
@@ -0,0 +1,431 @@
...
+
+
+The Twitter use case
+--------------------
+
+- Each collection is isolated from the others, and only one person have all
+  permissions on all records.
+- Anybody can read everything.
+
+A "microblog" bucket is created, where new groups can be created by
```
leplatrem: A ``twitter`` bucket is created

docs/buckets.rst
Natim: Ben oui mon gars tu viens de créer le group moderators la ligne au dessus,

leplatrem: ``group:`` singular.

Natim: Ok yes good point :)

ametaireau:
OK, I think we're good with this, should we merge?

Natim:
Is the close an answer to the question?

ametaireau:
Sorry for the context miss here. This branch has been merged in another temporary one which contains all work related to permissions and buckets. see https://github.com/mozilla-services/kinto/pull/46

Issue #31, opened 2015-04-30
Schemas validation


leplatrem:
Let's start a discussion about schema validation !

I believe we want something similar to *Daybed*. Except that we won't
reinvent the wheel with custom schema formalism.

It looks like there is no way around JSON schema. We won't rewrite thousands
of lines of code using Colander (see https://github.com/Julian/jsonschema).
We will use existing validators, and frontends will build forms from schemas.

By default, collections could be schemaless. 

```html

PUT|PATCH|DELETE /collections/person/schema
{
    {
        "title": "Example Schema",
        "type": "object",
        "properties": {
            "firstName": {
                "type": "string"
            },
            "lastName": {
                "type": "string"
            },
            "age": {
                "description": "Age in years",
                "type": "integer",
                "minimum": 0
            }
        },
        "required": ["firstName", "lastName"]
    }
}

```

It is acceptable if the validation of the definition is done using Cornice/Colander.
Even though there are [meta-schemas](http://json-schema.org/documentation.html) :)

If a schema exists for a collection, records that don't match the schema are
refused with ``400 Bad Request`` (on PUT and after merge of PATCH).

Unlike Daybed, I propose that each user owns the schema of the collection.
Especially because the schema endpoint will probably be a resource :)
Schemas could be cached to avoid overhead of reading it from storage at each incoming record.

This implies that users can store heterogeneous data with the same collection name.

Hence, on first start, the JS application will have to check that the user did
not set any schema (e.g. at ``/collections/moz:readinglist:articles/schema``).
If she did, then confirm to replace it. If she hacks it afterwards and stores
invalid records, the JS application may crash, but that's okay.

Open questions

* Is the schema using the resource code of Cliquet ? If so, how do avoid overlap of stored collections ? We could use underscores prefix, and prevent public collections to have a name that starts with underscore :)
* What happen to existing data when schema is changed ? Probably ignore and wait for next
  PUT or PATCH ?
* What happen when records are shared between users ? Do we let other users
  records crash our application when fetching shared records ? We could probably
  run client-side validation using JSON schema on shared records.
* Do we want to provide a collection of custom formats or even types ?
  I'm thinking of recurrent needs for uuid4, geohash, GeoJSON objects, phone, postal codes...

leplatrem:
For example, generate Angular forms from JSON schema http://schemaform.io/

ametaireau:
> Unlike Daybed, I propose that each user owns the schema of the collection.
[...]

PR #39, opened 2015-05-26
Validate JSONSchema for collections (ref #31)

leplatrem:
* [x] Store collections schemas
* [x] Validate records of collections
* [x] Integrate with bucket matchdict
* [x] Write docs


leplatrem:
@Natim I revamped the pull-request with what we had discussed with @phrawzty.

Currently, the schema is a sub-resource of collection where PUT and DELETE are allowed.
I don't remember if we said that the schema should be a collection optional field instead.

I like it this way, but looking at the way the resource class is used for the schema, it could be more natural if it was a colllection optional field.

leplatrem:
Ok, no definitely, it makes a lot more sense if the schema is field of the collection resource. (for managing permissions etc.)

I'll see if I can get it done before my holidays :)

leplatrem:
Ok I did it :)
Please let it rot again : )Take over!

docs/api/collections.rst
```diff
@@ -183,3 +183,228 @@ Deleting a collection
...
+-----------------------
+
+Just modify the ``schema`` attribute of the collection object:
+
+**Example request**
+
+.. code-block:: bash
+
+    $ echo '{
+      "data": {
```
Natim: Why would you put it in data?

docs/api/collections.rst
```diff
@@ -183,3 +183,228 @@ Deleting a collection
...
+
+Records validation
+------------------
+
+Once a schema has been defined, the posted records must match it:
+
+.. code-block:: bash
+
+    $ echo '{
+        "body": "Fails if no title"
```
Natim: This is missing "data"

docs/api/collections.rst
```diff
@@ -183,3 +183,228 @@ Deleting a collection
...
+to manage it.
+
+If the validation succeeds, the record will receive a ``schema`` field with the
+schema version (i.e. the collection current ``last_modified`` timestamp).
+
+It becomes possible to use this ``schema`` field as a filter on the collection
+records endpoint in order to obtain the records that were not validated against a particular
+version of the schema.
+
+For example, ``GET /buckets/default/collections/articles/records?min_schema=123456``.
```
Natim: min_schema_version ?

docs/api/collections.rst
Natim: After thinking this through it might make sense to put it in data since it is an attribute of the collection.

kinto/tests/test_views_collections_schema.py
```diff
@@ -0,0 +1,132 @@
...
+        self.app.post_json('/buckets/blog/collections/articles/records',
+                           {'data': {'body': '<h1>Without title</h1>'}},
+                           headers=self.headers)
+
+    def test_records_receive_the_schema_as_attribute(self):
+        resp = self.app.post_json(RECORDS_URL,
+                                  {'data': VALID_RECORD},
+                                  headers=self.headers,
+                                  status=201)
+        self.assertEqual(resp.json['data']['schema'],
```
Natim: :+1:

Natim:
Looks good to me. It may also fix: #158

docs/api/collections.rst
```diff
@@ -183,3 +183,228 @@ Deleting a collection
...
+
+Just modify the ``schema`` attribute of the collection object:
+
+**Example request**
+
+.. code-block:: bash
+
+    $ echo '{
+      "data": {
+        "schema": {
```
ametaireau: So, schema is inside the `data` attribute?

ametaireau:
I'm glad to have this feature ready.

However, I was more thinking about having the schema defined alongside the `data` attribute (e.g. at the same level).

Is there any reason why this is not the case?

Natim:
Currently the Cliquet Protocol defines object attributes to be in the data attributes.
It is also a vision shared with JSON-API specification.

In this implementation, we are just providing a new property to the collection which is the schema.
Other property may come such as signing, description, title.

Having them all in the data attribute makes sense because schema is finally a collection attribute.

Natim:
However, as for permissions we could say that schema is a specific attribute that must not be inside the data attribute and that we should keep data attribute for arbitrary application specific data.

What I like in the implemented solution, is that we do not have to change the protocol and make a special case for that.

PR #44, opened 2015-06-02
Manage collections by bucket

leplatrem:
Based on #41

### DIFF
--- a/kinto/views/records.py
+++ b/kinto/views/records.py
@@ -7,17 +7,17 @@ class RecordSchema(schema.ResourceSchema):
 
 
 @resource.register(
-    name="collection",
+    name="record",
     record_path=("/buckets/{{bucket_id}}"
                  "/collections/{{collection_id}}/records/{{id}}"),
     collection_path=("/buckets/{{bucket_id}}"
                      "/collections/{{collection_id}}/records"))
-class Collection(resource.BaseResource):
+class Record(resource.BaseResource):
 
     mapping = RecordSchema()
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


leplatrem:
@Natim r?

Especially one point : 

For records, is ``bucket_id`` the parent_id of records ? and ``collection_id`` the ``collection id`` as it was implemented here ?

Or ``collection_id`` is the ``parent_id``, and ``records`` is the ``collection_id`` (i.e. the object type) ?

Natim:
Following our conversation, I think we have:

- ``parent_id`` which is the ``object_uri`` of the parent ie: ``/buckets/<bucket_id/collections/<collection_id>``
- ``collection_id`` should be ``resource_name`` or ``object_type`` and contains the object type ie: ``record``, ``collection``, ``bucket``

kinto/views/records.py
```diff
@@ -0,0 +1,26 @@
...
+                 "/collections/{{collection_id}}/records/{{id}}"),
+    collection_path=("/buckets/{{bucket_id}}"
+                     "/collections/{{collection_id}}/records"))
+class Record(resource.BaseResource):
+
+    mapping = RecordSchema()
+
+    def __init__(self, *args, **kwargs):
+        super(Record, self).__init__(*args, **kwargs)
+        self.collection.collection_id = self.request.matchdict['collection_id']
```
Natim: ``self.collection.collection_id = 'record'``

Natim: Maybe can we get it from the ``resource_name`` directly?

leplatrem:
Thanks @Natim for your feedback, I did these in #48 to avoid merging hell :)

PR #45, opened 2015-06-03
[WIP] Import Cloud Storage backend

leplatrem:
* [x] Imported code from Cliquet 1 code base
* [ ] Setup tests
* [ ] Rebase on master after merge of refactors
* [ ] Polish documentation

### DIFF
--- a/kinto/storage.py
+++ b/kinto/storage.py
@@ -1,0 +1,305 @@
+from functools import wraps
+
+import requests
+import six
+from requests.exceptions import RequestException
+
+from cliquet import logger
+from cliquet.storage import (
+    StorageBase, exceptions, Filter,
+    DEFAULT_ID_FIELD, DEFAULT_MODIFIED_FIELD, DEFAULT_DELETED_FIELD)
+from cliquet.storage.memory import apply_sorting, get_unicity_rules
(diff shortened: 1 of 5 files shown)


kinto/storage.py
```diff
@@ -0,0 +1,305 @@
...
+
+API_PREFIX = "/v0"
+
+
+FILTERS = {
+    COMPARISON.LT: 'lt_',
+    COMPARISON.MIN: 'min_',
+    COMPARISON.MAX: 'max_',
+    COMPARISON.NOT: 'not_',
+    COMPARISON.EQ: '',
```
tarekziade: missing string or is this normal ?

ametaireau: This is normal: when no filter is specified it means that it's ``field=test``. See [the kinto documentation](http://kinto.readthedocs.org/en/latest/api/cliquet/resource.html#filtering) about filtering.

ametaireau: Consider this string as what's prefixed to the name of the field.

leplatrem:
Even though this is pretty cool and fun, we have no use-case so far for a Kinto storage backend.

Dropping.

PR #37, opened 2015-05-15
[WIP] Add buckets

Natim:
I am planning to use json:api format for this.

PR #48, opened 2015-06-05
Add missing resources

leplatrem:
Based on #44 

Revamp of #37 

* [x] Add buckets resource
* [x] Add collections resource
* [x] Add groups resource
* [x] Raise 404 if parent does not exist
* ~~[ ] Plug with branch of permissions in https://github.com/mozilla-services/cliquet/pull/288~~
* [x] Write some specs

**After review**

* [x] Clean-up tests
* [x] Implement deletion of children objects

### DIFF
--- a/kinto/tests/test_views_buckets.py
+++ b/kinto/tests/test_views_buckets.py
@@ -9,15 +9,25 @@ class BucketViewTest(BaseWebTest, unittest.TestCase):
     collection_url = '/buckets'
     record_url = '/buckets/beers'
 
+    def setUp(self):
+        super(BucketViewTest, self).setUp()
+        resp = self.app.put_json(self.record_url,
+                                 MINIMALIST_ITEM,
+                                 headers=self.headers)
+        self.record = resp.json  # XXX: ['data']
+
     def test_buckets_do_not_support_post(self):
         self.app.post(self.collection_url, headers=self.headers,
                       status=405)
 
     def test_buckets_can_be_put_with_simple_name(self):
-        response = self.app.put_json(self.record_url,
-                                     MINIMALIST_ITEM,
-                                     headers=self.headers)
-        self.assertEqual(response.json['id'], 'beers')
+        self.assertEqual(self.record['id'], 'beers')
+
+    def test_collection_endpoint_lists_them_all(self):
+        resp = self.app.get(self.collection_url, headers=self.headers)
+        records = resp.json['items']  # XXX: ['data']
+        self.assertEqual(len(records), 1)
+        self.assertEqual(records[0]['id'], 'beers')
 
     def test_buckets_name_should_be_simple(self):
         self.app.put_json('/buckets/__beers__',
--- a/kinto/tests/test_views_buckets.py
+++ b/kinto/tests/test_views_buckets.py
@@ -41,7 +41,7 @@ class BucketViewTest(BaseWebTest, unittest.TestCase):
 
 class BucketDeletionTest(BaseWebTest, unittest.TestCase):
 
-    record_url = '/buckets/beers/collections/barley'
+    record_url = '/buckets/beers'
 
     def test_buckets_can_be_deleted(self):
         self.app.put_json(self.record_url, MINIMALIST_ITEM,
--- a/kinto/tests/test_views_buckets.py
+++ b/kinto/tests/test_views_buckets.py
@@ -1,7 +1,6 @@
-from .support import BaseWebTest, unittest, get_user_headers
-
-
-MINIMALIST_ITEM = dict()
+from .support import (BaseWebTest, unittest, get_user_headers,
+                      MINIMALIST_BUCKET, MINIMALIST_GROUP,
+                      MINIMALIST_COLLECTION, MINIMALIST_RECORD)
 
 
 class BucketViewTest(BaseWebTest, unittest.TestCase):
@@ -12,7 +11,7 @@ class BucketViewTest(BaseWebTest, unittest.TestCase):
     def setUp(self):
         super(BucketViewTest, self).setUp()
         resp = self.app.put_json(self.record_url,
-                                 MINIMALIST_ITEM,
+                                 MINIMALIST_BUCKET,
                                  headers=self.headers)
         self.record = resp.json  # XXX: ['data']
 
@@ -34,7 +33,7 @@ class BucketViewTest(BaseWebTest, unittest.TestCase):
 
     def test_buckets_name_should_be_simple(self):
         self.app.put_json('/buckets/__beers__',
-                          MINIMALIST_ITEM,
+                          MINIMALIST_BUCKET,
                           headers=self.headers,
                           status=400)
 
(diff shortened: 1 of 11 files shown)


leplatrem:
@Natim I would suggest we skip the integration of permissions for this one. We wait for https://github.com/mozilla-services/cliquet/pull/288 to be merged, and rework another PR

leplatrem:
r? @ametaireau @Natim :)

Some parts of the views code is not dry at all ! But I believe we will merge our utils to handle this :) This is now open to review :)

kinto/views/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+class NameGenerator(generators.Generator):
+    def __call__(self):
+        try:
+            ascii_letters = string.ascii_letters
+        except ImportError:
+            ascii_letters = string.letters  # NOQA
+        alphabet = ascii_letters + string.digits + '-'
+        letters = [random.choice(alphabet) for x in range(8)]
+        return ''.join(letters)
```
leplatrem: Note: this is not collision-proof. IHMO it shouldn't prevent merging, but could deserve an enhancement ticket :)

kinto/tests/test_views_groups.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_unknown_bucket_raises_404(self):
+        other_bucket = self.collection_url.replace('beers', 'sodas')
+        self.app.get(other_bucket, headers=self.headers, status=404)
+
+    def test_groups_are_isolated_by_bucket(self):
+        other_bucket = self.record_url.replace('beers', 'water')
+        self.app.get(other_bucket, headers=self.headers, status=404)
+
+    def test_groups_creation_fails_if_bucket_is_not_found(self):
+        pass
```
leplatrem: This should be removed.

kinto/tests/test_views_collections.py
```diff
@@ -1,69 +1,70 @@
...
+                          headers=self.headers)
+        self.app.delete(self.record_url, headers=self.headers)
+        self.app.get(self.record_url, headers=self.headers,
+                     status=404)
+
+    def test_records_of_collection_are_deleted_too(self):
+        pass
+
+    def test_permissions_associated_are_deleted_too(self):
+        pass
```
leplatrem: Oops, to be done !

kinto/tests/test_views_collections.py
```diff
@@ -1,69 +1,70 @@
...
+        self.app.put_json('/buckets/beers', {}, headers=self.headers)
+
+    def test_collections_can_be_deleted(self):
+        self.app.put_json(self.record_url, MINIMALIST_ITEM,
+                          headers=self.headers)
+        self.app.delete(self.record_url, headers=self.headers)
+        self.app.get(self.record_url, headers=self.headers,
+                     status=404)
+
+    def test_records_of_collection_are_deleted_too(self):
```
leplatrem: Oops to be done !

kinto/tests/test_views_collections.py
```diff
@@ -1,69 +1,70 @@
...
+    def test_unknown_bucket_raises_404(self):
+        other_bucket = self.collection_url.replace('beers', 'sodas')
+        self.app.get(other_bucket, headers=self.headers, status=404)
+
+    def test_collections_are_isolated_by_bucket(self):
+        other_bucket = self.record_url.replace('beers', 'water')
+        self.app.get(other_bucket, headers=self.headers, status=404)
+
+    def test_collections_creation_fails_if_bucket_is_not_found(self):
+        pass
```
leplatrem: This should be removed.

kinto/tests/test_views_buckets.py
```diff
@@ -0,0 +1,63 @@
...
+                     status=404)
+
+    def test_every_collections_are_deleted_too(self):
+        pass
+
+    def test_every_groups_are_deleted_too(self):
+        pass
+
+    def test_every_records_are_deleted_too(self):
+        pass
```
leplatrem: To be done !

kinto/views/__init__.py
Natim: We add a nice thing in Daybed: https://github.com/spiral-project/daybed/blob/cb5efbf9b2ef326154a638c242d64ae001600ba4/daybed/backends/id_generators.py#L18-L41

kinto/views/groups.py
```diff
@@ -0,0 +1,35 @@
...
+
+    mapping = GroupSchema()
+
+    def __init__(self, *args, **kwargs):
+        super(Group, self).__init__(*args, **kwargs)
+
+        bucket_id = self.request.matchdict['bucket_id']
+        object_exists_or_404(self.request,
+                             collection_id='bucket',
+                             object_id=bucket_id)
```
leplatrem: collection_id comes from internal Cliquet, and should be explicit : 
```python
from kinto.views.buckets import Bucket

collection_id = Bucket(self.request).collection.collection_id
```
??

kinto/views/groups.py
```diff
@@ -0,0 +1,35 @@
...
+
+        parent_id = '/buckets/%s' % bucket_id
```
leplatrem: Object URI building is not DRY.

kinto/views/__init__.py
leplatrem: yes good point !

ametaireau: This doesn't prevent merging of this, but I think the name generation should be pluggable. Having an id generator is something we had in daybed and I think we should continue like this: it helps users to generate whatever they want as a new name. for instance the koremutake generator for daybed @natim pointed.

kinto/views/buckets.py
```diff
@@ -0,0 +1,39 @@
...
+
+from kinto.views import NameGenerator
+
+
+buckets_options = {
+    'collection_methods': ('GET',)
+}
+
+
+@resource.register(name="bucket", **buckets_options)
```
ametaireau: Why not passing the bucket options directly here?

ametaireau: (it would use the call arguments syntax, which is easier to follow)

kinto/views/buckets.py
```diff
@@ -0,0 +1,39 @@
...
+class Bucket(resource.BaseResource):
+
+    def __init__(self, *args, **kwargs):
+        super(Bucket, self).__init__(*args, **kwargs)
```
ametaireau: When reading this, I see that we're inheriting a bunch of methods. Should we instead add hooks in the `BaseResource` class?

kinto/views/__init__.py
leplatrem: Created issue in Cliquet https://github.com/mozilla-services/cliquet/issues/292

kinto/views/buckets.py
leplatrem: What do you have in mind ?

I would go for https://github.com/mozilla-services/cliquet/issues/32 instead of resource hooks :)

kinto/views/buckets.py
```diff
@@ -0,0 +1,37 @@
...
+                   record_path='/buckets/{{id}}')
+class Bucket(resource.BaseResource):
+
+    def __init__(self, *args, **kwargs):
+        super(Bucket, self).__init__(*args, **kwargs)
+        # Buckets are not isolated by user, unlike Cliquet resources.
+        self.collection.parent_id = ''
+        self.collection.id_generator = NameGenerator()
+
+    def delete(self):
```
ametaireau: So, we aren't deleting permissions associated with this bucket?

ametaireau:
That's pretty cool and I think we should merge it for now. I'll create a new issue to have the permissions deleted on deletion of the collections / buckets.

PR #52, opened 2015-06-09
Add and Remove group from user principals.

Natim:


leplatrem:
I really wonder how you had the courage to write all this without tests! ;))

kinto/views/groups.py
```diff
@@ -29,3 +31,38 @@ def __init__(self, *args, **kwargs):
...
+        for group in body['items']:
+            for member in group['members']:
+                group_id = '%s/%s' % (get_object_id(self.request), group)
+                permission.remove_user_principal(
+                    self.request.authenticated_userid,
+                    group_id)
+        return body
+
+    def process_record(self, new, old=None):
+        if old is not None:
```
leplatrem: I believe it's the contrary :)

kinto/views/groups.py
```diff
@@ -29,3 +31,38 @@ def __init__(self, *args, **kwargs):
...
+            # Add the group to the member principal
+            group_id = get_object_id(self.request)
+            permission.add_user_principal(user_id, group_id)
+
+        for member in removed_members:
+            # Remove the group from the member principal
+            group_id = get_object_id(self.request)
+            permission.remove_user_principal(user_id, group_id)
+
+        return new
```
leplatrem: I think that should happen after the storage update was performed. 

Even if using ``process_record()`` is more elegant, I guess that with the existing code (and no event system), we would have to override every endpoint post/put/patch in order to run this code after the storage operation was executed :|

Natim: Yes I did that first but then I decided to use process_record instead. In either way if one of the two fails we will be out of sync.

ametaireau:
See #57 instead.

PR #59, opened 2015-06-11
Plug custom authz

leplatrem:
* [x] Merge ``permission.py`` and ``authorization.py``
* [x] Inherit ``ProtectedResource`` instead of ``BaseResource``
* [x] Add tests for basic views permissions for each kind object
* [x] Add tests for inheritance from bucket
* [x] Add tests for inheritance from collection
* [ ] ~~Add tests for permissions via users groups~~ → https://github.com/mozilla-services/kinto/pull/58
* [ ] ~~Handle list of principals from settings for bucket creation~~ → https://github.com/mozilla-services/cliquet/pull/308


kinto/tests/support.py
```diff
@@ -42,6 +42,9 @@ def _get_test_app(self, settings=None):
 
     def get_app_settings(self, additional_settings=None):
         settings = cliquet_support.DEFAULT_SETTINGS.copy()
+        settings['cliquet.cache_backend'] = 'cliquet.cache.memory'
+        settings['cliquet.storage_backend'] = 'cliquet.storage.memory'
+        settings['cliquet.permission_backend'] = 'cliquet.permission.redis'
```
ametaireau: So, why everything is in memory but the permission backend?

Natim: This was to be able to use ``redis-cli monitor`` to debug the permission backend.
I supose that we should put them all on the memory backend now that tests are passing.

kinto/tests/test_views_collections.py
```diff
@@ -35,12 +35,15 @@ def test_collections_name_should_be_simple(self):
                           headers=self.headers,
                           status=400)
 
-    def test_unknown_bucket_raises_404(self):
+    def test_unknown_bucket_raises_403(self):
```
ametaireau: why should unknown raise a 403?

Natim: Because you don't know if it exists or if you don't have access to it.
The permission backend doesn't give you a allowed if the the object doesn't exists.
So we should call the storage backend before the permission backend if we wanted a 404 here.

Also most security guys will tell you that it is better to return a 403 in that case.

Natim: More information about this here: http://security.stackexchange.com/questions/46171/is-it-a-good-practice-to-show-403-unauthorized-access-error-to-user

ametaireau: They will actually tell you it's better to return a 404, otherwise you disclose that the resource exists.

ametaireau: Also, s/guys/folks.

Natim: > They will actually tell you it's better to return a 404, otherwise you disclose that the resource exists.

Oh yes you are right. I guess returning always 404 or always a 403 is not really a big change. Also I for a webservice I would rather return a 403 in all cases rather than a 404 in all cases.

Why ? Because it is important to tell the user that something is wrong with the permission she sets. In the case of unknown it is because an non existing object has no permission sets to it.

Also if you've got permissions on the bucket you will get a 404.

ametaireau: The link you pasted doesn't buy us much: it says it's a good practice and it's a bad practice.

[The HTTP spec](http://pretty-rfc.herokuapp.com/RFC2616#status.403) states:

>  If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead. 

I think that's for us to make a choice here: do we want to disclose this information or not?

A 403 has one advantage over a 404: it's clear that the resource isn't accessible by the connected user, so for instance, if she's not connected with the right account she can find it out and act properly.

It also have a disadvantage:  It leaks the fact that the resource exists to another user, which could then use this information in many different ways.

ametaireau:
r+ minus the discussion on 403 vs 404.

kinto/tests/test_views_collections.py
Natim: > t also have a disadvantage: It leaks the fact that the resource exists to another user, which could then use this information in many different ways.

As I said in this case the resource doesn't exists so it obviously doesn't disclose the fact that it exists :)

ametaireau: Here, we are returning a 403 if the resource doesn't exist. When a resource doesn't exist, a 404 should be returned. 403 means that the resource exists but that you cannot access it.

Natim: According to the documentation: 

> Status code 403 responses are the result of the web server being configured to deny access, for some reason, to the requested resource by the client.

Which doesn't implies necessarily that the resource exists. It just implies that the user cannot have any information about the resource.

I like the idea to return 404 when the user has the right to know that the resource doesn't exists and 403 in case it don't have the right to know either if it exists or not.

PR #83, opened 2015-06-26
use pytest

magopian:
Caveats:
- it's not using the nose-mocha-reporter (I guess it could be adapted to pytest though, below is a screenshot using pytest-sugar)
- there's no `cover-min-percentage` option that I know of for pytest-cov

screen shot 2015-06-26 at 11 23 27

PR #85, opened 2015-06-29
Use pytest

Natim:
Replace #83


magopian:
r+wc, thanks for picking up the ball :+1:

Issue #70, opened 2015-06-17
Create parent objects if missing ?


leplatrem:
Currently, in order to post a record in a new collection, the client shall previously perform two ``PUT`` requests (on bucket and on collection).

I believe this is not very convenient for clients (@n1k0?).

What if, on record POST/PUT, we create the collection and the bucket ? *(if they are missing and if the client is allowed to)*

Natim:
This is something we want at least for the default bucket.

leplatrem:
Is the «default bucket» what I naively called «personal bucket» ?

Natim:
Actually we could have both:
- A default one where everybody could create collection and when looking for ``/buckets/default/collections`` will get only its collection
- A personal one created automatically for him ('~' or 'personal')

PR #71, opened 2015-06-29
Personal bucket ~

Natim:
Related #70 

In the original plan, we had imagined a special bucket named ``~``, that redirects to an implicitly created bucket whose name is the user id.

We shall implement it !

See https://github.com/mozilla-services/kinto/blame/master/docs/api/collections.rst#L15

### DIFF
--- a/docs/api/collections.rst
+++ b/docs/api/collections.rst
@@ -11,10 +11,17 @@ A collection is a mapping with the following attribute:
 
 .. note::
 
-    By default users have a bucket that is used for their own data.
-    Application can use this default bucket with the ``default`` shortcut.
-    ie: ``/buckets/default/collections/contacts`` will be the current
-    user contacts.
+    By default users are assigned to a bucket that is used for their
+    personal data.
+
+	Application can use this default bucket with the ``default``
+    shortcut: ie ``/buckets/default/collections/contacts`` will be
+    the current user contacts.
+
+    Internally the user default bucket is assigned to an id that can
+    later be used to share data from a user personnal bucket.
+
+    Collection on a the default bucket are created silently on first access.
 
 
 Creating a collection
--- a/docs/api/collections.rst
+++ b/docs/api/collections.rst
@@ -14,7 +14,7 @@ A collection is a mapping with the following attribute:
     By default users are assigned to a bucket that is used for their
     personal data.
 
-	Application can use this default bucket with the ``default``
+    Application can use this default bucket with the ``default``
     shortcut: ie ``/buckets/default/collections/contacts`` will be
     the current user contacts.
 
--- a/docs/api/collections.rst
+++ b/docs/api/collections.rst
@@ -9,12 +9,12 @@ A collection is a mapping with the following attribute:
 
 * ``permissions``: (*optional*) the :term:`ACLs <ACL>` for the collection object
 
-.. .. note::
+.. note::
 
-..     By default users have a bucket that is used for their own data.
-..     Application can use this default bucket with the ``default`` shortcut.
-..     ie: ``/buckets/default/collections/contacts`` will be the current
-..     user contacts. See #71 for more discussion.
+    By default users have a bucket that is used for their own data.
+    Application can use this default bucket with the ``default`` shortcut.
+    ie: ``/buckets/default/collections/contacts`` will be the current
+    user contacts.
 
 
 Creating a collection
--- a/kinto/tests/support.py
+++ b/kinto/tests/support.py
@@ -46,11 +46,9 @@ class BaseWebTest(object):
         settings.update(**DEFAULT_SETTINGS)
         settings['cliquet.cache_backend'] = 'cliquet.cache.memory'
         settings['cliquet.storage_backend'] = 'cliquet.storage.memory'
-        settings['cliquet.permission_backend'] = 'cliquet.permission.redis'
+        settings['cliquet.permission_backend'] = 'cliquet.permission.memory'
         settings['cliquet.project_name'] = 'cloud storage'
         settings['cliquet.project_docs'] = 'https://kinto.rtfd.org/'
-        settings['multiauth.authorization_policy'] = (
-            'kinto.authorization.AuthorizationPolicy')
         settings['cliquet.userid_hmac_secret'] = "this is not a secret"
 
         if additional_settings is not None:
--- a/kinto/tests/support.py
+++ b/kinto/tests/support.py
@@ -17,8 +17,8 @@ MINIMALIST_COLLECTION = {'data': dict()}
 MINIMALIST_GROUP = {'data': dict(members=['fxa:user'])}
 MINIMALIST_RECORD = {'data': dict(name="Hulled Barley",
                                   type="Whole Grain")}
-USER_PRINCIPAL = 'basicauth:8a931a10fc88ab2f6d1cc02a07d3a81b5d4768f' \
-                 '6f13e85c5d8d4180419acb1b4'
+USER_PRINCIPAL = 'basicauth:aaedca130273574dd2bd6c3acad57f3545b662a974fa4320' \
+                 '236f25fe474676d6'
 
 
 class BaseWebTest(object):
@@ -51,6 +51,7 @@ class BaseWebTest(object):
         settings['cliquet.project_docs'] = 'https://kinto.rtfd.org/'
         settings['multiauth.authorization_policy'] = (
             'kinto.tests.support.AllowAuthorizationPolicy')
+        settings['cliquet.userid_hmac_secret'] = "azerty"
 
         if additional_settings is not None:
             settings.update(additional_settings)
--- a/kinto/tests/support.py
+++ b/kinto/tests/support.py
@@ -4,9 +4,9 @@ except ImportError:
     import unittest  # NOQA
 
 import webtest
-from cliquet import utils
 from pyramid.security import IAuthorizationPolicy
 from zope.interface import implementer
+from cliquet import utils
 from cliquet.tests import support as cliquet_support
 from kinto import main as testapp
 from kinto import DEFAULT_SETTINGS
@@ -17,8 +17,8 @@ MINIMALIST_COLLECTION = {'data': dict()}
 MINIMALIST_GROUP = {'data': dict(members=['fxa:user'])}
 MINIMALIST_RECORD = {'data': dict(name="Hulled Barley",
                                   type="Whole Grain")}
-USER_PRINCIPAL = 'basicauth:aaedca130273574dd2bd6c3acad57f3545b662a974fa4320' \
-                 '236f25fe474676d6'
+USER_PRINCIPAL = 'basicauth:3a0c56d278def4113f38d0cfff6db1b06b84fcc4384ee890' \
+                 'cf7bbaa772317e10'
 
 
 class BaseWebTest(object):
@@ -46,12 +46,12 @@ class BaseWebTest(object):
         settings.update(**DEFAULT_SETTINGS)
         settings['cliquet.cache_backend'] = 'cliquet.cache.memory'
         settings['cliquet.storage_backend'] = 'cliquet.storage.memory'
-        settings['cliquet.permission_backend'] = 'cliquet.permission.memory'
+        settings['cliquet.permission_backend'] = 'cliquet.permission.redis'
         settings['cliquet.project_name'] = 'cloud storage'
         settings['cliquet.project_docs'] = 'https://kinto.rtfd.org/'
         settings['multiauth.authorization_policy'] = (
-            'kinto.tests.support.AllowAuthorizationPolicy')
-        settings['cliquet.userid_hmac_secret'] = "azerty"
+            'kinto.authorization.AuthorizationPolicy')
+        settings['cliquet.userid_hmac_secret'] = "this is not a secret"
 
         if additional_settings is not None:
             settings.update(additional_settings)
@@ -79,9 +79,8 @@ class BaseWebTest(object):
 @implementer(IAuthorizationPolicy)
 class AllowAuthorizationPolicy(object):
     def permits(self, context, principals, permission):
-        if USER_PRINCIPAL in principals:
-            return True
-        return False
+        # Cliquet default authz policy uses prefixed_userid.
+        return USER_PRINCIPAL in ([context.prefixed_userid] + principals)
 
     def principals_allowed_by_permission(self, context, permission):
         raise NotImplementedError()  # PRAGMA NOCOVER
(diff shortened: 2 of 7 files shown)


kinto/tests/test_default_bucket.py
```diff
@@ -0,0 +1,39 @@
...
+from cliquet.utils import hmac_digest
+
+
+class DefaultBucketViewTest(BaseWebTest, unittest.TestCase):
+
+    bucket_url = '/buckets/default'
+    collection_url = '/buckets/default/collections/tasks'
+
+    def setUp(self):
+        super(DefaultBucketViewTest, self).setUp()
```
leplatrem: Probably useless :)

kinto/views/buckets.py
```diff
@@ -1,8 +1,57 @@
...
+from cliquet.utils import hmac_digest
+from cliquet.views.batch import build_request
 
 from kinto.views import NameGenerator
 
 
+@view_config(route_name='default_bucket', permission=NO_PERMISSION_REQUIRED)
+def default_bucket(request):
+    if (not hasattr(request, 'prefixed_userid') or
+            request.prefixed_userid is None):
```
leplatrem: getattr(obj, 'prefixed_user_id', None)

kinto/views/buckets.py
```diff
@@ -1,8 +1,57 @@
...
+        raise HTTPForbidden  # Pass through the forbidden_view_config
+
+    settings = request.registry.settings
+    hmac_secret = settings['cliquet.userid_hmac_secret']
+    bucket_id = hmac_digest(hmac_secret, request.prefixed_userid)[:32]
```
leplatrem: Can you explain what you do here ?

I don't see any spec for this in the tests :)

kinto/views/buckets.py
```diff
@@ -1,8 +1,57 @@
...
+            'body': {"data": {}},
+            'headers': {'If-None-Match': '*'}
+        })
+        request.invoke_subrequest(subrequest)
+
+    subrequest = build_request(request, {
+        'method': request.method,
+        'path': path,
+        'body': request.body
+    })
```
leplatrem: If we do PUT on /buckets/default/ then it's done twice ?

kinto/views/buckets.py
```diff
@@ -1,8 +1,57 @@
...
+
+    # Make sure bucket exists
+    # XXX: Is there a better way to do a GET or CREATE?
+    subrequest = build_request(request, {
+        'method': 'PUT',
+        'path': '/buckets/%s' % bucket_id,
+        'body': {"data": {}},
+        'headers': {'If-None-Match': '*'}
+    })
+    request.invoke_subrequest(subrequest)
```
leplatrem: We probably want to test that it does not raise a 412 on the second request

leplatrem:
Ok, that's one way to do it indeed :)

I would have tried (*but no thought about it yet*) to modify what we had here:
https://github.com/mozilla-services/kinto/blob/master/kinto/views/collections.py#L17-L19

kinto/views/buckets.py
Natim: I am building the bucket_id of the user using an hmac in order to make sure it is unguessable.

kinto/views/buckets.py
Natim: Why? Actually 412 is a feature, it means that the bucket already existed.

kinto/views/buckets.py
Natim: Yes, once to create it if needed (with a 412 if not needed) and once to update the bucket information (data and permission).

kinto/views/buckets.py
Natim: Oh you means adding a test to make sure two call on the same bucket in a row works?

Natim:
My first go was actually to modify the bucket and collection view to create the bucket and the collection if they didn't exists yet.

But this was meaning:

 - Checking the ``bucket_id`` and ``collection_id`` are correct
 - Adding the correct permissions

Maybe it worth to try to do it your way anyway.

kinto/views/buckets.py
leplatrem: Exactly, add a test to verify that the ``412`` is not exposed to the user on the second request

kinto/views/buckets.py
leplatrem: Yes, it makes sense.

Don't you think we could add a small comment here to remember that we could do some optimization if needed later ?

Natim: Like ``#XXX: Improvment Avoid doing PUT twice on PUT /buckets/default``

leplatrem: +1 (and collection ?)

kinto/views/buckets.py
Natim: Actually it was raising a 412 :8ball:

kinto/views/buckets.py
Natim: Actually I added the code to not do the call twice.

leplatrem: Oh great !

kinto/views/buckets.py
```diff
@@ -1,8 +1,68 @@
...
+        except HTTPPreconditionFailed:
+            # The bucket already existed
+            pass
+
+    # Make sure the collection exists
+    subpath = request.matchdict['subpath']
+    if subpath.startswith('/collections/'):
+        collection_id = subpath.split('/')[2]
+        if request.method.lower() != 'put' or \
+           not request.path.endswith(collection_id):
```
leplatrem:
```python
is_put_collection = (request.method.lower() == 'put' and request.path.endswith(collection_id))

if not is_put_collection:
    ....
```

kinto/views/buckets.py
```diff
@@ -1,8 +1,68 @@
...
+                'method': 'PUT',
+                'path': '/buckets/%s/collections/%s' % (
+                    bucket_id, collection_id),
+                'body': {"data": {}},
+                'headers': {'If-None-Match': '*'}
+            })
+            try:
+                request.invoke_subrequest(subrequest)
+            except HTTPPreconditionFailed:
+                # The bucket already existed
```
leplatrem: The collection already existed

leplatrem:
I was going to say **r+**, but I realize that some parts of the docs might deserve an update (ex. ``docs/api/collections.rst``)

Natim:
I have uncommented the note in the docs/api/collections

leplatrem:
r+

docs/api/collections.rst
```diff
@@ -9,12 +9,12 @@ A collection is a mapping with the following attribute:
...
 * ``permissions``: (*optional*) the :term:`ACLs <ACL>` for the collection object
 
-.. .. note::
+.. note::
 
-..     By default users have a bucket that is used for their own data.
-..     Application can use this default bucket with the ``default`` shortcut.
-..     ie: ``/buckets/default/collections/contacts`` will be the current
-..     user contacts. See #71 for more discussion.
+    By default users have a bucket that is used for their own data.
```
ametaireau: "users are assigned a bucket"

kinto/tests/support.py
```diff
@@ -17,8 +17,8 @@
 MINIMALIST_GROUP = {'data': dict(members=['fxa:user'])}
 MINIMALIST_RECORD = {'data': dict(name="Hulled Barley",
                                   type="Whole Grain")}
-USER_PRINCIPAL = 'basicauth:8a931a10fc88ab2f6d1cc02a07d3a81b5d4768f' \
-                 '6f13e85c5d8d4180419acb1b4'
+USER_PRINCIPAL = 'basicauth:aaedca130273574dd2bd6c3acad57f3545b662a974fa4320' \
```
ametaireau: Why is the principal changing here?

kinto/tests/support.py
```diff
@@ -51,6 +51,7 @@ def get_app_settings(self, additional_settings=None):
         settings['cliquet.project_docs'] = 'https://kinto.rtfd.org/'
         settings['multiauth.authorization_policy'] = (
             'kinto.tests.support.AllowAuthorizationPolicy')
+        settings['cliquet.userid_hmac_secret'] = "azerty"
```
ametaireau: I would rather use a secret which explicitly says it's not secret (`this is not secret`)

ametaireau: And I guess this is why the principal has been changed :)

kinto/tests/support.py
Natim: This is because the user_hmac_id was not set in the config and is now set.

kinto/tests/support.py
Natim: This means I will have to change all the principals again :(

kinto/tests/test_default_bucket.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_bucket_id_is_an_uuid(self):
+        bucket = self.app.get(self.bucket_url, headers=self.headers)
+        bucket_id = bucket.json['data']['id']
+        try:
+            uuid.UUID(bucket_id)
+        except ValueError:
+            self.fail('bucket_id: %s is not a valid UUID.' % bucket_id)
+
+    def test_second_call_on_default_bucket_doesnt_raise_a_412(self):
```
ametaireau: Why would it raise a 412 ?

Natim: It does in case:
 - The bucket already exists: https://github.com/mozilla-services/kinto/pull/71/files#diff-580dee783aeea3c8487d07890dfdbcb9R35
 - The collection already exists: https://github.com/mozilla-services/kinto/pull/71/files#diff-580dee783aeea3c8487d07890dfdbcb9R35

docs/api/collections.rst
```diff
@@ -9,12 +9,19 @@ A collection is a mapping with the following attribute:
...
+    personal data.
+
+	Application can use this default bucket with the ``default``
+    shortcut: ie ``/buckets/default/collections/contacts`` will be
+    the current user contacts.
+
+    Internally the user default bucket is assigned to an id that can
+    later be used to share data from a user personnal bucket.
+
+    Collection on a the default bucket are created silently on first access.
```
leplatrem: When I read this, I realize that I missed a big part of the spec: **objects are created on access**.

I don't want to retain the merge of this, but we may want to open an issue and restrict the creation of objects to "PUT/POST" requests.

Otherwise, once I know the bucket id of someone, I can fill it with millions of collections, just by trying to reach them with GET.

kinto/views/buckets.py
```diff
@@ -1,8 +1,70 @@
+from pyramid.httpexceptions import HTTPForbidden, HTTPPreconditionFailed
+from pyramid.security import NO_PERMISSION_REQUIRED
+from pyramid.view import view_config
+
 from cliquet import resource
+from cliquet.utils import hmac_digest
+from cliquet.views.batch import build_request
```
leplatrem: maybe move this utils then

docs/api/collections.rst
Natim: The magic is it only works if you use the default endpoint so you cannot create collections for other people.

leplatrem: OH very good point !

kinto/tests/test_default_bucket.py
ametaireau: Can you explicit that in the tests?

Natim: Is ``test_call_on_default_bucket_when_it_exists_doesnt_raise_412`` good enough?

docs/api/collections.rst
ametaireau: Well, is it possible if you share the bucket:read permission with someone else?

Natim: No because the creation code is in the ``/buckets/default`` endpoint only so that the code won't be executed if you call /buckets/{user_bucket_id}

kinto/views/buckets.py
```diff
@@ -1,8 +1,70 @@
...
+        raise HTTPForbidden  # Pass through the forbidden_view_config
+
+    settings = request.registry.settings
+    hmac_secret = settings['cliquet.userid_hmac_secret']
+    # Build the user unguessable bucket_id UUID from its user_id
+    bucket_id = hmac_digest(hmac_secret, request.prefixed_userid)[:32]
+    path = request.path.replace('default', bucket_id)
+
+    # Make sure bucket exists
+    is_bucket_put = (request.method.lower() == 'put' and
```
ametaireau: `bucket_creation`?

kinto/views/buckets.py
```diff
@@ -1,8 +1,70 @@
...
+        subrequest = build_request(request, {
+            'method': 'PUT',
+            'path': '/buckets/%s' % bucket_id,
+            'body': {"data": {}},
+            'headers': {'If-None-Match': '*'}
+        })
+        try:
+            request.invoke_subrequest(subrequest)
+        except HTTPPreconditionFailed:
+            # The bucket already existed
```
ametaireau: nit: already exists

kinto/views/buckets.py
Natim: I have created https://github.com/mozilla-services/cliquet/pull/344 for that.

kinto/views/buckets.py
```diff
@@ -1,8 +1,70 @@
...
+            pass
+
+    # Make sure the collection exists
+    subpath = request.matchdict['subpath']
+    if subpath.startswith('/collections/'):
+        collection_id = subpath.split('/')[2]
+        is_collection_put = (request.method.lower() == 'put' and
```
ametaireau: `bucket_creation`?

kinto/views/buckets.py
Natim: ``bucket_put`` can be update or creation.

kinto/views/buckets.py
Natim: ``collection_put`` can be update or creation.

kinto/views/buckets.py
```diff
@@ -1,8 +1,70 @@
...
+                'path': '/buckets/%s/collections/%s' % (
+                    bucket_id, collection_id),
+                'body': {"data": {}},
+                'headers': {'If-None-Match': '*'}
+            })
+            try:
+                request.invoke_subrequest(subrequest)
+            except HTTPPreconditionFailed:
+                # The collection already existed
+                pass
```
ametaireau: We could put the bucket check+creation and collection check+creation in separate functions to ease readability and have a smaller function.

ametaireau:
r+ with nits.

kinto/views/buckets.py
leplatrem: merged!

PR #91, opened 2015-06-29
Changelog review.

Natim:
r? @ametaireau

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -1,0 +1,10 @@
+Contributors
+============
+
+* Alexis Metaireau <alexis@mozilla.com>
+* Fil <fil@rezo.net>
+* Mathieu Agopian <mathieu@agopian.info>
+* Mathieu Leplatre <mathieu@mozilla.com>
+* Nicolas Perriault <nperriault@mozilla.com>
+* Rémy Hubscher <rhubscher@mozilla.com>
+* Tarek Ziade <tarek@mozilla.com>
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -119,9 +119,8 @@ man_pages = [
 # (source start file, target name, title, author,
 #  dir menu entry, description, category)
 texinfo_documents = [
-  ('index', 'Kinto', u'Kinto Documentation',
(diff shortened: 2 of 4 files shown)


CONTRIBUTORS.rst
```diff
@@ -0,0 +1,10 @@
+Contributors
+============
+
+* Alexis Metaireau <alexis@mozilla.com>
+* Fil <fil@rezo.net>
+* Mathieu Agopian <mathieu@agopian.info>
+* Mathieu Leplatre <mathieu@mozilla.com>
+* Nicolas Perriault <nperriault@mozilla.com>
+* Rémy Hubscher <rhubscher@mozilla.com>
+* Tarek Ziade <tarek@mozilla.com>
```
ametaireau: Cool, we need to make sure we update this file each time we do a release (if we had templates of what we should do for a release, we could include this step)

docs/conf.py
```diff
@@ -121,6 +121,7 @@
 texinfo_documents = [
   ('index', 'Kinto', u'Kinto Documentation',
    u'Mozilla Services — Da French Team', 'Kinto',
-   'A storage service for data sync, share with fine-grained permissions.',
+   'A storage service for data storage, synchronization and sharing '
```
ametaireau: "A storage service for storage" sounds weird to me. The fact that we have fine grained permissions is IMO too much detailed information. I would rephrase this with "A remote storage service with syncing and sharing abilities"?

setup.py
```diff
@@ -1,14 +1,18 @@
 import os
+import codecs
 from setuptools import setup, find_packages
 
 here = os.path.abspath(os.path.dirname(__file__))
 
-with open(os.path.join(here, 'README.rst')) as f:
+with codecs.open(os.path.join(here, 'README.rst'), encoding='utf-8') as f:
```
ametaireau: we could put that in a function named `read_file`.

ametaireau:
LGTM with comments.

CONTRIBUTORS.rst
Natim: I have updated this: https://github.com/mozilla-services/kinto/pull/86#issue-91756118

Not sure if we put the template somewhere else.

PR #93, opened 2015-06-29
Add pytest-cache

Natim:
r? @magopian

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -8,3 +8,4 @@ pytest
 pytest-cache
 pytest-cover
 pytest-sugar
+pytest-xdist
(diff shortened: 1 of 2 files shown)


magopian:
yep, that allows using the `--lf` (for "last failed") to the test run command, which will only re-run the tests that failed. It's awesome, you should use it ;)

Also, if you want to give it a try:
- `pip install pytest-xdist`
- `py.test --cov-report term-missing --cov-fail-under 100 --cov kinto -n 4 --nosugar`

pytest-xdist is not compatible with pytest-sugar I believe, hence the `--nosugar`.

This will run the tests in parallel on 4 workers/cores, which should be much faster ;)

Natim:
Yes it is much faster.

Natim:
PR Updated.

Natim:
I don't think pytest-xdist is a good idea here since test are instable with it and not that much faster :)

magopian:
I agree. `pytest-cache` would still be good though

Natim:
Yes see https://github.com/mozilla-services/kinto/pull/98

PR #95, opened 2015-06-29
Default memory update documentation

leplatrem:
I may have missed some bits.

Please check the Docker build (couldn't do it from the airport wifi :))

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -43,13 +43,27 @@ Also, the set of settings mentionned below might deserve some review or adjustme
 Monitoring
 ----------
 
+In order to enable *Cliquet* monitoring features like *statsd*, install
+extra requirements:
+
+::
+
+    pip install "cliquet[monitoring]"
+
+And configure its URL:
+
+.. code-block :: ini
+
+    # StatsD
+    cliquet.statsd_url = udp://carbon.server:8125
+
+
+In order to enable *Heka* logging output:
+
 .. code-block :: ini
 
     # Heka
     cliquet.logging_renderer = cliquet.logs.MozillaHekaRenderer
-
-    # StatsD
-    cliquet.statsd_url = udp://carbon.server:8125
 
 
 With the following configuration, all logs are structured in JSON and
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -14,18 +14,24 @@ Running in production
 Recommended settings
 --------------------
 
-Most default setting values in the application code base are suitable for production.
+Most default setting values in the application code base are suitable
+for production.
 
-But the set of settings mentionned below might deserve some review or adjustments:
+Once `PostgreSQL is installed <postgresql-install>`_, the settings about
+backends as shown in :file:`config/kinto.ini` can be uncommented in order
+to use *PostgreSQL*.
 
+Also, the set of settings mentionned below might deserve some review or adjustments:
 
 .. code-block :: ini
 
+    kinto.flush_endpoint_enabled = false
     cliquet.http_scheme = https
     cliquet.paginate_by = 100
     cliquet.batch_max_requests = 25
     cliquet.storage_pool_maxconn = 50
     cliquet.cache_pool_maxconn = 50
+    cliquet.permission_pool_maxconn = 50
     fxa-oauth.cache_ttl_seconds = 3600
 
 :note:
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -14,12 +14,8 @@ Run locally
 For development
 ---------------
 
-By default, *Kinto* persists the records and internal cache in a PostgreSQL
-database.
-
-The default configuration will connect to the ``postgres`` database on
-``localhost:5432``, with user/password ``postgres``/``postgres``.
-See more details below about installation and setup of PostgreSQL.
+By default, for convenience, *Kinto* persists the records, permissions and
+internal cache in a **volatile** memory backend.
 
 ::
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -2,4 +2,5 @@ FROM python:2.7
 WORKDIR /code
 ADD . /code
 RUN python setup.py develop
+RUN pip install cliquet[postgresql]
```
Natim: Don't we want monitoring here as well?

docs/configuration.rst
```diff
@@ -14,24 +14,30 @@ Running in production
...
-But the set of settings mentionned below might deserve some review or adjustments:
+Once `PostgreSQL is installed <postgresql-install>`_, the settings about
+backends as shown in :file:`config/kinto.ini` can be uncommented in order
+to use *PostgreSQL*.
 
+Also, the set of settings mentionned below might deserve some review or adjustments:
 
 .. code-block :: ini
 
+    kinto.flush_endpoint_enabled = false
```
Natim: Shouldn't this be the default?

docs/installation.rst
```diff
@@ -12,12 +12,8 @@ Run locally
...
 ---------------
 
-By default, *Kinto* persists the records and internal cache in a PostgreSQL
-database.
-
-The default configuration will connect to the ``postgres`` database on
-``localhost:5432``, with user/password ``postgres``/``postgres``.
-See more details below about installation and setup of PostgreSQL.
+By default, for convenience, *Kinto* persists the records, permissions and
+internal cache in a **volatile** memory backend.
```
Natim: which means if you restart the server you loose your data and that if you start multiple process, you will have random behaviors.

kinto/__init__.py
```diff
@@ -14,6 +14,8 @@
 
 
 DEFAULT_SETTINGS = {
+    'cliquet.project_name': 'cloud storage',
+    'cliquet.project_docs': 'https://kinto.rtfd.org/',
```
Natim: https://kinto.readthedocs.org/

kinto/__init__.py
```diff
@@ -14,6 +14,8 @@
...
     'cliquet.cache_backend': 'cliquet.cache.memory',
```
Natim: Also this cannot be overridded yet, see: https://github.com/mozilla-services/kinto/issues/90

kinto/tests/support.py
```diff
@@ -44,11 +44,6 @@ def _get_test_app(self, settings=None):
     def get_app_settings(self, additional_settings=None):
         settings = cliquet_support.DEFAULT_SETTINGS.copy()
         settings.update(**DEFAULT_SETTINGS)
-        settings['cliquet.cache_backend'] = 'cliquet.cache.memory'
```
Natim: We should probably keep that so that we can change the default settings without having to manage that.

Dockerfile
leplatrem: monitoring came already with ``setup.py develop`` since cliquet is supposed to already have it in setup.

docs/configuration.rst
leplatrem: Yes it is, but I believe it is better to review that the value is not set to true on prod :)

Dockerfile
Natim: I made it optionnal as well in https://github.com/mozilla-services/kinto/pull/100

docker-compose.yml
```diff
@@ -10,5 +10,9 @@ web:
   ports:
    - "8888:8888"
   environment:
+    CLIQUET_SESSION_BACKEND: cliquet.cache.postgresql
```
Natim: Isn't it CACHE_BACKEND nowadays?

docs/configuration.rst
```diff
@@ -14,24 +14,30 @@ Running in production
...
     cliquet.storage_pool_maxconn = 50
     cliquet.cache_pool_maxconn = 50
+    cliquet.permission_pool_maxconn = 50
     fxa-oauth.cache_ttl_seconds = 3600
 
 :note:
 
     For an exhaustive list of available settings and their default values,
-    refer to `the source code <https://github.com/mozilla-services/cliquet/blob/2.0.0/cliquet/__init__.py#L26-L78>`_.
+    refer to `the source code <https://github.com/mozilla-services/cliquet/blob/2.1.0/cliquet/__init__.py#L26-L78>`_.
```
Natim: Should we put master here to avoid forgetting to update it? Cliquet 2.2 has been released already so this rotted.

docs/installation.rst
```diff
@@ -56,11 +52,31 @@ Recommended settings for production are listed :ref:`in a dedicated section
...
 
- (*requires PostgreSQL 9.3 or higher*).
+(*requires PostgreSQL 9.4 or higher*).
+
+*Cliquet* dependencies do not include *PostgreSQL* tooling and drivers by
+default. In order to install them, run:
+
+::
+
+    pip install "cliquet[postgresql]"
```
Natim: I think we can now use ``pip install "kinto[postgresql]"``

docs/installation.rst
```diff
@@ -56,18 +53,30 @@ Recommended settings for production are listed :ref:`in a dedicated section
...
 
-Install Python dependencies
----------------------------
+*Cliquet* dependencies do not include *PostgreSQL* tooling and drivers by
+default. In order to install them, run:
 
 ::
 
-    pip install "kinto[postgresql]"
+    pip install "cliquet[postgresql]"
```
leplatrem: @Natim I left it because we've seen that it does not work

Natim: No we didn't :)

Natim: It just doesn't work with ``.[postgresql]`` but not with ``kinto[postgresql]``

Natim: Well apparently you are right. I think we need to fix that.

PR #96, opened 2015-06-29
Restore dependencies versions in setup

leplatrem:
You might had good reasons to remove them. So let's use this PR to discuss your choice.

My concern is about optional dependencies (``cliquet[monitoring]``), how do we remember to restore that when releasing ?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,11 +11,11 @@ with open(os.path.join(here, 'CHANGELOG.rst')) as f:
 
 
 REQUIREMENTS = [
-    'colander',
-    'cornice',
-    'six',
-    'waitress',
-    'cliquet'
+    'colander==1.0',
+    'cornice==1.0.0',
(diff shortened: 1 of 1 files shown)


Natim:
I didn't remove them, I tagged them for the release: https://github.com/mozilla-services/kinto/commit/30faf29432d3c99ad80997f9118a003495c10bbc

Natim:
> about optional dependencies [...] how do we remember to restore that when releasing

If we need to restore them, then we should not remove them.

The aim is not to force postgresql dependency when installing Kinto.

Natim:
I have restored the dependencies for now.
https://github.com/mozilla-services/kinto/commit/64794fdd11356937fdd551b0d20eaa552a4f031e

Natim:
Discussion can takes place here: https://github.com/mozilla-services/kinto/issues/89

PR #99, opened 2015-06-30
Verify pypy supports for Kinto.

Natim:
Cliquet now has pypy support, we shall verify that kinto has it as well.


Natim:
It works :)

PR #90, opened 2015-06-30
Fix default settings URL and project name

Natim:
Actually, this code https://github.com/mozilla-services/kinto/blob/cab991c928dcca17510a5c7a2762d1b64c192526/kinto/__init__.py#L16-L25 is overridding what is in the file.

It should be used as Kinto default settings and override cliquet default settings but should be overriddable by the configuration file.

### DIFF
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -17,6 +17,8 @@ DEFAULT_SETTINGS = {
     'cliquet.cache_backend': 'cliquet.cache.memory',
     'cliquet.permission_backend': 'cliquet.permission.memory',
     'cliquet.storage_backend': 'cliquet.storage.memory',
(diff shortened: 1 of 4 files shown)


Natim:
I wanted to write a test for that, but I don't know how to write it.

Something like writting a new .ini file?

leplatrem:
As far as I understood, the .ini settings come into the ``**settings`` value of the kinto __init__ main function. So it might be possible to test it by calling main() with custom setting values, without having to introduce new .ini files.

Natim:
#343 has been merged, restarting tests.

Natim:
Also we should take a decision regarding the memory backend by default.
In this PR we are keeping cliquet default.

leplatrem:
r- if we change the default backend without a full overview of documentation (like started in #95)

PR #104, opened 2015-07-03
Use cliquet extra requires instead of raw deps

leplatrem:
@Natim r?

It seems to work well, and it's bit better than raw deps. What do you think ?


Natim:
Here is my alternative proposal: https://github.com/mozilla-services/kinto/commit/8107941d92066ff346e21e04588b39da801dad0b

Natim:
This is still better that what we had.

PR #107, opened 2015-07-03
Fix docs syntax/structure, add a TOC to fix RTD template

magopian:
Some small fixes to the syntax and structure of some of the docs.

Regarding the "add a TOC to fix ReadTheDocs template": https://read-the-docs.readthedocs.org/en/latest/theme.html#how-the-table-of-contents-builds
Adding the TOC will make it display correctly in the left menu, and also add the "prev/next" buttons at the bottom of each pages. Much easier to navigate the docs!

There's also a second commit for improvements to the tutorial, and small fixes. I could provide this in another PR if needed. Please take note that I wasn't able to run the full tutorial, because of issues I had following along (see the issues #113 and #115 for example), so I may be missing some things.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -13,15 +13,14 @@ particulary relevant as a storage backend:
 Sync user data between devices
 ==============================
 
-Let's say that we want to do a TodoMVC backend that will sync user
-tasks between her devices.
-
+Let's say that we want to do a `TodoMVC <http://todomvc.com/>`_ backend that
+will sync user tasks between the devices.
 
 In order to separate data between each user, we will use the default
 *personal bucket*.
 
-Unlike other buckets, the collections in the ``default`` bucket are created
-implicitly.
+Unlike other buckets, the :ref:`collections <collections>` in the ``default``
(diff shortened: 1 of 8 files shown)


docs/tutorial.rst
```diff
@@ -115,6 +114,10 @@ Let us fetch our new collection of tasks:
     }
 
 
+Keep a note of the ``ETag`` and of the ``last_modified`` values returned (here
+respectively ``"1434641474977"`` and ``"1434641515332"``), we'll need them for
```
leplatrem: I would have expected them to be equal. Created an issue, discussion happens there: https://github.com/mozilla-services/cliquet/issues/356

leplatrem:
r+

docs/tutorial.rst
Natim: Yes this was fixed in Kinto 1.2, it is related to this bug: https://github.com/mozilla-services/kinto/pull/80

Natim:
r+ with nit (I think we should remove the paragraph about our comments first)

Natim:
I merge and I will update the http calls with 1.2 reality.

PR #121, opened 2015-07-06
Kinto should work with default: remove custom authorization policy.

Natim:


kinto/tests/test_views_buckets.py
```diff
@@ -24,7 +23,6 @@ def get_app_settings(self, extra=None):
         settings['cliquet.bucket_read_principals'] = 'system.Authenticated'
         return settings
 
-    @authorize(True, 'kinto.tests.support.AllowAuthorizationPolicy')
     def test_buckets_are_global_to_every_users(self):
         self.app.get(self.record_url, headers=get_user_headers('alice'))
```
Natim: This should be true with the current kinto authorization.

kinto/tests/test_views_flush.py
```diff
@@ -50,7 +47,6 @@ def test_returns_404_if_not_enabled_in_configuration(self):
         app = self._get_test_app(settings=extra)
         app.post('/__flush__', headers=self.headers, status=404)
 
-    @authorize(authz_class='kinto.tests.support.AllowAuthorizationPolicy')
```
Natim: This should be true with the current kinto authorization, since __flush__ doesn't need any permission (NO_PERMISSION_REQUIRED)

ametaireau:
That's indeed useful to have the default authz policy used here!

ametaireau:
r+ (tests pass)

PR #88, opened 2015-07-06
500 when trying to GET /buckets

Natim:
```
  File "~/mozilla/kinto/authorization.py", line 129, in build_permissions_set
    for obj, permission_list in inheritance_tree[bound_permission].items():
KeyError: 'None:read'
```

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -42,6 +42,16 @@ adjustments:
     refer to `the source code <https://github.com/mozilla-services/cliquet/blob/2.2.0/cliquet/__init__.py#L26-L78>`_.
 
 
+By default, nobody can read buckets list. You can change that using:
+
+.. code-block :: ini
+
+    cliquet.bucket_read_principals = system.Authenticated
+
+Beware that if you do so, everyone will be able to list bucket
+information (including user's personal buckets).
+
+
 Monitoring
 ----------
 
--- a/kinto/authorization.py
+++ b/kinto/authorization.py
@@ -122,6 +122,8 @@ def build_permissions_set(object_uri, unbound_permission,
 
     obj_type = get_object_type(object_uri)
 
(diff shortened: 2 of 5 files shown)


Natim:
This commit fixes the 500 but doesn't fix the fact that any user can query the list of buckets.

My next commit will make a test fail.

Natim:
I think we should discuss this issue.
I am pretty sure it is not what we want.

I think we should deactivate this endpoint by default.

Also this lists user buckets and because it is listing the permissions on the bucket GET , it is easy to link a bucket to a user.

Natim:
After discussion what we want to do is to return a 403 by default + add a documentation note to explain how to authorize somebody to call this endpoint.

kinto/authorization.py
```diff
@@ -122,6 +122,9 @@ def build_permissions_set(object_uri, unbound_permission,
 
     obj_type = get_object_type(object_uri)
 
+    if obj_type is None:
```
leplatrem: Please add a comment (and a test?) for this ``if``

Natim: There is a test for it. I will add a comment. ( https://github.com/mozilla-services/kinto/pull/88/files#diff-78b677143edfdcc21ab2170f88157eacR159 )

docs/configuration.rst
```diff
@@ -41,6 +41,13 @@ adjustments:
     For an exhaustive list of available settings and their default values,
     refer to `the source code <https://github.com/mozilla-services/cliquet/blob/2.2.0/cliquet/__init__.py#L26-L78>`_.
 
+By default, anyone can read buckets list as well as bucket's
```
leplatrem: By default, nobody can read...

Natim:
Updated with what we've discussed.

PR #120, opened 2015-07-06
Default bucket UUID doesn't have dashes

Natim:
I've seen that default record ID's have got dashes whereas default bucket id doesn't.

Does it makes sense to try to be consistent here?

```
$ http GET http://localhost:8888/v1/buckets/e93a0bb5b7d16d4f9bfd81b6d737271c -v --auth 'mary:marypassword'
{
    "data": {
        "id": "e93a0bb5b7d16d4f9bfd81b6d737271c", 
        "last_modified": 1436191171386
    }, 
    [...]
}
```

### DIFF
--- a/kinto/tests/test_default_bucket.py
+++ b/kinto/tests/test_default_bucket.py
@@ -1,8 +1,10 @@
-import uuid
-from .support import (BaseWebTest, unittest, get_user_headers,
-                      MINIMALIST_RECORD)
+from six import text_type
+from uuid import UUID
 
 from cliquet.utils import hmac_digest
+
+from .support import (BaseWebTest, unittest, get_user_headers,
+                      MINIMALIST_RECORD)
 
 
 class DefaultBucketViewTest(BaseWebTest, unittest.TestCase):
@@ -17,7 +19,7 @@ class DefaultBucketViewTest(BaseWebTest, unittest.TestCase):
         hmac_secret = settings['cliquet.userid_hmac_secret']
         bucket_id = hmac_digest(hmac_secret, self.principal)[:32]
 
-        self.assertEqual(result['data']['id'], bucket_id)
+        self.assertEqual(result['data']['id'], text_type(UUID(bucket_id)))
         self.assertEqual(result['permissions']['write'], [self.principal])
 
     def test_default_bucket_collections_are_automatically_created(self):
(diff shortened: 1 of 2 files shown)


Natim:
Adding the necessary code.

ametaireau:
I don't understand why no tests are impacted by this change.

kinto/tests/test_default_bucket.py
```diff
@@ -15,7 +15,7 @@ def test_default_bucket_exists_and_has_user_id(self):
         result = bucket.json
         settings = self.app.app.registry.settings
         hmac_secret = settings['cliquet.userid_hmac_secret']
-        bucket_id = hmac_digest(hmac_secret, self.principal)[:32]
+        bucket_id = '%s' % UUID(hmac_digest(hmac_secret, self.principal)[:32])
```
ametaireau: nit: I would do `str()` rather than formatting like that.

leplatrem: +1 for ``six.text_type()`` :)

kinto/tests/test_default_bucket.py
```diff
@@ -15,7 +15,7 @@ def test_default_bucket_exists_and_has_user_id(self):
...
 
         self.assertEqual(result['data']['id'], bucket_id)
```
leplatrem: Instead of copying the code of the view to compute the bucket id in the test, it would make more sense to compare to an actual value (by forcing hmac_secret to a specific value for example)

kinto/tests/test_default_bucket.py
```diff
@@ -40,7 +40,7 @@ def test_bucket_id_is_an_uuid(self):
         bucket = self.app.get(self.bucket_url, headers=self.headers)
         bucket_id = bucket.json['data']['id']
         try:
-            uuid.UUID(bucket_id)
+            UUID(bucket_id)
         except ValueError:
             self.fail('bucket_id: %s is not a valid UUID.' % bucket_id)
```
leplatrem: ``def test_bucket_id_has_dashes(self)`` ?

kinto/tests/test_default_bucket.py
Natim: Yes I was using ``'%s' % `` to be compatible with both Py3 and Py2 but ``text_type`` works  :)

kinto/tests/test_default_bucket.py
```diff
@@ -13,11 +11,7 @@ class DefaultBucketViewTest(BaseWebTest, unittest.TestCase):
     def test_default_bucket_exists_and_has_user_id(self):
         bucket = self.app.get(self.bucket_url, headers=self.headers)
         result = bucket.json
-        settings = self.app.app.registry.settings
-        hmac_secret = settings['cliquet.userid_hmac_secret']
-        bucket_id = hmac_digest(hmac_secret, self.principal)[:32]
```
Natim: What I liked with this was the fact to check that a bucket_id is related to the connected user.

PR #131, opened 2015-07-13
bucket_id starting with default may use the wrong route.

Natim:
We should add a test to make sure we can still create bucket ids that starts with default.

### DIFF
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -41,7 +41,9 @@ def main(global_config, **settings):
         kwargs['ignore'] = 'kinto.views.flush'
 
     # Redirect default to the right endpoint
-    config.add_route('default_bucket', '/buckets/default{subpath:.*}')
+    config.add_route('default_bucket_collection',
+                     '/buckets/default/{subpath:.*}')
+    config.add_route('default_bucket', '/buckets/default')
 
     config.scan("kinto.views", **kwargs)
     return config.make_wsgi_app()
--- a/kinto/tests/test_default_bucket.py
+++ b/kinto/tests/test_default_bucket.py
@@ -81,3 +81,22 @@ class DefaultBucketViewTest(BaseWebTest, unittest.TestCase):
         resp = self.app.get(self.collection_url + '/records',
                             headers=headers, status=304)
         self.assertIn('Access-Control-Allow-Origin', resp.headers)
+
+    def test_bucket_id_starting_with_default_can_still_be_created(self):
+        # We need to create the bucket first since it is not the default bucket
+        resp = self.app.put_json(
+            self.bucket_url.replace('default', 'default-1234'),
+            {"data": {}}, headers=self.headers, status=201)
+        bucket_id = resp.json['data']['id']
+        self.assertEquals(bucket_id, 'default-1234')
+
+        # We can then create the collection
+        collection_url = '/buckets/default-1234/collections/default'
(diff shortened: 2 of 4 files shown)


kinto/tests/test_views_buckets.py
```diff
@@ -20,7 +20,8 @@ def setUp(self):
 
     def test_buckets_are_global_to_every_users(self):
         self.app.patch_json(self.record_url,
-                            {'permissions': {'read': [Authenticated]}},
+                            {'data': {},
```
leplatrem: nit: I guess we can omit this (unless you had a reason to specify it)

kinto/tests/test_default_bucket.py
```diff
@@ -81,3 +81,22 @@ def test_cors_headers_are_provided_on_errors(self):
...
+        # We can then create the collection
+        collection_url = '/buckets/default-1234/collections/default'
+        self.app.put_json(
+            collection_url,
+            {"data": {}},
+            headers=self.headers,
+            status=201)
+        resp = self.app.get('/buckets/default-1234/collections',
+                            headers=self.headers)
+        self.assertEquals(resp.json['data'][0]['id'], 'default')
```
leplatrem: nit²: the fact the collection id is ``default`` here does not bring too much value, and it is somehow confusing to understand the test

kinto/__init__.py
```diff
@@ -41,7 +41,9 @@ def main(global_config, **settings):
         kwargs['ignore'] = 'kinto.views.flush'
 
     # Redirect default to the right endpoint
-    config.add_route('default_bucket', '/buckets/default{subpath:.*}')
+    config.add_route('default_bucket_collection',
```
leplatrem: I guess you tried to add the route with the same name :)

kinto/__init__.py
```diff
@@ -41,7 +41,9 @@ def main(global_config, **settings):
...
+                     '/buckets/default/{subpath:.*}')
```
leplatrem: Now that I see this, I wonder: it could have been possible to define an optional group starting with / in the regexp ?
ex. ``/bucket/default{subpath:^(/.*)?$}``

leplatrem:
r+ with nits

I imagine there are several ways to do it, and this seems like a valid one :)

kinto/__init__.py
Natim: I like this idea, we should try it.

kinto/tests/test_default_bucket.py
Natim: Actually it was tasks before and it is here to make sure https://github.com/mozilla-services/kinto/pull/131/files#diff-580dee783aeea3c8487d07890dfdbcb9R76 is tested. One could be tempted to replace default and the test would break if the collection id is replaced.

kinto/tests/test_views_buckets.py
Natim: Actually with your fix on cliquet if you don't provide data you'll get a 400 (data should be present and is not defined)

See https://github.com/mozilla-services/kinto/issues/63

kinto/__init__.py
Natim: Yes it isn't possible.

kinto/__init__.py
Natim: It doesn't work :(

PR #143, opened 2015-07-16
We don't need install-dev for make serve.

Natim:


leplatrem:
I supposed there is some documentation change for this, at least in the contributing section, no ?

Natim:
there are no mention of make serve in the contributing part and this fix a bug in the installation part when using ``make serve`` with a tag.

leplatrem:
r+

PR #151, opened 2015-07-30
Add a section in the configuration about flushing.

ametaireau:
I'm actually not sure if it is the best place to do that. Another location I've been looking for this information was the endpoints, so it might be better to add it there.

In any case, we need to document this :-)

Thoughts?

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -220,6 +220,23 @@ To use a different ini file, the ``KINTO_INI`` environment variable
 should be present with a path to it.
 
 
+Activating the flush endpoint
+=============================
+
+When using Kinto in development mode, it might be helpful to have a way to
+flush all the data that is currently stored in the database.
+
+There is a way to enable this behaviour (it is deactivated by default for
+obvious security reasons). In your `.ini` file:
+
+.. code-block :: ini
+
+    kinto.flush_endpoint_enabled = true
+
+Then, you can issue a `POST` request to the `/__flush__` endpoint to flush all
+the data.
+
+
 .. Storage backend
 .. ===============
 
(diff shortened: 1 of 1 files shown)


docs/configuration.rst
```diff
@@ -220,6 +220,23 @@ To use a different ini file, the ``KINTO_INI`` environment variable
 should be present with a path to it.
 
 
+Activating the flush endpoint
+=============================
+
+When using Kinto in development mode, it might be helpful to have a way to
+flush all the data that is currently stored in the database.
```
Natim: that are currently stored (data is always a plural in english)

Natim:
LGTM with nit. I like it to be in the configuration section.

docs/configuration.rst
ametaireau: ah yes, I'm always doing this error :x

ametaireau:
@leplatrem thoughts?

docs/configuration.rst
ametaireau: I removed the "that is" ;)

docs/configuration.rst
```diff
@@ -220,6 +220,23 @@ To use a different ini file, the ``KINTO_INI`` environment variable
...
+
+There is a way to enable this behaviour (it is deactivated by default for
+obvious security reasons). In your `.ini` file:
+
+.. code-block :: ini
+
+    kinto.flush_endpoint_enabled = true
+
+Then, you can issue a `POST` request to the `/__flush__` endpoint to flush all
+the data.
```
leplatrem: -you

Then, a POST request [...] flushes all the data

leplatrem:
r+ with «you nit»

PR #153, opened 2015-07-30
Allow underscore in Kinto resource names.

Natim:
r? @ametaireau

### DIFF
--- a/kinto/views/__init__.py
+++ b/kinto/views/__init__.py
@@ -6,8 +6,6 @@ from pyramid import httpexceptions
 
 
 class NameGenerator(generators.Generator):
-    regexp = r'^[a-zA-Z0-9][a-zA-Z0-9_-]*$'
-
     def __call__(self):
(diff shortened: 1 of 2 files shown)


kinto/views/__init__.py
```diff
@@ -6,11 +6,14 @@
 
 
 class NameGenerator(generators.Generator):
+    regexp = r'^[a-zA-Z0-9][a-zA-Z0-9_-]*$'
```
ametaireau: So we fix it only for kinto and not for all cliquet resources?

ametaireau:
If we want to allow this only for kinto then I think this PR does what it should, but I wonder if we shouldn't do this directly in cliquet.

Natim:
Tests needs to be restarted after mozilla-services/cliquet#411 merge.

leplatrem:
I don't see any objection in (re)defining the regexp here, along with the generation in __call__. 
It would be even better IMO

Natim:
Restarting tests now that cliquet PR has been merged.

PR #160, opened 2015-08-20
Documentation returns enhancements by @leplatrem

ametaireau:


docs/api/buckets.rst
```diff
@@ -3,8 +3,7 @@
 Buckets
 #######
 
-A bucket is the parent object of collections and groups. It can be viewed as
-a namespace where all collections and groups are stored.
+A bucket is the parent object of collections and groups.
```
ametaireau: Why removing this information? When explaining what a bucket is, I find it very useful to say it's a namespace. Even yesterday, when working with Quentin, the first thing he said was "oh, okay, it's a namespace", without me saying it.

docs/api/index.rst
```diff
@@ -52,15 +55,21 @@ HTTP endpoints
...
 ==============
 
-In addition to the endpoints defined above, some aspects of the API might be
-of interest:
+Full detailed API documentation:
 
 .. toctree::
    :maxdepth: 2
 
+   buckets
```
ametaireau: I put these at the end of the listing on purpose: the table already mentions all the endpoints which are here.

docs/api/index.rst
```diff
@@ -52,15 +55,21 @@ HTTP endpoints
...
+   collections
+   records
+   groups
+   permissions
+   synchronization
```
ametaireau: but let sync :)

docs/api/synchronization.rst
```diff
@@ -0,0 +1,160 @@
...
+When going back online, the set of changes can be sent to the server using a
+:ref:`batch` request.
+
+
+.. _sync-implementations:
+
+Implementations
+===============
+
+The **current implementation of reference** for offline-first records synchronization is
```
ametaireau: reference implementation

docs/api/synchronization.rst
```diff
@@ -0,0 +1,160 @@
...
+#. Use concurrency control to send updates and deletes
+
+
+Polling for changes
+===================
+
+When fetching the collection records with ``?_since=<timestamp>``, *Kinto* returns
+every records that were created/updated/deleted **after** this timestamp.
+
+#. Cold sync → no ``?_since``.
```
ametaireau: Sorry, what?

docs/api/buckets.rst
leplatrem: Should re-explain the concepts in the API page ?

phrawzty: Agree w/ @ametaireau - referring to it as a namespace made it immediately obvious to me as well. :smile:

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+###########
+
+
+Terminology
+===========
+
+.. glossary::
+
+    Object
+        Anything that can be interacted with. Collections, records, buckets, groups
```
phrawzty: Either:
> [...] interacted with. For example, collections, records, buckets, and groups, are all objects.

Or:
> [...] interacted with: collections, records, buckets, and groups.

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+        are all objects.
+
+    Principal
+        An entity that can be authenticated. Principals can be individual people,
+        applications, services, or any group of such things.
+
+    Group
+        A group associates a name to a list of principals.
```
phrawzty: Might we consider "label" instead of "name" in this instance?

docs/api/buckets.rst
ametaireau: Ah, we're on the API page, didn't notice! You're correct then :-) We should link to the concept then to make it easier to follow, probably.

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+
+    Permission
+    Permissions
+        A permission is an action that can be performed on an object.
```
phrawzty: Is a permission the action *itself*, or is it the *right* to perform an action? Consider:
> A permission is the right to perform an action on an object.

This would imply another terminology item for *action*.
> Action
An action is an operation that can be performed on an object. Examples are "read", "write", and "create".

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+        Examples of permissions are «read», «write», or «create».
```
phrawzty: Double-quotes, not guillemets, in English. :wink:

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+
+    ACE
+    Access Control Entity
+        An ACE associates a permission to objects and principals, and allows
+        to describe rules like «*Members of group admins can create collections*».
+        Using #a pseudo-code syntax: ``collections:create = ['group:admins',]``.
+
+    ACL
```
phrawzty: > ACL
Access Control List

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+
+    Permission
+    Permissions
+        A permission is an action that can be performed on an object.
+        Examples of permissions are «read», «write», or «create».
+
+    ACE
+    Access Control Entity
+        An ACE associates a permission to objects and principals, and allows
+        to describe rules like «*Members of group admins can create collections*».
```
phrawzty: Double-quotes, not guillemets, in English. :wink:

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+        Using #a pseudo-code syntax: ``collections:create = ['group:admins',]``.
+
+    ACL
+        A list of ACEs
```
phrawzty: Terminate with a period.

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Objects
+=======
+
+Any set of objects defined in *Kinto* can be given a number of permissions.
```
phrawzty: (imho) This sentence is out of context - consider moving it to the end of this section, or removing it entirely.

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+    |  - id          |
+    |  - data        |
+    |  - permissions |
+    +----------------+
+
+
+Permissions
+===========
+
+On each of these objects, the set of permissions can be:
```
phrawzty: The comma is not necessary here.

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+| **create** | Any listed :term:`principal` can create |
+|            | a new *child object*.                   |
++------------+-----------------------------------------+
+
+Permissions are associated to objects.
+
+In the case of a creation, since an object can have several kinds of children, the
+permission is prefixed (for instance ``groups:create``, ``collections:create``).
+
+The following table lists all permissions that can be associated to each kind
```
phrawzty: > [...] lists all of the permissions that [...]

docs/api/permissions.rst
```diff
@@ -0,0 +1,188 @@
...
+of object.
+
++----------------+------------------------+----------------------------------+
+| Object         | Associated permissions | Description                      |
++================+========================+==================================+
+| Configuration  | ``buckets:create``     | Ability to create new buckets.   |
```
phrawzty: The prefix "Ability to" is implied and can thus be removed from this entire list.

docs/api/permissions.rst
ametaireau: A permission is the action itself. an ACE tells if you have the right or not to perform the action.

ametaireau: We could think about renaming this, but just not now (out of scope for this PR I believe)

PR #168, opened 2015-08-27
Explicitely list the supported backends.

ametaireau:
As suggested by @k4nar

r? @Natim


k4nar:
The ref should do the trick :) .

Natim:
Cool.

PR #158, opened 2015-08-28
Add custom (meta) data on buckets and collections 

Natim:
For some use-cases, it might become useful to be able to store some custom attributes in buckets or collections (e.g. metadata like application version, contact email or whatever).

Currently both Collection and Bucket resources do not define extra fields in their schema, and Cliquet drops unknown fields if not explicitly allowed.

We can either:
* Allow unknown fields in collection and buckets schemas
* Add a specific root level field (along ``data`` and ``permissions``)
* Add a specific field (called ``meta`` for example) in the schema that could receive anything.

The advantage of the latter is that custom fields do not interfere with anything in the protocol, and are trivial to implement. The inconvenient is having to put ``{data: {metadata: {email: "toto@yo.com"}}`` in the payload.

Thoughts ?

### DIFF
--- a/kinto/tests/test_views_collections.py
+++ b/kinto/tests/test_views_collections.py
@@ -64,6 +64,18 @@ class CollectionViewTest(BaseWebTest, unittest.TestCase):
                           headers=self.headers,
                           status=400)
 
+    def test_collections_can_handle_arbitrary_attributes(self):
+        collection = MINIMALIST_COLLECTION.copy()
+        fingerprint = "5866f245a00bb3a39100d31b2f14d453"
+        collection['data'] = {'fingerprint': fingerprint}
+        resp = self.app.put_json('/buckets/beers/collections/barley',
+                                 collection,
+                                 headers=self.headers,
+                                 status=200)
+        data = resp.json['data']
+        self.assertIn('fingerprint', data)
+        self.assertEqual(data['fingerprint'], fingerprint)
+
 
 class CollectionDeletionTest(BaseWebTest, unittest.TestCase):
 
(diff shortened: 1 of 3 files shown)


Natim:
``` http
$ echo '{"data": {"signing": "hash"}}' | http PUT http://localhost:8888/v1/buckets/default/collections/tasks --auth user:pass -v
PUT /v1/buckets/default/collections/tasks HTTP/1.1
Accept: application/json
Accept-Encoding: gzip, deflate
Authorization: Basic dXNlcjpwYXNz
Connection: keep-alive
Content-Length: 30
Host: localhost:8888
User-Agent: HTTPie/0.9.2

{
    "data": {
        "signing": "hash"
    }
}

HTTP/1.1 200 OK
Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Content-Length
Content-Length: 173
Date: Fri, 28 Aug 2015 12:52:35 GMT
Etag: "1440766355961"
Last-Modified: Fri, 28 Aug 2015 12:52:35 GMT
Server: waitress

{
    "data": {
        "id": "tasks", 
        "last_modified": 1440766355961, 
        "signing": "hash"
    }, 
    "permissions": {
        "write": [
            "basicauth:631c2d625ee5726172cf67c6750de10a3e1a04bcd603bc9ad6d6b196fa8257a6"
        ]
    }
}
```

leplatrem:
Note: This was closed, however it was not implemented on buckets.

Intentional ?

ametaireau:
We hadn't identified the need to store data on buckets so far, but we can add it if you have a use case for it?

Natim:
> We hadn't identified the need to store data on buckets so far, but we can add it if you have a use case for it?

Agreed on my side, but good catch.

PR #179, opened 2015-09-04
400 don t provide content length

Natim:
r? @n1k0

### DIFF
--- a/kinto/tests/test_default_bucket.py
+++ b/kinto/tests/test_default_bucket.py
@@ -56,9 +56,10 @@ class DefaultBucketViewTest(BaseWebTest, unittest.TestCase):
         self.app.get(self.collection_url, headers=self.headers)
 
     def test_querystring_parameters_are_taken_into_account(self):
-        self.app.get(self.collection_url + '/records?_since=invalid',
-                     headers=self.headers,
-                     status=400)
+        resp = self.app.get(self.collection_url + '/records?_since=invalid',
+                            headers=self.headers,
+                            status=400)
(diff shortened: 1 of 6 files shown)


kinto/tests/test_default_bucket.py
```diff
@@ -56,9 +56,10 @@ def test_second_call_on_default_bucket_collection_doesnt_raise_a_412(self):
         self.app.get(self.collection_url, headers=self.headers)
 
     def test_querystring_parameters_are_taken_into_account(self):
-        self.app.get(self.collection_url + '/records?_since=invalid',
-                     headers=self.headers,
-                     status=400)
+        resp = self.app.get(self.collection_url + '/records?_since=invalid',
+                            headers=self.headers,
+                            status=400)
+        self.assertIn('Content-Length', resp.headers)
```
n1k0: I wonder if we should create new tests for header checks; here for instance we're mixing querystring and headers. Thoughts?

Natim: I was just trying to debug, but I suppose you are right. I will had a test that does the right thing (checking for Allow-Control-Expose-Headers)

Natim:
I have created https://github.com/mozilla-services/cliquet/pull/436 but I cannot reproduce. Have we got this fixed in last Cliquet version?

leplatrem:
r- : I would not spread the same assertions like this everywhere !

An assertion in a test should reflect its title.

Natim:
Actually we do not plan to land this

PR #180, opened 2015-09-07
Crop SVG logos in overview page

leplatrem:
Before

capture d ecran de 2015-09-07 17 25 00

After

capture d ecran de 2015-09-07 17 24 28


n1k0:
Nit: Could we left-indent a little texts next to logos? Using CSS not SVG, obviously.

ametaireau:
We should add a note which credits the original authors if we remove them from the logos here.

PR #183, opened 2015-09-08
Add Disqus comments to documentation (fixes #159)

### LINKED ISSUE #159: Add Disqus comments on documentation pages ?
I saw that on [webpack documentation](http://webpack.github.io/docs/using-loaders.html#cli).

This could be a good way to get some contextual feedback on the documentation, as well as giving opportunities to build a *community*...

leplatrem:
capture d ecran de 2015-09-08 18 06 47

I don't like the present solution though. It was quick and dirty.

A nicer solution would be to have an extra block in footer instead:
Open a PR there https://github.com/snide/sphinx_rtd_theme/pull/241, let's see how it goes :)

### DIFF
--- a/docs/_templates/footer.html
+++ b/docs/_templates/footer.html
@@ -1,0 +1,45 @@
+<footer>
+  {% if next or prev %}
+    <div class="rst-footer-buttons" role="navigation" aria-label="footer navigation">
+      {% if next %}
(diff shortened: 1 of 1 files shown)


ametaireau:
I've identified a few things where having a custom theme would help us move things forward. For instance, it would help to be able to hide the menu on the index page.

So, we might fork this theme and add the missing bits ourselves.

Natim:
Shall we go with this solution for a start?

ametaireau:
I like it, but I'm not sure it makes sense to have it on the index page of the documentation.

Issue #53, opened 2015-06-10
Give details about 404 when parent object is missing


leplatrem:
With implementation of #48, the message of 404 provides no detail.

leplatrem:
This is currently [not possible with current code in cliquet](https://github.com/mozilla-services/cliquet/blob/2.6.2/cliquet/views/errors.py#L55-L56), where error message is hard-coded. Closing, since it does not add much value...

PR #187, opened 2015-09-10
Remove dead code (ref #53)

leplatrem:
@ametaireau @Natim  r?


Natim:
Looks good to me, I fixed the one flake8 error waiting on travis.

Natim:
But I am wondering if this is dead code or missing tests?

Natim:
I did some manual checking and nothing wrong seems happen.

leplatrem:
There might be some corner cases where this code was useful. But since the tests pass without it, it is dead code :)
If we figure out what were the corner cases, we'll add a test and restore it...

ametaireau:
yes, its the best strategy :-)

PR #188, opened 2015-09-10
Redirect to remove trailing slash (fixes #112)

### LINKED ISSUE #112: URLs ending with a '/' return a 404
Should it allow it like Django does: https://docs.djangoproject.com/en/1.8/ref/settings/#append-slash?

leplatrem:


Natim:
I woul rather put this code in cliquet directly and let people enable it with a config.

Natim:
Supplanted by https://github.com/mozilla-services/cliquet/pull/447

PR #189, opened 2015-09-10
Cache expires control from collection

leplatrem:
**Requires cliquet dev version**

* [x] Example `ini` setting for records
* [x] Control cache expires value from collection *metadata*
* [x] Control cache expires value from bucket ~~metadata (thus all sub-collections)~~ using settings
* [x] Control cache expires value from a collection using settings
* [x] Documentation

### DIFF
--- a/docs/api/collections.rst
+++ b/docs/api/collections.rst
@@ -8,6 +8,9 @@ A collection belongs to a bucket and stores records.
 A collection is a mapping with the following attribute:
 
 * ``schema``: (*optional*) a JSON schema to validate the collection records
+* ``cache_expires``: (*optional*, in seconds) add client cache headers on read-only requests.
+  :ref:`More details...<collection-caching>`
+
 
 .. note::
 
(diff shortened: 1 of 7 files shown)


leplatrem:
@tarekziade was it a specific requirement to be able to set cache duration «globally on bucket» for every underlying collections ?

Otherwise this PR is ready :) @Natim @ametaireau @tarekziade r?

docs/configuration/settings.rst
```diff
@@ -507,6 +507,56 @@ dangerous to leave on by default, and must therefore be enabled explicitly.
 Then, issue a `POST` request to the `/__flush__` endpoint to flush all
 the data.
 
+
+.. _configuration-client-caching:
+
+Client caching
```
ametaireau: This section remembers me that I duplicated the documentation in kinto and formatted it as a table. We might backport this change directly in cliquet instead, otherwise we would need to update the kinto documentation each time a setting is changed on cliquet.

ametaireau: Oh sorry, this is in Kinto! I though we were in Cliquet. My bad.

docs/configuration/settings.rst
```diff
@@ -507,6 +507,56 @@ dangerous to leave on by default, and must therefore be enabled explicitly.
...
+
+    # cliquet.bucket_cache_expires_seconds = 3600
+    # cliquet.collection_cache_expires_seconds = 3600
+    cliquet.record_cache_expires_seconds = 3600
+
+
+If set to ``0`` then the resource becomes uncacheable (``no-cache``).
+
+
+Nginx
```
ametaireau: This probably shouldn't go in this section but in the "recipes" for deployment instead. We could link there from here.

kinto/tests/test_views_collections_cache.py
```diff
@@ -0,0 +1,81 @@
...
+                               MINIMALIST_RECORD,
+                               headers=self.headers)
+        self.record = r.json['data']
+
+    def get_app_settings(self, extra=None):
+        settings = super(ForcedCacheExpiresTest, self).get_app_settings(extra)
+        settings['cliquet.record_cache_expires_seconds'] = 3600
+        return settings
+
+    def test_expire_and_cache_control_are_set(self):
```
ametaireau: nit: `test_expires_and_cache_control_headers_are_set`

kinto/tests/test_views_collections_cache.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.records_url = self.collection_url + '/records'
+
+        resp = self.app.post_json(self.records_url,
+                                  MINIMALIST_RECORD,
+                                  headers=self.headers)
+        self.record = resp.json['data']
+        self.record_url = self.records_url + '/' + self.record['id']
+
+    def test_cache_expires_must_be_an_integer(self):
```
ametaireau: should this go in another test case? The setup isn't necessary here if I read correctly.

kinto/tests/test_views_collections_cache.py
```diff
@@ -0,0 +1,81 @@
...
+        self.app.put_json(self.collection_url,
+                          {'data': {'cache_expires': 'abc'}},
+                          headers=self.headers,
+                          status=400)
+
+    def test_expire_and_cache_control_are_set_on_records(self):
+        r = self.app.get(self.records_url,
+                         headers=self.headers)
+        self.assertIn('Expires', r.headers)
+        self.assertIn('Cache-Control', r.headers)
```
ametaireau: Why don't we test the value of the returned `Cache-Control` header here?

kinto/tests/test_views_collections_cache.py
```diff
@@ -0,0 +1,81 @@
+from .support import (BaseWebTest, unittest, MINIMALIST_RECORD)
+
+
+class ForcedCacheExpiresTest(BaseWebTest, unittest.TestCase):
```
ametaireau: I don't understand what is the fundamental difference between this class and the latter one.

kinto/views/records.py
```diff
@@ -61,3 +61,26 @@ def process_record(self, new, old=None):
...
+        self._handle_cache_expires(self.request.response)
+        return result
+
+    def _handle_cache_expires(self, response):
+        """If the parent collection defines a ``cache_expires`` attribute,
+        then cache-control response headers are sent.
+
+        .. note::
+
+            Those headers are also sent if setting
```
ametaireau: consider: "if the [...] setting is defined"

kinto/tests/test_views_collections_cache.py
leplatrem: Ok thanks I will clarify this. (one tests `cache_expires` from settings, the other one from json api)

ametaireau:
Looks good to me overalls. I'm a bit concerned about some duplication of concerns between the Kinto and the Cliquet documentation, especially about settings, though.

leplatrem:
> duplication of concerns between the Kinto and the Cliquet documentation

For this particular case, I don't see any alternative. Especially because the setting name contains the name of the resource, which is generic in Cliquet, and limited to ``bucket``, ``group``, ``collection``, ``record`` in Kinto.

kinto/tests/test_views_collections_cache.py
leplatrem: Well, it is acceptable that *one* test method doesn't rely on setUp, as long as it is not half of all methods.

ametaireau:
Yes that makes sense. r+ :-)

kinto/tests/test_views_collections_cache.py
ametaireau: It's acceptable but not ideal. Let it be ;)

tarekziade:
> was it a specific requirement to be able to set cache duration «globally on bucket» for every underlying collections ?

I don't think we have this use case, but I guess being able to set default values for the whole server can make sense. As long as we have no cache at all by default

docs/api/collections.rst
```diff
@@ -476,3 +479,76 @@ to an empty mapping.
...
     }
+
+
+.. _collection-caching:
+
+Collection caching
+==================
+
+With the ``cache_expires`` attribute on a collection, it is possible to add client
+cache control response headers for read-only requests.
```
tarekziade: s/read-only/GET/  requests ?

tarekziade:
Yay, looking forward for this feature!

PR #195, opened 2015-09-16
First set of optimizations (fixes #192)

### LINKED ISSUE #192: Usage of default bucket is costly
Because we want underlying bucket and collections objects to be created implicitly when using the `default` bucket, we knew we'd have some overhead. 

However, we implemented the implicit creation using Pyramid *subrequests*, and this is particularly inefficient.

If I disable the implicit creation, a local simulation load test with postgresql goes from `35 RPS` to `53 FPS` (~50% win).

If this default bucket is used in a `/batch`, it means two levels of cascaded subrequests are performed.

> This is problematic especially because default is the main bucket of kinto.js

I could get some major improvement (from 35 to 48 RPS) by replacing the current subrequests with a direct interaction on the storage backend:

```python
# views/buckets.py

def create_bucket(request, bucket_id):
    """Create a bucket if it doesn't exists."""
    bucket_put = (request.method.lower() == 'put' and
                  request.path.endswith('buckets/default'))

    if not bucket_put:
        request.matchdict['id'] = bucket_id
        service = Bucket(request)
        try:
            service.collection.create_record({'id': bucket_id})
        except exceptions.UnicityError:
            pass


def create_collection(request, bucket_id):
    subpath = request.matchdict.get('subpath')
    if subpath and subpath.startswith('collections/'):
        collection_id = subpath.split('/')[1]
        collection_put = (request.method.lower() == 'put' and
                          request.path.endswith(collection_id))
        if not collection_put:
            request.matchdict['bucket_id'] = bucket_id
            request.matchdict['id'] = collection_id
            service = Collection(request)
            try:
                service.collection.create_record({'id': collection_id})
            except exceptions.UnicityError:
                pass
```

But I now wonder why we don't just do something like this:

```python
class Bucket(resource.ProtectedResource):
    def __init__(self, *args, **kwargs):
        super(Bucket, self).__init__(*args, **kwargs)
        if self.record_id == 'default':
            digest = hmac_digest(hmac_secret, request.prefixed_userid)
            self.record_id = text_type(UUID(digest[:32]))
```
I think I asked already, but I can't remember. We should write it in the buckets code :)

leplatrem:
* [x] Simple optimization (goes from 35 to 70 RPS, compared to 75 RPS on classic bucket)

### DIFF
--- a/loadtests/loadtest/__init__.py
+++ b/loadtests/loadtest/__init__.py
@@ -38,6 +38,9 @@ class BaseLoadTest(TestCase):
         self.session.auth = self.auth
         self.session.headers.update({'Content-Type': 'application/json'})
 
+        self.bucket = 'default'
+        self.collection = 'default'
+
         # Keep track of created objects.
         self._collections_created = {}
 
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,7 @@ $(INSTALL_STAMP): $(PYTHON) setup.py
 	$(VENV)/bin/pip install -Ue .
 	touch $(INSTALL_STAMP)
 
-install-postgres: $(INSTALL_STAMP)
-# Since pip install ".[postgresql]" is not enough...
+install-postgres: $(INSTALL_STAMP) $(DEV_STAMP)
 	$(VENV)/bin/pip install "cliquet[postgresql]"
 
 install-dev: $(INSTALL_STAMP) $(DEV_STAMP)
--- a/docs/tutorials/first-steps.rst
+++ b/docs/tutorials/first-steps.rst
@@ -568,15 +568,7 @@ And Bob can also create a task:
 
 .. code-block:: http
 
-    $ echo '{
-        "data": {
-            "description": "Bob new task",
-            "status": "todo"
-        },
-        "permissions": {
-            "read": ["basicauth:a103c2e714a04615783de8a03fef1c7fee221214387dd07993bb9aed1f2f2148"]
-        }
-    }' | \
+    $ echo '{"data": {"description": "Bob new task", "status": "todo"}}' | \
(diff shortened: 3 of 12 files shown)


leplatrem:
Made a wip for the second point. It is tricky, unless, we transform the `default` bucket id into a hash from the authorization code.
https://github.com/Kinto/kinto/commit/cb71b528e014d193c596da87118ae65ffeb812c5

@Natim I know every tests fail, but early feedback is welcomed :) Thanks!

loadtests/loadtest/__init__.py
```diff
@@ -38,6 +38,9 @@ def __init__(self, *args, **kwargs):
         self.session.auth = self.auth
         self.session.headers.update({'Content-Type': 'application/json'})
 
+        self.bucket = 'default'
+        self.collection = 'default'
```
Natim: I wouldn't call it default too.

leplatrem:
Some feedback about this after a couple of hours:

1.  It is not trivial at all.
* The authorization policy must know about the target `bucket_id`
* Using `collection.create_record()` is not enough. Because associated permissions have to be created too (ie. `write` to owner).
* Using a resource instance and calling `put()` is not trivial either. Especially because the request and context have to be prepared accordingly. Whereas we would have to do it during context instanciation (see point 2)
* The cliquet RouteFactory instantiate the current resource to determine if the `PUT` requires write or create. This means that the `default` bucket logics should happen before relying on Cliquet factory.

I tried different things but mainly have been trying to add a custom routefactory in Kinto where we would create the default bucket and collections.

In vain so far :)

leplatrem:
Before: ``35 FPS``

After: 
```
Duration: 60.08 seconds
Hits: 5517
Started: 2015-09-22 13:50:53.098524
Approximate Average RPS: 91
```

PR #197, opened 2015-09-23
Check and create default bucket objects only once per batch (ref #192)

leplatrem:
Requires https://github.com/mozilla-services/cliquet/pull/459

@Natim r?

### DIFF
--- a/kinto/tests/test_default_bucket.py
+++ b/kinto/tests/test_default_bucket.py
@@ -1,3 +1,4 @@
+import mock
 from six import text_type
 from uuid import UUID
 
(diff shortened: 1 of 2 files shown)


kinto/tests/test_default_bucket.py
```diff
@@ -105,3 +106,17 @@ def test_bucket_id_starting_with_default_can_still_be_created(self):
...
+        for i in range(nb_create):
+            request = {'method': 'POST',
+                       'path': self.collection_url + '/records',
+                       'body': MINIMALIST_RECORD}
+            batch['requests'].append(request)
+
+        with mock.patch.object(self.storage, 'create',
+                               wraps=self.storage.create) as patched:
+            self.app.post_json('/batch', batch, headers=self.headers)
+            self.assertEqual(patched.call_count, nb_create + 2)
```
leplatrem: > Note: Without the fix, ``call_count`` equals to 75

PR #202, opened 2015-09-25
Old PostgreSQL version error explaination in the FAQ.

Natim:


phrawzty:
While this answer may be technically correct, it leaves me unsatisfied:
* What is generating this error?
* When might this error message be seen?
* Can the problem be ignored?
* Why does upgrading to pgsql 9.4+ solve the problem?
* Does anything else need to be upgraded?

etc.

Natim:
- What is generating this error? **The PostgreSQL driver that cannot activate the JSONbin functionnality not present with old server version.**
- When might this error message be seen? **When using a PostgreSQL older than 9.4**
- Can the problem be ignored? **Nope the server doesn't run in that case and fails with this error**
- Why does upgrading to pgsql 9.4+ solve the problem? **Because after PostgreSQL9.4 server have got JSONBin support**
- Does anything else need to be upgraded? **No**

Natim:
@phrawzty Based on this new information how shall I rephrase the FAQ answer?

leplatrem:
Also, ideally I wouldn't mix *troubleshooting* and overview questions.

Natim:
>  I wouldn't mix troubleshooting and overview questions.

Yes good point, shall I create troubleshooting file?

phrawzty:
If this is really a FAQ it could link to the (hypothetical) troubleshooting page.

Natim:
Updated with your remarks.

Natim:
Updated with Dan rephrasing. Thanks :+1:

docs/troubleshooting.rst
```diff
@@ -0,0 +1,23 @@
...
+###############
+
+.. _troubleshooting:
+
+We are doing the best we can so you do not have to read this section.
+
+In case you do, please find the common problems you may have below.
+
+If you do not find your problem here, please file an issue or ask for
+help on `#storage <http://chat.mibbit.com/?server=irc.mozilla.org&channel=#storage>`_
```
leplatrem: I would prefer a link to communication channels http://kinto.readthedocs.org/en/latest/contributing.html#communication-channels

docs/troubleshooting.rst
```diff
@@ -0,0 +1,23 @@
...
+In case you do, please find the common problems you may have below.
+
+If you do not find a solution to your problem here, please
+:ref:`ask for help <communication_channels>`!
+
+
+Module object has no attribute 'register_json'
+==============================================
+
+Kinto uses PostgreSQL ``JSONBin`` feature, which allow Kinto to
```
phrawzty: Kinto uses *the* JSONBin feature of PostgreSQL, which is used to store native JSON objects efficiently. Support for this feature was added in PostgreSQL 9.4.

docs/overview.rst
```diff
@@ -222,3 +222,10 @@ Cliquet is a toolkit for designing micro-services. Kinto is a server built
 using that toolkit.
 
 `Read more (in french) about the differences <http://www.servicedenuages.fr/pourquoi-cliquet>`_.
+
+
+When running Kinto I have got an exception, what's wrong?
```
phrawzty: I'm seeing an Exception error - what's wrong?

docs/troubleshooting.rst
```diff
@@ -0,0 +1,23 @@
+Troubleshooting
+###############
+
+.. _troubleshooting:
+
+We are doing the best we can so you do not have to read this section.
+
+In case you do, please find the common problems you may have below.
```
phrawzty: That said, we've included solutions (or at least explanations) for some common problems below.

phrawzty:
`r+`

docs/troubleshooting.rst
```diff
@@ -0,0 +1,24 @@
...
+If you do not find a solution to your problem here, please
+:ref:`ask for help <communication_channels>`!
+
+
+Module object has no attribute 'register_json'
+==============================================
+
+Kinto uses the ``JSONBin`` feature of PostgreSQL, which is used to
+store native ``JSON objects`` efficiently. Support for this feature
+wasa added in PostgreSQL 9.4.
```
jvehent: [wazzaaa](https://www.youtube.com/watch?v=8ejcKyRAHAg)

PR #203, opened 2015-09-25
Check parent collection object only once per batch

leplatrem:
@Natim r?

Before:
```
            self.app.post_json('/batch', batch, headers=self.headers)
>           self.assertEqual(patched.call_count, 2)
E           AssertionError: 25 != 1

```


Natim:
r+

PR #216, opened 2015-10-18
added script to generate kinto.ini

oak11:
sample generated kinto.ini is in kinto/kinto/scripts/kinto.ini
the script for generating it is in the scripts folder.


Natim:
Thank you for starting that. We will review it shortly :)

kinto/scripts/command.py~
```diff
@@ -0,0 +1,85 @@
+import ConfigParser
```
Natim: Please do not commit this kind of files.

ShwetaOak: what should I have done instead?

oak11: what should I have done instead?

Natim: In mean the ~ ones. The .py are fine.

oak11: Oh, I'll keep that in mind. Is there anything else I should change/add?

Natim: Oh I though you would update the PR.

leplatrem:
Thanks for starting to tackle this !

As a general remark, I don't understand why it is necessary to duplicate `__init__.py` and `kinto.ini` into the `scripts` folder.

From what I understand from your script, I would instead:
* use relative path to `kinto/config/kinto.ini`
* prompt the user for a couple of settings (something like `storage engine ? [postgresql]` as a first version) 
* replace the appropriate values of the config
* output the file as you did :)

Tell us what you think !

Enjoy and good luck :)

oak11:
I have committed those changes.

leplatrem:
Great !

I think we can make it even simpler. If we want a command to get started quickly, then we don't need to ask for advanced stuff like uwsgi.

If some settings are defined in the .ini file, then Kinto will use the default values automatically. So as a start, we can start with the bare minimum.
See, for example, the one we use to run the load tests here : https://github.com/Kinto/kinto/blob/master/loadtests/server.ini

I also suggest that we ask for the backend only once, and use the same for all of them. For example, if I say that I want `postgresql` then postgresql is used for `storage` `cache` and `permission`. This way we don't prompt for so many things, and it helps to get started quickly !

In order to guide you, I propose this list of small tasks:

* [ ] Remove the files ending with `~`
* [ ] Remove the duplicated `__init__.py` file in `scripts/` folder
* [ ] Remove prompts for advanced configuration values to keep it to the bare minimum (like the one in load tests)
* [ ] Ask for the backend values only once and use the same in all of them

After that, we will make sure it works, we will improve it to be able to run with commandline arguments, we will document it, add more settings and advanced stuff etc. But first we have to reduce it the bare minimum that works and make it look shiny :)

Good luck !

kinto/scripts/command.py
```diff
@@ -0,0 +1,145 @@
+import ConfigParser
+import sys
+import os,binascii
+
+cfgfile = open("kinto/config/kinto.ini",'w') #enter path of file
```
leplatrem: This mean that you open the file with `write` mode. We probably don't want to erase it since it is under git control. Every time we will pull the changes with git, it will be overwritten.

I suggest that we read it but write to another one (like `config/dev.ini`)

oak11:
I'll make the changes mentioned above. However, I have a few doubts:
-So, uwsgi is default right?I don't ask if uwsgi should be enabled or not? 
-Also, for the backends, I only ask if the user wants postgresql at the backend or not? what happens when another platform is to be used at the backend?
-for multiauth policies, should there be an option for basicauth and fxa auth? If there is an option, fxa_auth will need client id and client secret correct?
-also, what does the socket file contain?

leplatrem:
Here are some answers, I hope my answers will guide you a bit :)

> -So, uwsgi is default right?I don't ask if uwsgi should be enabled or not?

uwsgi is not necessary for Kinto to run locally. It is more like a production setup, when the server should handle hundreds of requests per second. It is out of scope for the first version.
(See the minimalistic ini file I pointed in my previous comment)

> -Also, for the backends, I only ask if the user wants postgresql at the backend or not? what happens when another platform is to be used at the backend?

As a first step, I would suggest something like : 

```
> Which backend to use ? (1 - in-memory, 2 - postgresql, 3 - redis) [1]: 
2

OK, Kinto will be configured with backend «postgresql».
```

This is not an award-winning user interface, but it could work as a first version in my opinion!

> -for multiauth policies, should there be an option for basicauth and fxa auth? If there is an option, fxa_auth will need client id and client secret correct?

Authentication configuration can be complicated. I suggest that you leave it
with the defaults (i.e. basic authentication only). Users that need to configure that will still be able to edit the ini file manually anyway.

However, the client secret is used at several parts of the Kinto code as «a seed» to encrypt some sensitive data. This one **is** mandatory :)

> -also, what does the socket file contain?

Forget about socket, it is only used by uwsgi...

leplatrem:
As a side note...

We may want to investigate the relevance of existing tools that provide such templating/prompts. @ametaireau pointed https://github.com/audreyr/cookiecutter which lead me to something available with Pyramid : `pcreate`. http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/scaffolding.html

The documentation does not say much and it might not be obvious to get started. However, in the end, this could save us some efforts on raw input parsing and configuration templating...

It may be interesting to at least give it a try and look for examples to mimic them !

oak11:
That looks interesting! I am not familiar with it, but, I'll give it a shot. I'll try to create a .ini file using the cookiecutter and pcreate and then commit the changes.

oak11:
I have been able to create a .ini file using cookiecutter. However, I was unable to set the 8 bit password for kinto.userid_hmac_secret which is generated as a random string.
 I tried to parse it in
 hook/pre_gen_project.py , I get that the .ini file is not found.
hook/post_gen_project.py ,  I get the error ConfigParser.NoSectionError: No section: 'authentication_configuration' , which is there in the file.
please help.

leplatrem:
Oh, you went fast !

What feedback do you have on cookiecutter ? Is it nice to use as a user ? You chose it over `pcreate` for a particular reason ?

Don't worry if you don't commit while you do some investigations ! We can take our time and see if it fits well our needs :)

kinto/cookiedozer/hooks/pre_gen_project.py
```diff
@@ -0,0 +1,6 @@
+import ConfigParser
+import os,binascii
+
+cfgfile = open("/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}.ini",'w') #enter path of file
```
leplatrem: The leading `/` in the path here, migth be the source of your trouble

kinto/cookiedozer/hooks/pre_gen_project.py
```diff
@@ -0,0 +1,6 @@
...
+parser = ConfigParser.SafeConfigParser()
+parser.set('authentication_configuration','kinto.userid_hmac_secret', binascii.b2a_hex(os.urandom(8)))
```
leplatrem: In ini files the sections are those strings between brackets (e.g. ``'app:main'``)

oak11:
About cookiecutter, its easy to implement. I selected that because, I found more documentation on it, no particular reason as such. I tried creating a template using pcreate, it didnt work out quite well. It would be great if you could suggest a few more resources/ links on it. I guess I would be able to decide on which is better after giving pcreate a fair chance. 
In the meanwhile, cookiecutter uses Jinja2, how about using Jinja2? However, the parsers will still be used if we use Jinja2.

kinto/cookiedozer/hooks/pre_gen_project.py
oak11: I'll try that

leplatrem:
Great ! Indeed, cookiecutter is not a bad choice. It just installs a fair amount of additionnal tools and libraries where pcreate already comes installed with Kinto.

For example, if you run `pcreate -l` you'll see every available templates.

Some scaffolds examples in the wild:
* https://github.com/mozilla-services/cornice/tree/master/cornice/scaffolds
* https://github.com/Pylons/pyramid/tree/master/pyramid/scaffolds
* https://github.com/Pylons/pyramid_jqm/tree/master/pyramid_jqm
* https://github.com/Pylons/pyramid_jinja2/tree/master/pyramid_jinja2/scaffolds

However, I don't know if any of those will prompt the user for configuration values... I don't even know if pcreate has this feature :))

There are more tools listed here too : https://github.com/audreyr/cookiecutter#similar-projects

oak11:
Thanks! I'll try pcreate

oak11:
created a script to generate a .ini file. It prompts for backend choice. What other prompts need to be added?

almet:
@oak11 I'll close this pull request. Please open a new one with just the changes you want.

PR #217, opened 2015-10-18
Quick pass in configuration for new comers

leplatrem:


Makefile
```diff
@@ -19,6 +19,9 @@ $(INSTALL_STAMP): $(PYTHON) setup.py
 	$(VENV)/bin/pip install -Ue .
 	touch $(INSTALL_STAMP)
 
+install-monitoring: $(INSTALL_STAMP)
```
phrawzty: :+1:

config/kinto.ini
```diff
@@ -1,35 +1,37 @@
...
+
 [app:main]
 use = egg:kinto
-
-pyramid.debug_notfound = true
-
 kinto.http_scheme = http
-kinto.http_host = localhost:8888
-
-# kinto.project_docs = https://kinto.readthedocs.org/
```
phrawzty: Why remove `.project_docs` as an example?

config/kinto.ini
```diff
@@ -1,35 +1,37 @@
...
-# kinto.batch_max_requests = 25
-# kinto.retry_after_seconds = 30
-# kinto.eos =
-
 #
 # Auth configuration.
 #
+# http://kinto.readthedocs.org/en/latest/configuration/settings.html#authentication
+#
 kinto.userid_hmac_secret = b4c96a8692291d88fe5a97dd91846eb4
```
phrawzty: The [documentaton](http://kinto.readthedocs.org/en/latest/configuration/settings.html#authentication) explicitly states
> By default, Kinto doesn’t define a secret for you

Perhaps this value should be empty?

config/kinto.ini
Natim: Because you do not need to change the doc URL from the configuration. The default values is the right one in the scope of kinto.

config/kinto.ini
leplatrem: yes, or something like `replace-by-a-secret`

PR #219, opened 2015-10-19
Do not force host in default configuration (fixes #218)

### LINKED ISSUE #218: Setting kinto.http_host disturbs redirections
Scenario at PyconFr: 

- user A runs kinto on `0.0.0.0:8888` with make serve
- user B tries accesses on user A instance through local Wifi `192.168.1.12:8888`
- user B is redirected to `http://localhost:8888/v1`

Indeed, in default configuration `kinto.http_host` is equal to `localhost:8888`. And we use this setting to force `request.host` if it is defined.

We had done that in cliquet a while ago, because otherwise the pagination  `Next-Page` URL would not contain the port.

I realize this is not acceptable :) Especially because the same information is duplicated several times:

* `[server:main] host = 0.0.0.0 port = 8888`
* `kinto.http_host = 0.0.0.0:8888`
* `[uwsgi] http-socket = 127.0.0.1:8888`

leplatrem:
@Natim r?

### DIFF
--- a/config/kinto.ini
+++ b/config/kinto.ini
@@ -3,10 +3,9 @@ use = egg:waitress#main
 host = 0.0.0.0
 port = 8888
 
+
 [app:main]
 use = egg:kinto
-kinto.http_scheme = http
-kinto.http_host = 0.0.0.0:8888
 
 #
 # Backends.
@@ -75,6 +74,9 @@ multiauth.policies = basicauth
 # kinto.statsd_url = udp://localhost:8125
 # kinto.statsd_prefix = kinto-prod
 
+# kinto.http_scheme = https
+# kinto.http_host = kinto.services.mozilla.com
+
 # kinto.backoff = 10
 # kinto.retry_after_seconds = 30
 # kinto.eos =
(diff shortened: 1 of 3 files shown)


config/kinto.ini
```diff
@@ -75,6 +74,9 @@ multiauth.policies = basicauth
 # kinto.statsd_url = udp://localhost:8125
 # kinto.statsd_prefix = kinto-prod
 
+# kinto.http_scheme = http
+# kinto.http_host = 0.0.0.0:8888
```
Natim: This should never look like that, I should be ``kinto.http_host = localhost:8888``

Natim:
Most of what I think is described in #218. Falling back to WSGI might be a good idea on most setup yes.

leplatrem:
Thanks @Natim for the detailed explanation :) Now that you said it I feel dumb :D 

I updated it !

config/kinto.ini
```diff
@@ -83,7 +85,7 @@ multiauth.policies = basicauth
 # [uwsgi]
 # wsgi-file = app.wsgi
 # enable-threads = true
-# http-socket = 127.0.0.1:8888
+# http-socket = /run/wsgi/kinto.sock
```
Natim: As far as I know it should be 

```
# [uwsgi]
# wsgi-file = app.wsgi
# enable-threads = true
# socket = /run/uwsgi/kinto-latest.sock
# chmod-socket = 666
```

Natim:
r+ with nit on the uwsgi config names.

Natim:
Super

PR #223, opened 2015-10-20
Help for Proxy issues in using http tool

greeshmab:
For proxy issues even after downloading http and configuring it, we are supposed to use proxy everytime we use http command.

### DIFF
--- a/docs/tutorials/first-steps.rst
+++ b/docs/tutorials/first-steps.rst
@@ -29,7 +29,7 @@ We'll start with a relatively simple data model:
 
 Using the `httpie <http://httpie.org>` tool we can post a sample record in the
 ``tasks`` collection:
-For proxy issues we can use  "http --proxy=http:http://usrname:passwrd@host:port COMMAND" instead of "http COMMAND"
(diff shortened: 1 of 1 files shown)


docs/tutorials/first-steps.rst
```diff
@@ -29,6 +29,7 @@ We'll start with a relatively simple data model:
 
 Using the `httpie <http://httpie.org>` tool we can post a sample record in the
 ``tasks`` collection:
+For proxy issues we can use  "http --proxy=http:http://usrname:passwrd@host:port COMMAND" instead of "http COMMAND"
```
Natim: I am sure there should be a way to define the PROXY in the ENV. See https://github.com/jkbrzt/httpie#proxies

Natim: Please consider adding a note stating that:

Please [consider reading httpie documentation](https://github.com/jkbrzt/httpie#proxies) for more information If you need to configure a proxy for instance.

greeshmab: Yes, This would be more helpful

Natim: Please update your ``greeshmab:patch-1`` branch to make this little change so that we can merge it :+1:

leplatrem:
Thanks for this contribution :)

PR #214, opened 2015-10-22
405 error on default bucket doesn't return JSON

Natim:
```http
POST /v1/buckets/default/collections/bookmarks HTTP/1.1
Host: kinto.dev.mozaws.net

{
    "data": {
        "title": "Website", 
        "url": "http://coworking-pb.com/"
    }
}

HTTP/1.1 405 Method Not Allowed
Allow: GET, HEAD, PUT, PATCH, DELETE, OPTIONS
Connection: keep-alive
Content-Length: 77
Content-Type: text/plain; charset=UTF-8
Date: Sun, 18 Oct 2015 08:54:33 GMT
Server: nginx/1.4.6 (Ubuntu)

405 Method Not Allowed

The method POST is not allowed for this resource. 
```

### DIFF
--- a/kinto/tests/test_default_bucket.py
+++ b/kinto/tests/test_default_bucket.py
@@ -2,13 +2,16 @@ import mock
 from six import text_type
 from uuid import UUID
 
+from cliquet.errors import ERRORS
+from cliquet.tests.support import FormattedErrorMixin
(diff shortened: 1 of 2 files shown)


leplatrem:
If tests pass, this look like a sane change! r+

PR #226, opened 2015-10-22
Add a test to handle 400 errors details.

Natim:
r? @n1k0 @leplatrem


Natim:
#214 was too greedy, this is a fix.

PR #242, opened 2015-10-30
Fix kinto with latest cliquet

leplatrem:
* Recreate after delete now returns 201  (<--- @Natim r?)
* Cache headers are only set with anonymous requests


Natim:
I would probably also uncomment the 201 in the tutoriel loadtest.

PR #245, opened 2015-10-30
Cache-Control is now always set on no-cache by default.

Natim:


kinto/tests/test_views_collections_cache.py
```diff
@@ -108,7 +108,8 @@ def test_cache_expires_must_be_an_integer(self):
     def test_expires_and_cache_control_are_not_set_if_authenticated(self):
         r = self.app.get(self.records_url, headers=self.headers)
         self.assertNotIn('Expires', r.headers)
-        self.assertNotIn('Cache-Control', r.headers)
+        self.assertIn('Cache-Control', r.headers)
+        self.assertEqual(r.headers['Cache-Control'], 'no-cache')
```
leplatrem: micronit: since Cache-Control may contain other strings, assertIn() could be safer

PR #232, opened 2015-10-26
kinto migrate and start

s-utsch:
I'm still working on the init command, but I wanted to make the pull request for this contribution to see what kind of changes it'll be necessary to make.

PR #247, opened 2015-11-02
kinto command for init and start

Natim:
Superseed #232

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -1,0 +1,46 @@
+import argparse
+import sys
+from cliquet.scripts import cliquet
+from pyramid.scripts import pserve
+from pyramid.paster import bootstrap
+
+
+def main(args=None):
+        """The main routine."""
+        if args is None:
+                args = sys.argv[1:]
+
+        parser = argparse.ArgumentParser(description="Kinto commands")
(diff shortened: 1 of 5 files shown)


Natim:
@leplatrem final r?

Natim:
We should have the config_file parameter for the other command as well.

kinto/__main__.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if args['which'] == 'init':
+                if(args['config_file'] is None):
+                    env = bootstrap('config/kinto.ini')
+                else:
+                    config_file = format(args['config_file'])
+                    env = bootstrap(config_file)
+        elif args['which'] == 'migrate':
+                env = bootstrap('config/kinto.ini')
+                cliquet.init_schema(env)
```
leplatrem: I would suggest to use a local variable `config_file` everywhere.

From what I read here, `migrate` is always executed with `config/kinto.ini`

leplatrem:
r+

leplatrem:
(*don't forget to update the contributors ;)*)

Natim:
> (don't forget to update the contributors ;))

It can be done at release time isn't it?

PR #251, opened 2015-11-03
Fix 500 if backend fails during implicit creation on default bucket (fixes #236)

### LINKED ISSUE #236: 500 instead of 503 when storage backend fails with default bucket
Currently we do not catch a potential backend error when creating the implicit objects. And we use a reference to `bucket` and `collection` in the `finally` close...

```
  File "/home/travis/build/Kinto/kinto/kinto/views/buckets.py", line 155, in default_bucket
    create_collection(request, bucket_id)
  File "/home/travis/build/Kinto/kinto/kinto/views/buckets.py", line 123, in create_collection
    already_created[collection_uri] = collection
UnboundLocalError: local variable 'collection' referenced before assignment;
```

leplatrem:
@Natim r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,11 @@ This document describes changes between each past release.
 
 - Add a Kinto command for start and migrate operation (#129)
 
+**Bug fixes**
+
+- Fix 500 error response (instead of 503) when storage backend fails during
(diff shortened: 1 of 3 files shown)


leplatrem:
Tests fail because of #252

Natim:
r+

PR #233, opened 2015-10-28
[WIP] Default bucket: redirect instead of running subrequest

leplatrem:
* [x] POC
* [x] 307 Redirect (same verb) after implicit creations of objects in `default` bucket
* [x] Put bucket id+url in hello view
* [ ] ~~Implicit creation of object when `bucket_id` == `request.default_bucket_id` everywhere ?~~ No, so far. Hard to keep it consistent with authorization policies for example
* [x] Design nice layout of hello view fields (`{'user': {'id':.., 'bucket':...)`?)
* [x] Test the whole thing properly
* [x] Update docs
* [x] Add breaking change in protocol section in changelog 
* [ ] Check that kinto.js 1.0 (i.e. fetch) follows the redirections

Basically:

Default bucket can still be manipulated via the default endpoint.

```
http --follow GET http://localhost:8888/v1/buckets/default --auth natim:
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, ETag, Pragma, Cache-Control, Backoff
Content-Length: 187
Content-Type: application/json; charset=UTF-8
Date: Wed, 28 Oct 2015 10:47:23 GMT
Etag: "1446029243668"
Last-Modified: Wed, 28 Oct 2015 10:47:23 GMT
Server: waitress

{
    "data": {
        "id": "b4f6b878-44c5-dbdc-25e0-66e36e1673d0", 
        "last_modified": 1446029243668
    }, 
    "permissions": {
        "write": [
            "basicauth:71c5e3dd315eaf28ec38dbba4e11e67ec34892c7f67c6264506957dc92ae6481"
        ]
    }
}
```

Hello view gives bucket id and url (please `#help` for naming and json layout)

```
http GET http://localhost:8888/v1/  --auth natim:
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Retry-After, Content-Length, Alert, Backoff
Content-Length: 400
Content-Type: application/json; charset=UTF-8
Date: Wed, 28 Oct 2015 10:55:23 GMT
Server: waitress

{
    "bucket": {
        "id": "b4f6b878-44c5-dbdc-25e0-66e36e1673d0", 
        "url": "/v1/buckets/b4f6b878-44c5-dbdc-25e0-66e36e1673d0"
    }, 
    "documentation": "https://kinto.readthedocs.org/", 
    "hello": "kinto", 
    "settings": {
        "batch_max_requests": 25, 
[...]

PR #257, opened 2015-11-04
Add personnal bucket id in hello view

leplatrem:
Imported from #233

### DIFF
--- a/docs/api/buckets.rst
+++ b/docs/api/buckets.rst
@@ -300,3 +300,99 @@ Retrieving all buckets
                 }
             ]
         }
+
+
+.. buckets-default-id:
+
+Personal bucket «default»
+=========================
+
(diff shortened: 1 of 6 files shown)


docs/api/buckets.rst
```diff
@@ -300,3 +300,99 @@ Retrieving all buckets
                 }
             ]
         }
+
+
+.. buckets-default-id:
+
+Personal bucket «default»
```
Natim: We should add a note about implicit creation of both the default bucket and the collection in it, when using default and not when using the bucket id.

Natim: Well it is explained shortly here: https://github.com/Kinto/kinto/pull/257/files#diff-faad629fc7ad04c55c32e64e7a2d19e8R19

Natim:
WFM.

PR #259, opened 2015-11-05
Fix documentation warning and glitches

leplatrem:
I would have liked to enable the `-W` option in Sphinx to turn warnings into errors. But there is one warning I can't figure out how to fix: 
`kinto/docs/api/cliquet/index.rst:: WARNING: document isn't included in any toctree`


leplatrem:
@almet r?

almet:
Cool. For the last warning, we probably should find a way to exclude it from the copy, in here https://github.com/mozilla-services/cliquet/blob/master/cliquet_docs/__init__.py#L6

almet:
r+ otherwise.

PR #261, opened 2015-11-05
Copy kinto.ini from sample

leplatrem:
Just trying something... Thoughts ?


kinto/__main__.py
```diff
@@ -34,6 +37,10 @@ def main(args=None):
 
     args = vars(parser.parse_args())
     config_file = args['ini_file']
+
+    if not os.path.exists(config_file):
+        shutil.copy('config/kinto.ini.sample', config_file)
```
Natim: :+1:

Natim: Shall we do that in the init command instead? and tell the person that she need to start ``kinto init`` first?

In the make file we can also add the CONFIG_FILE as a dependency and run ``kinto init`` automatically if the file doesn't exists.

docs/tutorials/run-kinto.rst
```diff
@@ -73,8 +73,8 @@ Then install the package using the default configuration:
 ::
 
     pip install kinto
-    wget https://raw.githubusercontent.com/Kinto/kinto/master/config/kinto.ini
-    pserve kinto.ini
+    wget https://raw.githubusercontent.com/Kinto/kinto/master/config/kinto.ini.sample -O kinto.ini
+    kinto --ini kinto.ini start
```
leplatrem: We could just put the sample in the egg and suggest to run `kinto start` only :)

kinto/__main__.py
leplatrem: good idead!

docs/tutorials/run-kinto.rst
Natim: I would add the sample in the egg anyway + @oak11 work on the init command to do something like:

    kinto init
    kinto start

leplatrem:
This was implemented using template in #278

PR #266, opened 2015-11-07
script to generate .ini file

oak11:
please check the "prompt for backend added" commit.


oak11:
added prompt for backend choice, what other prompts should I add?

almet:
Hey. Thanks for the new pull request. However, this includes changes that aren't useful. I suggest you start from scratch, from a new branch, and you import the changes you want there.

There are multiple things you need to do, since you commited these changes in your master branch (doing so is considered a bad practice as it prevents you from working on multiple patches at the same time, for instance).

So, I would recommend you to remove your clone from github and start again, with a fresh clone (but first, save your work by copying it somewhere else, for instance!)

Then, we don't want to have a completely empty repository with just your files: we want to have it added. I'll let you figure out how to do so. In the meantime, I'll close the pull request here.

From a fresh clone:
```
$ git checkout master
$ git checkout -b generate-ini-file
```

Then add your files, and then:

```
$ git add newfile.py
$ git commit -am "Commit message"
$ git push origin generate-ini-file
```

And then issue a pull request.

Issue #238, opened 2015-10-30
Document DELETE /records


Natim:
As state by @stuartphilp the endpoint is not documented here: http://kinto.readthedocs.org/en/latest/api/records.html

leplatrem:
It is here but totally unclear indeed http://kinto.readthedocs.org/en/latest/api/cliquet/resource.html#delete-collection

PR #268, opened 2015-11-10
Fix Document Delete /records

jschneiders:


leplatrem:
Hi Janine,
Are trying to fix #238 ?
I don't understand what you did here. 238 should just consists in mentioning DELETE operation for the list of records, that's just it :)

leplatrem:
Thanks again for your contribution!
I will close this for now. But don't hesitate to open another one with just the missing line in docs!

PR #272, opened 2015-11-13
Add Python 3.5 to TravisCI

leplatrem:
@Natim r?


Natim:
Should we let both 3.4 and 3.5?

leplatrem:
Added as suggested :)

PR #273, opened 2015-11-13
Fix Docker with PostgreSQL and Docker Compose

leplatrem:
Using docker compose is now a lot lighter, it does not build the container locally, it just pulls it from Docker hub


Natim:
r+

PR #276, opened 2015-11-17
Document how to configure Kinto behind a CDN

Natim:
r? @leplatrem 
feedback? @Micheletto

### DIFF
--- a/docs/configuration/production.rst
+++ b/docs/configuration/production.rst
@@ -55,6 +55,19 @@ Beware that if you do so, everyone will be able to list bucket
 information (including user's personal buckets).
 
 
+Handling CDN
+------------
+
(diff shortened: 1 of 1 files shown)


Micheletto:
LGTM

docs/configuration/production.rst
```diff
@@ -55,6 +55,19 @@ Beware that if you do so, everyone will be able to list bucket
...
 
 
+Handling CDN
+------------
+
+If you want to put your kinto behind a CDN you must make sure to define the right host or you will leak the main server host.
+
+.. code-block:: ini
+
+    kinto.http_host = <cdn_hostname>
```
leplatrem: we could probably put a sample hostname here (e.g.  `cdn.firefox.com`). It would help to know if the value should contain protocol scheme or not

PR #278, opened 2015-11-18
Kinto init command

Natim:
Superseeded #271


kinto/config/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    values['secret'] = binascii.b2a_hex(os.urandom(32))
+
+    backend = input("Which backend to use? "
+                    "(1 - postgresql, 2 - redis, default - memory) ").strip()
+
+    if backend == '1':
+        # Postgresql configuration
+        values['storage_backend'] = "cliquet.storage.postgresql"
+        values['storage_url'] = ("postgres://postgres:postgres@localhost/"
+                                 "postgres")
```
leplatrem: I would put this URI in a variable instead of repeating it 3 times

kinto/config/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+        with codecs.open(destination, 'w+', encoding='utf-8') as output:
+            output.write(rendered)
+
+
+def init(config_file):
+    values = {}
+    values['secret'] = binascii.b2a_hex(os.urandom(32))
+
+    backend = input("Which backend to use? "
+                    "(1 - postgresql, 2 - redis, default - memory) ").strip()
```
leplatrem: What happen if user types 3 :) ?

kinto/config/__init__.py
Natim: It is configuring the memory backend.

Natim: Ok

leplatrem: Yes I can read that. But we don't complain?

kinto/config/__init__.py
leplatrem: We don't complain?

Natim: The idea was to default to memory but we can discuss this if you think otherwise.

leplatrem:
Ok great ! Let's go through some missing parts:

* [ ] Add a rule in Makefile to run init when config file is missing (like in #261)
* [ ] Add default config/kinto.ini to .gitignore
* [ ] Update documentation to remove wget of remote ini files
* [ ] Update quickstart to mention this command
* [ ] Remove `cp config/kinto.ini config/dev.ini` from documentation
* [ ] Warn (or create backup or whatever) when destination file already exists

kinto/config/__init__.py
leplatrem: default when empty string is seized yes, but surrounding the input with a `while True: ... if backend in ('', '1', '2'): break` would not change much

leplatrem:
r+

PR #281, opened 2015-11-19
Fix doc about default storage backend setting

leplatrem:
@almet r?


Natim:
Why don't you change the two others too?

leplatrem:
Oops! Done. Thanks!

Natim:
LGTM :+1:

PR #298, opened 2015-11-24
Fix DockerFile with default configuration (fixes #296)

### LINKED ISSUE #296: Docker doesn't start anymore because config/kinto.ini doesn't exists.
v37hk

leplatrem:
@Natim r?

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 import argparse
 import os
 import sys
+from six.moves import input
 from cliquet.scripts import cliquet
 from pyramid.scripts import pserve
 from pyramid.paster import bootstrap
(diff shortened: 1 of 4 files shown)


kinto/__main__.py
```diff
@@ -40,12 +46,25 @@ def main(args=None):
...
     if args['which'] == 'init':
-        if not os.path.exists(config_file):
-            init(config_file)
-        else:
+        if os.path.exists(config_file):
             print("%s already exist." % config_file, file=sys.stderr)
             sys.exit(1)
 
+        backend = args['backend']
+        if not backend:
```
Natim: I think I would keep that check in the input() methods when passing backend=None to the method.

leplatrem: you mean prompting in the `init()` method ?

I found it more consistent to have interaction with user along the CLI code rather than along templating.

Do you have any strong argument?

Natim:
r+

PR #300, opened 2015-11-26
Versioning doc

tarekziade:
Stuff to do 

- [x] Make sure the whole team agrees on the document
- [x] Change the root view accordingly by renaming the field  (happenin also in cliquet)
- [x] Finalize the documentation and validate the incrementation scenarios for Kinto.js

For the incrementation scenarios for Kinto.js, the API version(s) which are compatible with the client, will be hard coded into the kinto.js release.


docs/api/1.0/index.rst
```diff
@@ -0,0 +1,96 @@
+.. _kinto-api-endpoints:
```
almet: I don't think we need the 1.0 folder here, since the docs are versionned.

tarekziade: It seems to me that the docs are not versioned on each HTTP API increment, but on each Python package release, which is different.

So how can we have document with:

- a list of all API versions ?
- for each new API version a specific document with the diff with the previous API version

docs/api/1.0/index.rst
```diff
@@ -0,0 +1,96 @@
...
+Full reference
+==============
+
+Full detailed API documentation:
+
+.. toctree::
+   :maxdepth: 2
+
+   cliquet/utilities
+   cliquet/versioning
```
leplatrem: Careful with this then, there is now two documents about versioning. I suggest you get rid of this one.

docs/api/1.0/index.rst
leplatrem: nit: If we plan on having `.. added-on:: 1.X` in the doc, then the folder could be named `1.X` instead of `1.0` :)

tarekziade: good point! :+1:

docs/api/1.0/index.rst
tarekziade: sure.

docs/glossary.rst
```diff
@@ -0,0 +1,20 @@
+.. _kinto-glossary:
+
+Glossary
+########
+
+
+.. glossary::
+
+    Cliquet Protocol
+            A system of rules that explains the way to interact with Kinto,
```
Natim: the way to interact with Kinto **resources**

docs/glossary.rst
```diff
@@ -0,0 +1,20 @@
...
+            and how its data is organized.
+
+    HTTP API
+            The definition of a set of HTTP request messages along with the
```
Natim: I don't understand this definition, I would say the list of HTTP resources endpoint defined for Kinto.

docs/versioning.rst
```diff
@@ -0,0 +1,37 @@
...
+
+
+Kinto implements an :term:`HTTP API` that can be consumed by clients like
+`Kinto.js <http://kintojs.readthedocs.org/>`_.
+
+The :term:`HTTP API` is described :ref:`here <kinto-api-endpoints>` and is
+subject to changes. It follows the :term:`Cliquet Protocol`.
+
+When the :term:`HTTP API` is changed, its version is incremented.
+The :term:`HTTP API` version follows a :term:`Semantic Versioning`
```
Natim: Shouldn't we have the PATCH release number if we say we are following Semantic versioning for our HTTP API?

Natim: Maybe we can leave it without PATCH and say that every api change is a minor change but we will not have patch changes at the API level.

docs/glossary.rst
n1k0: Yeah, *request messages* is vague.

docs/versioning.rst
n1k0: If we follow semver, adding a feature without breaking BC still implies bumping a minor - which might happening very often. Are we okay with this?

docs/versioning.rst
```diff
@@ -0,0 +1,37 @@
...
+pattern and uses this rule to be incremented:
+
+1. any change to the :term:`HTTP API` that is backward compatible increments
+   the **MINOR** number, and the modification in the documentation should reflect
+   this with a header like "Added in 1.x".
```
n1k0: You mean a *heading* in the docs, right, not a HTTP header?

docs/versioning.rst
```diff
@@ -0,0 +1,37 @@
...
+2. any change to the :term:`HTTP API` that is backward incompatible increments
+   the **MAJOR** number, and the differences are summarized at the begining of
+   the documentation, a new document for that **MAJOR** version is created.
+
+We want to avoid **MAJOR** changes as much as possible in the future, and stick
+with 1.x as long as we can.
+
+A client that interacts with a Kinto server can query the server to know what
+is its :term:`HTTP API` version. This is done with a query on the root view,
+as described in :ref:`the root API description <api-utilities>`.
```
n1k0: I'd add *it's up to clients to ensure they fully support the HTTP API version provided by the server*

docs/api/versioning.rst
```diff
@@ -0,0 +1,37 @@
...
+   the documentation, a new document for that **MAJOR** version is created.
+
+We want to avoid **MAJOR** changes as much as possible in the future, and stick
+with 1.x as long as we can.
+
+A client that interacts with a Kinto server can query the server to know what
+is its :term:`HTTP API` version. This is done with a query on the root view,
+as described in :ref:`the root API description <api-utilities>`.
+
+The response returns an **http_api_version** key that contains the
```
n1k0: Nit: the JSON response

docs/api/versioning.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Kinto implements an :term:`HTTP API` that can be consumed by clients like
+`Kinto.js <http://kintojs.readthedocs.org/>`_.
+
+The :term:`HTTP API` is described :ref:`here <kinto-api-endpoints>` and is
+subject to changes. It follows the :term:`Cliquet Protocol`.
+
+When the :term:`HTTP API` is changed, its version is incremented.
+The :term:`HTTP API` version follows a :term:`Semantic Versioning`
+pattern and uses this rule to be incremented:
```
n1k0: We should add a note that we'll observe semver only for major and minor, not patch.

n1k0:
LGTM. Much clearer.

docs/glossary.rst
tarekziade: I thought it was more than that. For instance "GET /" is part of the protocol, as defined in Cliquet docs, and it's not a Resource (according to the definition in the Cliquet glossary)

docs/glossary.rst
tarekziade: This is the definition from Wikipedia = https://en.wikipedia.org/wiki/Application_programming_interface#Web_APIs

I am ok to change it, but in "HTTP resources endpoint" , I don't understand what the word resource means.

docs/glossary.rst
Natim: resources and utility endpoints ?

docs/glossary.rst
Natim: This: https://en.wikipedia.org/wiki/Web_resource

docs/glossary.rst
tarekziade: why not just "endpoints" ?

docs/glossary.rst
leplatrem: In the docs, we use several time the word *enpoint* (=verb + URL). It would be great to mention it somewhere around here

tarekziade: This is the definition of.. Web Resource ;)  For the definition of HTTP API it's closer to https://en.wikipedia.org/wiki/Web_API.

So what about:

multiple publicly exposed endpoints that accept HTTP requests and respond with the requested data, in the form of JSON, to interact with the data held in the Kinto Server

leplatrem:
Works for me!

docs/glossary.rst
leplatrem: I'm ok with that.

I wish we'd have *endpoint* like in Cliquet glossary (http://cliquet.readthedocs.org/en/latest/reference/glossary.html)

docs/api/versioning.rst
tarekziade: done

docs/glossary.rst
```diff
@@ -0,0 +1,19 @@
...
+
+Glossary
+########
+
+
+.. glossary::
+
+    Cliquet Protocol
+            A system of rules that explains the way to interact with Kinto
+            :term:`endpoints`, and how Kinto data is organized.
```
leplatrem: I think it *endpoints* should be defined in the glossary too then :)

leplatrem:
r+ but better if sphinx warnings can be fixed :)

docs/glossary.rst
tarekziade: It's already in the Cliquet glossary. Ideally we should have a unified glossary across the two docs. I guess we can import it like we do for other cliquet docs

tarekziade:
I am now importing cliquet glossary so the Kinto doc can use the terms described there. The Kinto has its own glossary for terms that are Kinto-specific. 

@leplatrem let me know if this works for you

docs/conf.py
```diff
@@ -28,10 +28,25 @@
 # it
 
 # Copy the docs from Cliquet inside these ones.
-destination = os.path.join(__HERE__, 'api', 'cliquet')
+# XXX how do we copy on 2.x ?
+destination = os.path.join(__HERE__, 'api', '1.x', 'cliquet')
```
Natim: Do you think a version of kinto will eventually implement both API version?

tarekziade: You mean 1.x and the 2.x line ? I don't see why/how

tarekziade: But the docs will need to show all API versions

Natim: I was thinking that the package documentation should only show the API that is implemented.
Note that we are displaying all version of the documentation in ReadTheDocs, so even if a documentation is only showing a version of the API, it will still be possible to read all version of the API.

tarekziade: That's the problem with having the HTTP API documentation inside a Python package. The HTTP API documentation should be separated from the implementation. I understand that it's easier to maintain if it's in the Kinto repo, but the kinto package releases should not interfere with the HTTP API versions.

The target readership does not care about the implementation. They want the API doc, and they want to be able to see how 1.1 is different from 2.3 for instance.

Natim: I was thinking that we could have the API implemented in Kinto package documented in the package documentation and only that. If we want to compare different version of APIs, we can do it by comparing documentation packages version as well. (Note that ReadTheDocs keep publishing all the different version of the Kinto packages)

We can probably also list the different APIs in another documentation repository (which could import them from the Kinto packages)

PR #305, opened 2015-11-30
use cliquet versionning doc to avoid duplicate

tarekziade:
This document  :  https://github.com/mozilla-services/cliquet/blob/master/cliquet_docs/api/versioning.rst

And this document : https://github.com/Kinto/kinto/blob/master/docs/api/versioning.rst

..are duplicates.  Let's have Kinto pull the Cliquet one so we avoid duplication (which always leads to problems)


leplatrem:
We had doubts about that. Especially because Kinto and Kinto.js are mentionned explicitly here, whereas in Cliquet we tried to keep a generic doc that can work for any project.

But I'm ok with that if you are too :) 

r+

PR #312, opened 2015-12-03
Mention how to restrict private URLs with NGinx (fixes #250)

### LINKED ISSUE #250: Mention that __heartbeat__ should not be served publicly in production
Since it requires no authentication and performs write operations on backends, it should be available publicly.

leplatrem:
@Natim r?


Natim:
It should work yes :)

PR #329, opened 2015-12-15
Fix configuration

oak11:
@almet @Natim  @leplatrem


kinto/__main__.py
```diff
@@ -9,7 +9,7 @@
 
 from kinto.config import init
 
-CONFIG_FILE = 'config/kinto.ini'
+CONFIG_FILE = 'kinto/config/kinto.ini'
```
almet: Why do you need to specify the kinto prefix here?

oak11: so that the kinto.ini file is created within the kinto/config folder( where kinto.tpl is present) . If only config/kinto.ini is written, it creates a new config folder in the root directory and kinto.ini is created inside that.

almet: I believe that's actually the behaviour we want: the "kinto" folder is the one where all the kinto files are contained, and you shouldn't temper it. Rather, creating a "config" folder where you run the command seems to make more sense.

oak11: oh, okay I'll change it back to config/kinto.ini

kinto/__main__.py
```diff
@@ -49,7 +49,7 @@ def main(args=None):
         if os.path.exists(config_file):
             print("%s already exist." % config_file, file=sys.stderr)
             sys.exit(1)
-
+
```
almet: Please remove this empty line :)

oak11: done

almet:
Okay, great. Now we have to write the tests!

kinto/config/__init__.py
```diff
@@ -1,46 +1,46 @@
...
 HERE = os.path.abspath(os.path.dirname(__file__))
 
+def render_template(template, destination, values):
+   
+   here = os.path.abspath(os.path.dirname(__file__))
+   template = os.path.join(here, 'kinto.tpl') 
+   destination = os.path.join(here, 'kinto.ini')
 
-def render_template(template, destination, **kwargs):
-    template = os.path.join(HERE, template)
```
Natim: Please keep this version with the makedirs and using the HERE variable.

kinto/config/__init__.py
```diff
@@ -1,46 +1,46 @@
...
 
     else:
-        values['storage_url'] = ''
-        values['cache_url'] = ''
-        values['permission_url'] = ''
+        values.update({'storage_url' : '',
+                     'cache_url' : '',
+                     'permission_url' : ''})
 
-    render_template("kinto.tpl", config_file, **values)
```
Natim: Why is this change necessary?

oak11:
Cool, tell me the locations to put tests for kinto/config/__init_.py  and  kinto/__main__.py  I'll write the tests and commit them.

kinto/config/__init__.py
almet: We thought it was actually better to pass the variables as a dict rather than as parameters.

almet:
Tests for config should go in `kinto/tests/test_config.py` and don't bother about having tests for `kinto.__main__` for now.

kinto/config/__init__.py
Natim: I don't agree, why would I write:

    render_template("kinto.tpl", config_file, {"secret": "abc", "storage_backend": "bla", ...})

Isn't it better to write:

    render_template("kinto.tpl", config_file, secret="abc", storage_backend="bla", ...)

kinto/config/__init__.py
almet: yes, the only think you might want to change here is the mode which should be "w" instead of "w+".

kinto/config/__init__.py
almet: Both solutions work for me, and I don't consider this really important. I've see with Shweta that the version without the kwargs would make more sense (be more explicit), but we can reverse it to the previous behaviour if you want.

kinto/tests/test_config.py
```diff
@@ -0,0 +1,25 @@
+from ..config import render_template as origin
```
almet: You need to import with absolute locations. `from kinto.config import render_template`. You don't need to import it as origin.

kinto/tests/test_config.py
```diff
@@ -0,0 +1,25 @@
...
+from .support import unittest
+
+class ConfigTest(unittest.TestCase):
+
+#to check if values dict is not empty
```
almet: Please define the comments from inside the method, and not on top of it. eg.

```python

def test_values_is_not_empty(self):
    # Your comment here.
```

That being said, your comment here just repeats the name of the method, so your don't need to add one.

kinto/tests/test_config.py
```diff
@@ -0,0 +1,25 @@
...
+    def test_values_is_not_empty(self):
+        self.assertNotEqual(origin.values, {})
```
almet: It's not clear to me what the intent is here. What are you trying to test?

oak11: I am trying to see if values has elements. If it has elements (is not empty) then it is correct- as it will have elements only if it has passed through the if statements and init function. 
If it is empty (has no elements), there was a failure somewhere.

Natim:
Superseeded by #349

Issue #321, opened 2015-12-11
Move rapidly changing parts of docs to the wiki


leplatrem:
For example platform specific installation, list of demos etc...

Suggested by @jwhitlock :)

PR #332, opened 2015-12-15
Merge app example page (ref #321)

leplatrem:
@Natim r?

### DIFF
--- a/docs/tutorials/app-examples.rst
+++ b/docs/tutorials/app-examples.rst
@@ -1,10 +1,9 @@
+.. _app-examples:
+
 Application examples
 ####################
 
-Existing applications
-=====================
+Demos
+=====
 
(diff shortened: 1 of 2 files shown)


docs/tutorials/app-examples.rst
```diff
@@ -1,10 +1,9 @@
+.. _app-examples:
+
 Application examples
 ####################
 
-Existing applications
-=====================
-
-* :github:`Kinto React Boilerplate <Kinto/kinto-react-boilerplate>`: A starter kit for JavaScript application.
-* :github:`Routina <leplatrem/Routina>`: An offline-first demo Web app.
```
Natim: Does it means we should not have some here?

leplatrem: I first wanted to keep a list of *featured apps*, but then thought it could conceal the wiki page...

PR #334, opened 2015-12-15
Remove the link to Twitter, it's almost useless.

almet:


Natim:
I think it good to tell people to use the #kinto hashtag when they want to speak about it, isn't it?

almet:
Well in practice it's useless to me, since I can't subscribe to it.

Natim:
Then it might be wise to create a kinto-storage account?

Natim:
With some Twitter clients you can :)

PR #340, opened 2015-12-16
Bump protocol version to 1.1 because of protocol changes in release 1.9

leplatrem:
Discussed with @almet


Natim:
We should probably update the doc with the API changelog.

leplatrem:
> API changelog.

:+1:

almet:
Now, that's a good idea!

leplatrem:
Updated

PR #341, opened 2015-12-16
Do not require cliquet master branch in dev

leplatrem:
As discussed with @Natim @almet


Natim:
Even more clever than I planned to do :+1:

almet:
PPPerfect!

PR #342, opened 2015-12-16
Move default values to dedicated column in docs (fixes #255)

### LINKED ISSUE #255: Add column for default values in settings documentation
We should add the default values in the config doc. e.g.: what's the default size of the pool?

leplatrem:
@almet r?

### DIFF
--- a/docs/_static/theme_overrides.css
+++ b/docs/_static/theme_overrides.css
@@ -48,3 +48,17 @@
 #overview #key-features .wy-table-responsive:nth-of-type(1) table td {
     font-size: 1.2em;
 }
+
+/* tweak to limit width of column for settings default values */
+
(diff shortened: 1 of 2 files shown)


Natim:
Yes much better :+1:

almet:
Thanks for doing this. I'm actually not sure about the results, when looking at them.

small

In this example, the table is almost unreadable...

almet:
Also, I wonder if the default value should come at the end (after the explanation and not before) ?

leplatrem:
That's not so complicated with modern editors...
screencast_17-12-2015_10 08 26

But I think I prefer the default next to the setting name.

I will try to do something in CSS to set a max-width on the second column

almet:
Cool :)

leplatrem:
capture d ecran de 2015-12-17 16-17-45

Issue #337, opened 2015-12-16
kinto init fails when the config folder already exists


leplatrem:
**Steps to reproduce**

First, create the default file. The parent folder `config/` will be created.
```
$ kinto --backend=memory init
```

If ran again, the command exits gracefully:
```
$ kinto --backend=memory init
config/kinto.ini already exist.
```

But if we delete the `kinto.ini` file and leave the `config/` folder, or if we specify another filename in the `config/` folder, then the commands fails:

```
$ kinto --backend=memory --ini=config/dev.ini init
Traceback (most recent call last):
  File "/home/mathieu/Code/Mozilla/kinto/.venv/bin/kinto", line 9, in <module>
    load_entry_point('kinto', 'console_scripts', 'kinto')()
  File "/home/mathieu/Code/Mozilla/kinto/kinto/__main__.py", line 66, in main
    init(config_file, backend)
  File "/home/mathieu/Code/Mozilla/kinto/kinto/config/__init__.py", line 46, in init
    render_template("kinto.tpl", config_file, **values)
  File "/home/mathieu/Code/Mozilla/kinto/kinto/config/__init__.py", line 12, in render_template
    os.makedirs(folder)
  File "/home/mathieu/Code/Mozilla/kinto/.venv/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: 'config'
```

Natim:
This was fixed by #349

PR #346, opened 2015-12-16
bug fix for: kinto init fails when the config folder already exists

oak11:
@almet  @leplatrem  @Natim  this solved it for me. For issue: https://github.com/Kinto/kinto/issues/337

### DIFF
--- a/kinto/config/__init__.py
+++ b/kinto/config/__init__.py
@@ -9,10 +9,7 @@ def render_template(template, destination, **kwargs):
     template = os.path.join(HERE, template)
 
     folder = os.path.dirname(destination)
-    if os.path.exists(folder):
-            print("%s already exists. kinto.ini will be created" % folder)
-         
-    else:
+    if os.path.exists(folder) == False:
             os.makedirs(folder)
 
     with codecs.open(template, 'r', encoding='utf-8') as f:
--- a/kinto/config/__init__.py
+++ b/kinto/config/__init__.py
@@ -9,8 +9,8 @@ def render_template(template, destination, **kwargs):
     template = os.path.join(HERE, template)
 
     folder = os.path.dirname(destination)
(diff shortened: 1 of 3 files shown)


kinto/config/__init__.py
```diff
@@ -9,7 +9,11 @@ def render_template(template, destination, **kwargs):
     template = os.path.join(HERE, template)
 
     folder = os.path.dirname(destination)
-    os.makedirs(folder)
+    if os.path.exists(folder):
+            print("%s already exists. kinto.ini will be created" % folder)
```
almet: We probably don't want to print out here. Just let the "if not os.path.exists(folder)", and add a test :-)

kinto/config/__init__.py
```diff
@@ -9,7 +9,8 @@ def render_template(template, destination, **kwargs):
     template = os.path.join(HERE, template)
 
     folder = os.path.dirname(destination)
-    os.makedirs(folder)
+    if os.path.exists(folder) == False:
+            os.makedirs(folder)
```
almet: Is there too much spaces there? (should be 4 I think)
+            os.makedirs(folder)
 Styling with Markdown is supported Write Preview

Is there too much spaces there? (should be 4 I think)
Attach files by dragging & dropping,  Sélect. fichiers selecting them, or pasting from the clipboard.
Comment Cancel

kinto/config/__init__.py
```diff
@@ -9,7 +9,8 @@ def render_template(template, destination, **kwargs):
     template = os.path.join(HERE, template)
 
     folder = os.path.dirname(destination)
-    os.makedirs(folder)
+    if os.path.exists(folder) == False:
```
leplatrem: We usually don't compare with boolean values, but use negation like `if not os.path.exists(folder)`

PR #349, opened 2015-12-18
bug fix for: kinto init fails when the config folder already exists

oak11:
@Natim  @almet  @leplatrem  please check and point out the mistakes here

### DIFF
--- a/kinto/config/__init__.py
+++ b/kinto/config/__init__.py
@@ -6,9 +6,7 @@ HERE = os.path.abspath(os.path.dirname(__file__))
 
 
 def render_template(template, destination, **kwargs):
-    template = os.path.join(HERE, template)
-
-   
+    template = os.path.join(HERE, template)  
     folder = os.path.dirname(destination)
    
     if not os.path.exists(folder):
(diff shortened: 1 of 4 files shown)


kinto/tests/test_config.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ConfigTest(unittest.TestCase):
+
+#to check if values dict is not empty    
+        
+#To check if rendered template generates a file
+   
+    def test_init(self):
+        
+        destination = os.path.join(os.path.abspath(os.path.dirname(__file__)), test.ini)
```
Natim: You probably want to use a temporary file. [Using mktemp](https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp)

kinto/tests/test_config.py
```diff
@@ -0,0 +1,15 @@
...
+        
+        self.assertEqual(init(destination,postgresql),kinto1)
```
Natim: What is kinto1 here?

kinto/config/__init__.py
```diff
@@ -8,8 +8,11 @@
 def render_template(template, destination, **kwargs):
     template = os.path.join(HERE, template)
 
+
```
Natim: We should remove this line.

Natim:
Shweta can you fix the PEP8 errors showed here: https://travis-ci.org/Kinto/kinto/jobs/97646030#L134

Issue #309, opened 2015-12-01
Trigger Pyramid event on flush


leplatrem:
Some plugins, like kinto-attachment, would require to some cleaning where the server is flushed. Currently there is no way for the plugin to be notified.

leplatrem:
Done in #354

PR #354, opened 2015-12-28
Trigger event on server flush.

johngian:


johngian:
Fixes #309

almet:
Thanks, that's useful. It also showcases that we're missing documentations for events, since no docs are to be updated yet :(.

kinto/tests/test_views_flush.py
```diff
@@ -57,3 +83,9 @@ def test_removes_every_records_of_everykind(self):
         self.app.get(self.collection_url,
                      headers=self.alice_headers,
                      status=403)
+
+    def test_event_triggered_post(self):
+        self.app.get(self.collection_url, headers=self.headers)
+        self.app.get(self.collection_url, headers=self.alice_headers)
+        self.app.post('/__flush__', headers=self.headers, status=202)
+        self.assertEqual(len(self.events), 1)
```
almet: We probably should check here that the event is of the right type.

kinto/views/flush.py
```diff
@@ -8,9 +8,16 @@
                 path='/__flush__')
 
 
+class ServerFlushed(object):
```
almet: This should go in a `kinto.events` module (to be created)

johngian: That was my initial take (following ``cliquet.events`` paradigm) but I wasn't sure if we needed to create that for just one event. I will move the changes to the new module.

almet:
Thanks!

leplatrem:
Excellent, I need this for my kinto-attachments plugin !

PR #362, opened 2016-01-13
Add a deploy button.

Natim:

### DIFF
--- a/docs/get-started.rst
+++ b/docs/get-started.rst
@@ -17,8 +17,16 @@ Mozilla demo server
 
 A Kinto instance is running at https://kinto.dev.mozaws.net/v1/
 
-It should be enough to get started, but the records are flushed every day
-at 7:00 AM UTC.
+It should be enough to get started, but the records are flushed every
+day at 7:00 AM UTC.
+
+
+Deploy an instance on Heroku
+============================
+
+You want to get started with a working online Kinto server right now?
+
+`Click here to deploy your instance <https://dashboard.heroku.com/new?button-url=https%3A%2F%2Fgithub.com%2FKinto%2Fkinto-heroku&template=https%3A%2F%2Fgithub.com%2FKinto%2Fkinto-heroku>`_ for free on Heroku (Free plan for less than 10000 rows).
 
 
 .. _run-kinto-docker:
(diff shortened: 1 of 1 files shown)


almet:
Hi! As I was saying this morning during the standup, I'm not really in favor of having a link to heroku on our landing page.

However, having a "deploy" button that links to a section of the documentation where different ways to deploy are listed seems a more acceptable idea to me.

I can do these changes if you don't want to make them.

Natim:
@almet is it closer to what you had in mind?

Natim:
deploy

deploy

almet:
We need to chose between "get started" and "deploy now", otherwise the text is too long.

Natim:
Let's do some A/B testing then :+1:

docs/tutorials/run-kinto.rst
```diff
@@ -54,7 +66,7 @@ with a *PostgreSQL* container.
 
 ::
 
-    wget https://raw.githubusercontent.com/Kinto/kinto/master/docker-compose.yml
+    wget http://kinto-storage.org/docker-compose.yml
```
leplatrem: How is this file published ? Updated ?

docs/tutorials/run-kinto.rst
```diff
@@ -95,6 +107,7 @@ Then install the package using the default configuration:
 
     pip install kinto
     kinto init
+    kinto migrate
```
leplatrem: Good catch!

leplatrem:
I also vote for two buttons, even if they point to the same page :) but my opinion may not represent most of our users

docs/tutorials/run-kinto.rst
leplatrem: From what I understand we can leave the «official» docker compose

leplatrem: (once #376 is merged)

leplatrem:
We could also just leave «Get started», and then the «Deploy on heroku» paragraph becomes very obvious on the first look

docs/tutorials/run-kinto.rst
Natim: The file is just committed in the kinto-website repository and gh-pages branch.

PR #367, opened 2016-01-13
Install postgresql libraries on kinto init (fixes #313)

### LINKED ISSUE #313: kinto init should install postgresql dependencies

leplatrem:
@Natim what do you think? 

It might not be the ideal solution, but it would still help a lot newcomers...

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -2,6 +2,8 @@ from __future__ import print_function
 import argparse
 import os
 import sys
+from subprocess import call
+
 from six.moves import input
 from cliquet.scripts import cliquet
 from pyramid.scripts import pserve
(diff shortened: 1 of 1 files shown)


Natim:
We might install the wrong cliquet version.

leplatrem:
But that command is what we suggest to do in the docs :/

Natim:
Yes let's do that in anycase it will just try that if dependencies are not installed yet so we should be fine.

kinto/__main__.py
```diff
@@ -65,6 +67,16 @@ def main(args=None):
 
         init(config_file, backend)
 
+        # Install postgresql libraries if necessary
+        if backend == "postgresql":
+            try:
+                import psycopg2  # NOQA
+            except ImportError:
+                ret = call('pip install cliquet[postgresql]'.split())
```
Natim: Note that if the venv is not activated it will fail. Cannot we call pip using a python command?

Natim: http://stackoverflow.com/a/15950647/186202

Natim:
Oh yeah.

leplatrem:
\o/

PR #372, opened 2016-01-14
Remove duplicate glossary

leplatrem:
Every term is already present in http://kinto.readthedocs.org/en/latest/cliquet_glossary.html


docs/concepts.rst
```diff
@@ -97,47 +97,6 @@ however, is controlled from the :ref:`server configuration <configuration>`.
     If a parent defines a permission, it is (*currently*) not possible to restrict
     it in its child objects.
 
-
```
Natim: Should we add a link to the glossary there then?

PR #376, opened 2016-01-15
Even lighter docker image.

Natim:
@leplatrem r?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,25 @@
 # Mozilla Kinto server
-FROM python:3.5
+FROM stackbrew/debian:sid
+MAINTAINER Storage Team irc://irc.freenode.net/#kinto
+
 ADD kinto /code/kinto
-ADD README.rst /code/
-ADD CHANGELOG.rst /code/
-ADD CONTRIBUTORS.rst /code/
-ADD setup.py /code/
-ADD MANIFEST.in /code/
-RUN pip install cliquet[postgresql,monitoring]
-RUN pip install -e /code
+ADD CHANGELOG.rst README.rst CONTRIBUTORS.rst setup.py MANIFEST.in /code/
 ENV KINTO_INI /etc/kinto/kinto.ini
-RUN kinto --ini $KINTO_INI --backend=memory init
-CMD kinto --ini $KINTO_INI migrate && kinto --ini $KINTO_INI start
+
+# Install build dependencies, build the virtualenv and remove build
+# dependencies all at once to build a small image.
+RUN \
+    apt-get update; \
+    apt-get install -y python3 python3-pip python3-venv git build-essential make; \
+    apt-get install -y python3-dev libssl-dev libffi-dev libpq5 libpq-dev; \
+    python3 -m venv /home/kinto; \
+    /home/kinto/bin/pip install -U pip; \
+    /home/kinto/bin/pip install cliquet[postgresql,monitoring]; \
+    /home/kinto/bin/pip install -e /code; \
+    /home/kinto/bin/kinto --ini $KINTO_INI --backend=memory init; \
+    apt-get remove -y -qq git build-essential git make python3-pip python3-venv libssl-dev libffi-dev libpq-dev python3-dev; \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,13 +1,25 @@
...
+ADD CHANGELOG.rst README.rst CONTRIBUTORS.rst setup.py MANIFEST.in /code/
 ENV KINTO_INI /etc/kinto/kinto.ini
-RUN kinto --ini $KINTO_INI --backend=memory init
-CMD kinto --ini $KINTO_INI migrate && kinto --ini $KINTO_INI start
+
+# Install build dependencies, build the virtualenv and remove build
+# dependencies all at once to build a small image.
+RUN \
+    apt-get update; \
+    apt-get install -y python3 python3-pip python3-venv git build-essential make; \
```
leplatrem: nit: make is already in build-essential

nit2: why git ?

Dockerfile
```diff
@@ -1,13 +1,25 @@
...
+    apt-get install -y python3-dev libssl-dev libffi-dev libpq5 libpq-dev; \
```
leplatrem: nit: libpq5 is redundant with libpq-dev

leplatrem: nit2: python3 is redundant with python3-dev

Dockerfile
```diff
@@ -1,13 +1,25 @@
...
+    python3 -m venv /home/kinto; \
+    /home/kinto/bin/pip install -U pip; \
+    /home/kinto/bin/pip install cliquet[postgresql,monitoring]; \
+    /home/kinto/bin/pip install -e /code; \
+    /home/kinto/bin/kinto --ini $KINTO_INI --backend=memory init; \
+    apt-get remove -y -qq git build-essential git make python3-pip python3-venv libssl-dev libffi-dev libpq-dev python3-dev; \
```
leplatrem: Add a comment like: `Uninstall from container every lib that are used for installation only`

Dockerfile
Natim: For both ``libpq5`` and ``python3`` it is to mark them as manually installed in order for them not to be uninstalled with removing ``python3-dev`` and ``libpq-dev``

Dockerfile
Natim: git is in case there are dependency link

Dockerfile
```diff
@@ -1,13 +1,24 @@
...
+ADD CHANGELOG.rst README.rst CONTRIBUTORS.rst setup.py MANIFEST.in /code/
 ENV KINTO_INI /etc/kinto/kinto.ini
-RUN kinto --ini $KINTO_INI --backend=memory init
+
+# Install build dependencies, build the virtualenv and remove build
+# dependencies all at once to build a small image.
+RUN \
+    apt-get update; \
+    apt-get install -y python3 libpq5; \
+    apt-get install -y build-essential git python3-dev python3-setuptools libssl-dev libffi-dev libpq-dev; \
```
leplatrem: micronit: you don't uninstall setuptools :)

leplatrem:
r+!

leplatrem:
from what I understand we can delete the kinto-docker repo after that

Dockerfile
Natim: And I don't uninstall pip, should I?

Natim: That's about it:

```
  File "/usr/local/bin/kinto", line 5, in <module>
    from pkg_resources import load_entry_point
ImportError: No module named 'pkg_resources'
```

leplatrem: ah oh ok :)

nanonit: you can put it on the same line as python3 and libpq5 then :)

Dockerfile
```diff
@@ -1,13 +1,24 @@
 # Mozilla Kinto server
-FROM python:3.5
+FROM debian:sid
```
Natim: stackbrew/debian    sid                 e560b1d2575a        8 days ago          117.2 MB
debian              sid                 e560b1d2575a        8 days ago          117.2 MB

Look at the IMAGE ID, they are the exact same images :)

Issue #277, opened 2015-11-17
Plugins views hide default bucket


leplatrem:
Currently, the default bucket views are registered after `cliquet.initialize()`.

The plugin views (`kinto.includes`) are registered during `cliquet.initialize()`.

The plugin views are hence registered before `/buckets/default/*`, and thus hide it.

This is a limitation for *kinto-attachment* which registers `/buckets/{bucket_id}/collections/{collection_id}/records/{id}/attachment`. 

Possible solutions:
* register default bucket views before `cliquet.initialize()`. *Note: version prefix will thus have to be managed explicitly*
* add an *kinto ready* event on which plugins can hook to do last minute config modifications

Thoughts?

Natim:
I am wondering if we shouldn't put the default_bucket feature as a plugin and let the plugin order handle that current limitation.

PR #380, opened 2016-01-19
Force default bucket plugin only if necessary

leplatrem:
@Natim r?

### DIFF
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -42,13 +42,17 @@ def main(global_config, config=None, **settings):
                        version=__version__,
                        default_settings=DEFAULT_SETTINGS)
 
-    config.include('kinto.plugins.default_bucket')
+    settings = config.get_settings()
+
+    # In Kinto API 1.x, a default bucket is available.
+    # Force its inclusion if not specified in settings.
+    if 'kinto.plugins.default_bucket' not in settings['includes']:
+        config.include('kinto.plugins.default_bucket')
 
     # Retro-compatibility with first Kinto clients.
     config.registry.public_settings.add('cliquet.batch_max_requests')
 
     # Scan Kinto views.
-    settings = config.get_settings()
     kwargs = {}
     flush_enabled = asbool(settings.get('flush_endpoint_enabled'))
     if not flush_enabled:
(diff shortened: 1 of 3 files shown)


kinto/__init__.py
```diff
@@ -42,13 +42,17 @@ def main(global_config, config=None, **settings):
                        version=__version__,
                        default_settings=DEFAULT_SETTINGS)
 
-    config.include('kinto.plugins.default_bucket')
+    settings = config.get_settings()
+
+    # In Kinto API 1.x, a default bucket is available.
+    # Force its inclusion if not specified in settings.
+    if 'kinto.plugins.default_bucket' not in settings['includes']:
+        config.include('kinto.plugins.default_bucket')
```
almet: Should we issue a warning and ask for it to be included in the configuration explicitely now?

Natim:
I don't think people should include it in the config so IMHO we can let it as it is.

leplatrem:
> I don't think people should include it in the config so IMHO we can let it as it is.

But that's the only way to have default bucket work with kinto-attachment ! And that's the main reason  we put the default bucket as a plugin :) See #277 

> Should we issue a warning and ask for it to be included in the configuration explicitely now?

yes, I think so

almet:
Then, how do you remove the feature?

Natim:
As far as I know we don't aim to make it an optional feature do we?

leplatrem:
> Then, how do you remove the feature?

We cannot at this stage.

Currently the client of v1 expects the default bucket to create collections implicitly.

Being able to remove the default bucket could be part of the API v2.

leplatrem:
Ready to review/merge!

kinto/config/kinto.tpl
```diff
@@ -36,6 +36,11 @@ multiauth.policies = basicauth
 # kinto.experimental_collection_schema_validation = true
 
 #
+# Plugins
+#
+kinto.includes = kinto.plugins.default_bucket
```
Natim: It seems strange to me to add it there if it is already added automatically.

leplatrem: So that adding it automatically could be removed in 2.X, whereas config files can stay

Natim:
r+

PR #386, opened 2016-01-21
Activate POST on collections (fixes #237)

### LINKED ISSUE #237: Activate POST on collections
$ curl -H "Content-Type: application/json" \
              -X POST -d '{"data": {"test": "some_data"}}' --user testuser:abc123 \
              https://kinto.dev.mozaws.net/v1/buckets/test_bucket/collections

    {"errno":115,"message":"Method not allowed on this endpoint.","code":405,"error":"Method Not Allowed"}

johngian:


leplatrem:
Excellent ! Thanks for your contribution !

Before we merge, could you please : 

* Add yourself to the contributors files
* Bump the ``HTTP_API_VERSION`` to 1.2 (new feature)
* Add an entry in API changelog `docs/api/index.rst`

kinto/tests/test_views_collections.py
```diff
@@ -107,3 +103,26 @@ def test_records_of_collection_are_deleted_too(self):
...
+    def setUp(self):
+        super(CollectionCreationTest, self).setUp()
+        self.app.put_json('/buckets/beers', MINIMALIST_BUCKET,
+                          headers=self.headers)
+
+    def test_collection_can_be_created_with_post(self):
+        r = self.app.post_json(self.collections_url,
+                               MINIMALIST_COLLECTION,
+                               headers=self.headers)
+        self.assertEqual(r.status_code, 201)
```
leplatrem: Could we write a test here to assert the value of id ?

kinto/tests/test_views_collections.py
```diff
@@ -107,3 +103,26 @@ def test_records_of_collection_are_deleted_too(self):
...
+
+    def test_collection_can_be_specified_in_post(self):
+        collection = 'barley'
+        r = self.app.post_json(self.collections_url,
+                               {'data': {'id': collection}},
+                               headers=self.headers)
```
leplatrem: We could write a test to specify the behaviour when a collection with this id already exists (returns 200 instead of 201 with existing object)

docs/api/1.x/index.rst
```diff
@@ -57,7 +59,7 @@ Cheatsheet
 +----------+----------------------------------------------------------------------------------------------+---------------------------------------------------------+
 | `GET`    | :ref:`/buckets/(bucket_id)/collections/(collection_id)/records <records-get>`                | :ref:`Retrieve stored records <records-get>`            |
 +----------+----------------------------------------------------------------------------------------------+---------------------------------------------------------+
-| `DELETE  | :ref:`/buckets/(bucket_id)/collections/(collection_id)/records <records-delete>`             | :ref:`Delete stored records <records-delete>            |
+| `DELETE` | :ref:`/buckets/(bucket_id)/collections/(collection_id)/records <records-delete>`             | :ref:`Delete stored records <records-delete>`           |
```
almet: Thanks for the fix :)

almet:
Out of curiosity: what's your use case for this? :-)

kinto/tests/test_views_collections.py
johngian: You mean here to check that ``id`` returned is not empty? Next tests checks explicitly the id.

leplatrem: Yes, check that is contains something like uuid (does it? if it does just check the length for ex.)

johngian:
@leplatrem ``HTTP_API_VERSION`` is already ``1.2``. Should I just add the changes to the changelog under ``v1.2``?

johngian:
@almet No use case for now :) I just like the idea.
I have tried some other alternatives but they felt way more complicated for my needs. Kinto has a much more lightweight approach. Plus, I like the stack.

leplatrem:
> HTTP_API_VERSION is already 1.2. Should I just add the changes to the changelog under v1.2?

Good point! Yes I guess since 1.2 was added in the current master and not released yet. So forget what I said about bumping :)

Thanks !

johngian:
@leplatrem PR updated

leplatrem:
Excellent! Thank you again!

PR #387, opened 2016-01-25
Group permissions are not working

aegaas:
Team,

I'm fairly new to python development, but I wanted to build a project on kinto that uses group permissions. Here you see a test that expresses what I'm seeing in the real world, that when I setup a group write permissions on a record, the user that should be able to read and write to it cannot and he gets a 403 instead.

I understand that this functionality might depend on cliquet in order to get fixed, but I wanted to surface this to the team to see if its a known issue or maybe I was doing something wrong.

PR #388, opened 2016-01-26
Record accessible by group members

Natim:
Superseeded #387


leplatrem:
r+, Aaron should be added to contribs :)

PR #391, opened 2016-01-26
Improve formatting of code block in tutorials

leplatrem:
Before:

capture d ecran de 2016-01-26 18-30-28

After

capture d ecran de 2016-01-26 18-30-55


Natim:
r+

PR #404, opened 2016-01-29
Expose the ETag header in 304 responses.

Natim:
Add a test to show the behavior in https://github.com/mozilla-services/cliquet/issues/631

### DIFF
--- a/kinto/plugins/default_bucket/__init__.py
+++ b/kinto/plugins/default_bucket/__init__.py
@@ -120,7 +120,8 @@ def default_bucket(request):
     try:
         response = request.invoke_subrequest(subrequest)
     except httpexceptions.HTTPException as error:
-        if error.content_type == 'application/json':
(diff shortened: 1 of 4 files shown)


leplatrem:
See https://github.com/mozilla-services/cliquet/issues/631#issuecomment-176761484

leplatrem:
@almet r?

almet:
That looks good to me!

almet:
r+

kinto/plugins/default_bucket/__init__.py
```diff
@@ -120,7 +120,8 @@ def default_bucket(request):
     try:
         response = request.invoke_subrequest(subrequest)
     except httpexceptions.HTTPException as error:
-        if error.content_type == 'application/json':
+        is_redirect = error.status_code < 400
```
Natim: I would have tested 304 only there.

PR #406, opened 2016-01-29
Fixed broken hyperlink.

taylorzane:
The link to python.org was referencing a local route instead of an absolute route.


almet:
Thanks!

PR #413, opened 2016-01-31
Fix link to Heroku deploy (fixes #412)

### LINKED ISSUE #412: broken link in docs about single-click deployment on Heroku
There is a broken sphinx reference link on http://kinto.readthedocs.org/en/latest/overview.html It looks like: 

screenshot 2016-01-31 01 12 20

michielbdejong:
In terms of rst format, I left it as a normal relative hyperlink here, not sure if it's better to use `:ref:` like in the link a few lines up, and what the target should be then?

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -247,7 +247,7 @@ and we're a small team.
 
 Currently we make sure it's :ref:`easy to run with Docker or Python pip <get-started>`.
 
-We also have a `single-click deployment on Heroku <get-started>`.
+We also have a `single-click deployment on Heroku <get-started.html#deploy-an-instance-on-heroku>`.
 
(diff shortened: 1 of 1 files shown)


michielbdejong:
Fixes #412.

kumar303:
Yeah, it's probably better as `:ref:`. You could do it in two steps. Add this to the target section in `get-started.rst` 

````
.. _deploy-an-instance-on-heroku:

Deploy an instance on Heroku
============================
...
````
Then you can reference it from the other page like:
````
:ref:`single-click deployment on Heroku <deploy-an-instance-on-heroku>`
````

michielbdejong:
Thanks! Continued in #416.

PR #414, opened 2016-01-31
Fix Travis CI build for Python 3.5

### LINKED ISSUE #411: Interpreter not found for Python 3.5 on TravisCI
capture d ecran de 2016-01-29 21-48-19

cbjadwani:
Fixes #411

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,18 @@ addons:
 env:
     - TOX_ENV=py27
     - TOX_ENV=py34
-    - TOX_ENV=py35
+    #- TOX_ENV=py35
     - TOX_ENV=pypy
     - TOX_ENV=cliquet-master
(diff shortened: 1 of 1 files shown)


leplatrem:
Thanks for your contribution!

I checked Travis and it worked well.

I just wonder if we should do the same for py27, py34 and pypy now :) What do you think ?

almet:
Yes, it seems that matrix is the way to go for a better travis integration.

cbjadwani:
Matrix is not required for other python versions.  This fix was needed for 3.5 to workaround [travis-ci/travis-build#4794](https://github.com/travis-ci/travis-ci/issues/4794)

leplatrem:
I agree to merge it like this. Let's iterate if we can do better!

Thanks again!

PR #426, opened 2016-02-05
Reduce retry-after default duration

leplatrem:
@almet r?


almet:
Great! Should we change it [in cliquet](https://github.com/mozilla-services/cliquet/blob/35b37c9e14c8ac21c15c189918fa34e6a1df91c4/cliquet/__init__.py#L73) as well?

Natim:
r+

leplatrem:
> Great! Should we change it in cliquet as well?

Why not

PR #428, opened 2016-02-08
Delete buckets and collections

leplatrem:
While I was add consistency in kinto.py, I realized that we could had `DELETE` endpoints to buckets and collections.

Of course, it deletes only the buckets/collections where the user has write permissions.

* [x] Implement missing methods
* [x] Rely on events instead of subclassing resource
* [x] Wait for https://github.com/mozilla-services/cliquet/pull/645 to be merged
* [x] Add tests for recursivity (should wipe everything under)
* [x] Update docs


leplatrem:
This is now open for review/merge! 

@magopian, FYI this PR switches from inherited method in the resource class to event subscribers. 
@Natim r?

CHANGELOG.rst
```diff
@@ -6,6 +6,14 @@ This document describes changes between each past release.
 1.12.0 (unreleased)
 ===================
 
+**Protocol**
+
+- Delete every (writable) buckets using ``DELETE /v1/buckets``
+- Delete every (writable) collections using ``DELETE /v1/buckets/<>/collections``
```
magopian: Did you want a `<bucket id>` here instead of `<>`?

docs/api/1.x/index.rst
```diff
@@ -43,7 +45,11 @@ Cheatsheet
 +----------+----------------------------------------------------------------------------------------------+---------------------------------------------------------+
 | **Collections**                                                                                                                                                   |
 +----------+----------------------------------------------------------------------------------------------+---------------------------------------------------------+
-| `POST`   | :ref:`/buckets/(bucket_id)/collections <collection-post>`                                    | :ref:`Create a collection <collection-put>`             |
```
magopian: good catch for the `collection-post` instead of `collection-put` here ;)

docs/api/index.rst
```diff
@@ -15,7 +15,8 @@ Changelog
 ''''''''''''''''
 
 - Allow bucket to get arbitrary attributes.
-
+- Delete every (writable) buckets using ``DELETE /v1/buckets``
+- Delete every (writable) collections using ``DELETE /v1/buckets/<>/collections``
```
magopian: Same here, did you want to have `<bucket id>` instead of `<>`?

kinto/views/buckets.py
```diff
@@ -23,12 +26,20 @@ def get_parent_id(self, request):
...
+            for_resources=('bucket',),
+            for_actions=(ACTIONS.DELETE,))
+def on_buckets_deleted(event):
+    """Some buckets were deleted, delete sub-resources.
+    """
+    storage = event.request.registry.storage
+
+    for change in event.impacted_records:
+        bucket = change['old']
+        parent_id = '/buckets/%s' % bucket['id']
```
magopian: nit: don't you use the better/faster/stronger `.format` for string interpolation?

kinto/views/groups.py
```diff
@@ -1,6 +1,8 @@
 import colander
```
magopian: Changes in this file aren't related to the PR, are they? You took the opportunity to change the group deletion to using the subscriber too?

magopian:
From what I understand of the code I read, this looks good to me. If someone else with better knowledge of the codebase could double-check, that would be nice ;)

There's a test failure, from the load tests, is it related to what you were discussing with @Natim on irc about cliquet having timestamp unicity issues?

kinto/views/groups.py
leplatrem: Yes, you're right, I move everything from inheritance to events

kinto/views/buckets.py
leplatrem: erf we should, but not only here...

http://makina-corpus.com/blog/metier/the-worlds-simplest-python-template-engine :)

PR #429, opened 2016-02-09
Add hurl.it links to the tutorials to follow them from your browser

magopian:
This is a test to open up the discussion: do we want links to [hurl.it](https://www.hurl.it) ([source](https://github.com/defunkt/hurl)).

Here's an example of url that pre-populates everything (but the captcha ;) for the very first example in the `first-steps` tutorial: 
[post a sample record](https://www.hurl.it/?method=POST&url=https://kinto.dev.mozaws.net/v1/buckets/default/collections/tasks/records&username=token&password=my-secret&headers={%22Content-Type%22:[%22application/json%22]}&body={%22data%22:%20{%22description%22:%20%22Write%20a%20tutorial%20explaining%20Kinto%22,%20%22status%22:%20%22todo%22}})

### DIFF
--- a/docs/tutorials/first-steps.rst
+++ b/docs/tutorials/first-steps.rst
@@ -56,6 +56,10 @@ We use the Mozilla demo server:
         http POST https://kinto.dev.mozaws.net/v1/buckets/default/collections/tasks/records \
              -v --auth 'token:my-secret'
 
+Here's a quick link to test the same request from your browser using the
+`hurl.it <https://www.hurl.it>`_ web service:
+`post a sample record <https://www.hurl.it/?method=POST&url=https://kinto.dev.mozaws.net/v1/buckets/default/collections/tasks/records&username=token&password=my-secret&headers={%22Content-Type%22:[%22application/json%22]}&body={%22data%22:%20{%22description%22:%20%22Write%20a%20tutorial%20explaining%20Kinto%22,%20%22status%22:%20%22todo%22}}>`_.
+
 .. code-block:: http
 
     HTTP/1.1 201 Created
@@ -95,6 +99,8 @@ Let us fetch our new collection of tasks:
     $ http GET https://kinto.dev.mozaws.net/v1/buckets/default/collections/tasks/records \
            -v --auth 'token:my-secret'
 
+`Get the collection of tasks <https://www.hurl.it/?method=GET&url=https://kinto.dev.mozaws.net/v1/buckets/default/collections/tasks/records&username=token&password=my-secret>`_.
+
 .. code-block:: http
 
     HTTP/1.1 200 OK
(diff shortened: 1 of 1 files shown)


almet:
Thanks for taking the time to write this.

Here is a capture of what I see when clicking on the link and then posting the data to the kinto server. The commercial service hurl.it displays ads.

hurl.

That would probably be beneficial to have our own instance of this service rather than relying on a third party: that would let us remove the ads we don't want.

Also, maybe there could be a way to not get the user out of the context of the tutorial each time he/she clicks on a link: that would be great to have the same experience but staying inside the documentation.

We were discussing this with @n1k0 last time, and I wonder if users will actually use these features? (I for once would copy the curl/httpie links in my terminal). @magopian for instance, would you use such links if they were present?

leplatrem:
The formatted response is not showing well, any idea?

capture d ecran de 2016-02-09 16-04-55

magopian:
@almet I would personally use it, yes, I find it far more convenient and ergonomic to edit params in a web form, than on the command line (think about very long command lines, wrapping over several lines, and you can't position the cursor using your mouse).

It would indeed be much better if we had our own instance of hurl, and even better if we could have all that in the context of the page itself (but then, switching to a terminal is also a context switch, maybe even more than switching tabs).

So, maybe, just maybe, we could take this in steps?
1/ add the urls to hurl
2/ install / use our own instance (and modify the links to point to it)
3/ study the feasibility of having the forms inline with the docs (maybe use something like swagger? or... the kinto-js lib ;)

aurelg:
@almet When browsing the documentation without (the possibility to have) a terminal, such links would probably be very useful. Think mobile devices for instance.

almet:
Thanks for your feedback. That sounds like a good course of actions. I just don't want to point our users to a commercial website if we can avoid it :)

almet:
Here it is! http://hurl.kinto-storage.org

magopian:
that was blazingly fast

almet:
After some time looking around, it seems that the free version of hurl doesn't contain the same features that the closed version has. I wonder if we should look at other potential alternatives.

almet:
Still the same company, but I like how they provide the command line and next to it a "send to hurl" link: https://www.embedcurl.com/

almet:
After looking a bit more, it seems that this free version of hurl actually proposes to have a bookmark to the requests/responses, but you need to click the "permalink" link at the bottom of the page.

It seems that this permalink is actually [a hash of the parameters](https://github.com/defunkt/hurl/blob/master/app/app.rb#L249-L250).

almet:
Relevant — https://twitter.com/ametaireau/status/697129422172897281

magopian:
After thinking some more about it, I agree that it doesn't make sense to use a closed service in our official docs.

This means there's the following solutions:
1/ modify the current open source solution to support parsing the url fragment into form inputs, the way it's done on hurl.it
2/ implement another solution, that could be embedded straight into the documentation: this would be even better and more ergonomic/useful than having to open a new tab for each API call, and could also allow us to "pipe" the output of one API call into the form for the next API call (making it dynamic, being able to re-use the IDs, ETags and others). However, this would mean quite a lot of work
3/ leave the documentation as it is, and maybe add a note about the existence of web services that can be used in place of the command line

leplatrem:
Are you still willing to merge this?

almet:
I think we can close this pull request for now. I created #502 for the idea of having executable examples (that can run from the documentation directly).

Also, as per a discussion this morning, it seems that for the documentation, it might make sense to use [the docbox project](https://github.com/mapbox/docbox).

PR #448, opened 2016-02-18
python as a plugin required to run kinto in config/kinto.ini file

ayusharma:
Python as a plugin required to solve the following warning and issue:
```
*** Operational MODE: preforking ***
!!!!!!!!!!!!!! WARNING !!!!!!!!!!!!!!
no request plugin is loaded, you will not be able to manage requests.
you may need to install the package for your language of choice, or simply load it with --plugin.
!!!!!!!!!!! END OF WARNING !!!!!!!!!!
spawned uWSGI master process (pid: 29841)
spawned uWSGI worker 1 (pid: 29842, cores: 1)
spawned uWSGI worker 2 (pid: 29843, cores: 1)
spawned uWSGI worker 3 (pid: 29844, cores: 1)
*** no app loaded. going in full dynamic mode ***
*** no app loaded. going in full dynamic mode ***
*** no app loaded. going in full dynamic mode ***
```

### DIFF
--- a/docs/configuration/production.rst
+++ b/docs/configuration/production.rst
@@ -333,6 +333,7 @@ Here's an example:
     single-interpreter = true
     buffer-size = 65535
     post-buffering = 65535
+    plugins = python
 
 To use a different ini file, the ``KINTO_INI`` environment variable
 should be present with a path to it.


Natim:
Is it when you want to handle multiple python versions?

I cannot find something else than this in the UWSGI documentation: http://uwsgi-docs.readthedocs.org/en/latest/WSGIquickstart.html#bonus-multiple-python-versions-for-the-same-uwsgi-binary

ayusharma:
Yes. You are right. But there is no plugin attribute is mentioned in documentation. It is necessary to mention `plugins = python` in `config/kinto.ini` file to keep kinto working. So I added it to doc.

leplatrem:
Thanks !

PR #451, opened 2016-02-22
Improve synchronisation docs

leplatrem:
@Natim @michielbdejong @n1k0 feeback ?

capture d ecran de 2016-02-26 17-25-39

capture d ecran de 2016-02-22 16-34-37

capture d ecran de 2016-02-22 16-34-44
capture d ecran de 2016-02-26 17-25-09
#

### DIFF
--- a/docs/api/1.x/synchronisation.rst
+++ b/docs/api/1.x/synchronisation.rst
@@ -132,8 +132,8 @@ For example, start with some recent records in order to populate a UI,
 and then fetch older records in background.
 
 #. Obtain a few pages of recent records using the *newest first* strategy from above
-#. In background, fetch old records using ``_sort=-lastmodified`` and ``_before=MIN(local_records[last_modified])``
-#. Recent changes can be obtained using ``_sort=-lastmodified`` and ``_since=MAX(local_records[last_modified])``
+#. In background, fetch old records using ``_sort=-last_modified`` and ``_before=MIN(local_records[last_modified])``
+#. Recent changes can be obtained using ``_sort=-last_modified`` and ``_since=MAX(local_records[last_modified])``
 
 .. image:: ../../images/sync-both.svg
 
--- a/docs/api/1.x/synchronisation.rst
+++ b/docs/api/1.x/synchronisation.rst
@@ -38,6 +38,7 @@ strategies to poll the server for changes.
 
 .. important::
 
+    * Timestamps are unique.
     * Deleted records have an attribute ``delete: true``.
     * Created/updated records are both returned in their new version.
     * Since *Kinto* does not keep any history, there is no *diff* for updates.
--- a/docs/api/1.x/synchronisation.rst
+++ b/docs/api/1.x/synchronisation.rst
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
 
 
 
-The basic idea is to keep a local database up to date with the Kinto server.
-Remote changes are downloaded and applied
-on the local data. Local changes are uploaded using HTTP headers to control
-concurrency and overwrite.
+The basic idea is to keep a local database up to date with the Kinto server:
+
+* Remote changes are downloaded and applied on the local data.
+* Local changes are uploaded using HTTP headers to control concurrency and overwrites.
+
 
 In short:
 
 #. Poll for remote changes using ``?_since=<timestamp>``
 #. Apply changes locally
-#. Apply creations remotely
-#. Use concurrency control to send updates and deletes
+#. Send local creations
+#. Use concurrency control to send local updates and deletes
 
 
-Polling for changes
-===================
+Polling for remote changes
+==========================
 
-When fetching the collection records with ``?_since=<timestamp>``, *Kinto* returns
-every records that was created/updated/deleted **after** this timestamp.
+*Kinto* supports range queries for timestamps. Combining them with the sort parameter
+allows to fetch changes in a particular order.
 
-#. Cold sync → no ``?_since``.
-#. ``GET /buckets/default/collections/articles?_since=<timestamp>`` (with header ``If-Match: <timestamp>``)
-#. If response is ``304 Not modified``, done → up to date, nothing to do.
-#. If response is ``200 OK`` store ``ETag`` response header value for next
-   synchronisation and handle the obtained list of changes.
+Depending on the context (latest first, readonly, etc.), there are several
+strategies to poll the server for changes.
 
-.. note::
+.. important::
 
     * Deleted records have an attribute ``delete: true``.
     * Created/updated records are both returned in their new version.
     * Since *Kinto* does not keep any history, there is no *diff* for updates.
 
 
-Paginated changes
------------------
+Pagination
+----------
 
-By default, *Kinto* does not paginate the records list. If an explicit limit is
-set in the server settings or using the ``?_limit=<nb>`` parameter, then polling for
-changes will be paginated.
+By default, *Kinto* does not paginate the records list. Since an explicit limit can
+be set in the server settings, clients must handle pagination when polling for
+changes.
 
-It basically consists in fetching the list until the ``Next-Page`` is present.
+In order to reduce the size of response payloads, the client can also force the
+pagination using the ``?_limit=<nb>`` querystring parameter.
 
-Since changes can occur between the first and the last page, the synchronisation
-process is a bit more complex.
+Pagination basically consists in fetching the list until the ``Next-Page`` response header
+is present. The ``Next-Page`` header is the **full** URL of the next page.
 
-#. Cold sync → no ``?_since``.
-#. ``GET /buckets/default/collections/articles?_since=<timestamp>`` (with header ``If-Match: <timestamp>``)
-#. If response is ``304 Not modified``, done → up to date, nothing to do.
-#. If no ``Next-Page`` response header, done → store ``ETag`` response header value for next
-   synchronisation.
-#. If ``Next-Page`` response header is present → store the ``ETag`` response header value into a variable ``timestamp``
-   and go on to it (it's an url) using a ``If-Match: <timestamp>`` request header.
-#. If response is ``200 OK`` → repeat previous step.
-#. If response is ``412 Precondition Failed``, some changes occured since the last
-   page → Store the ``ETag``response header into ``before``
-#. Fetch and handle the changes using ``GET collections?_since=<timestamp>&_before=<before>``,
-   and store the ``ETag``response header in to ``timestamp``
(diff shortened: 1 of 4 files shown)


docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+Strategy #1 — Oldest first
+--------------------------
+
+The simplest way to obtain the changes is to sort the records by timestamp
+ascending.
+
+We will use ``sort=last_modified`` and ``_since=<timestamp>``:
+
+#. First sync: ``timestamp := 0``
+#. Next sync: ``timestamp := MAX(local_records['last_modified'])``
```
michielbdejong: `timestamp := MAX(local_records['last_modified'])-1` because it may be that two or more changes happened during the same timestamp, and you did not get all of them during the first sync.

This is especially relevant if you combine this with `_limit`. For instance, in FxSync you sometimes see 1000 records with the exact same timestamp, so you shouldn't discard 950 of them after fetching only the first 50.

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+#. Fetch ``GET /buckets/<bid>/collections/<cid>/records?_sort=last_modified&_since=<timestamp>``
+#. If response is ``200 OK``, handle the list of remote changes.
+#. If response has ``Next-Page`` header, follow full URL in header.
```
michielbdejong: `Next-Page` response header will only occur if a limit was set

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+#. If list of changes is empty, **done** → up-to-date.
+
+.. image:: ../../images/sync-oldest.svg
+
+If an error occurs during the obtention of pages,
+the synchronisation can be resumed transparently, since it relies on the highest
+timestamp successfully stored locally.
```
michielbdejong: This is not the only reason; it's also because it keeps track of the id of the first record of the next page to fetch, not its index in the list. Suppose you had retrieved records `a,b,c,d,e` in the first batch, and then record `b` is changed. If you would then do `offset=5`, you would skip the first 5 records, including record `f` (because the list now is `a,c,d,e,f`). The trick is to not do `offset=5` but `start_at=f`.

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+
+
+Strategy #2 — Newest first
+--------------------------
+
+In order to populate a UI, it might be relevant to obtain the latest changes first.
+
+Syncing newest records first is a bit more complex since changes can occur between
+the obtention of the first and the last pages.
```
michielbdejong: I think the `Next-Page` system will take care of that; suppose you retrieved `z,y,x,w,v`, record `b` changes, and you continue retrieving records `u...,d,c,a`. On the next sync run, you'll get only the `b` record that was updated since you started your initial sync run

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+
+#. First sync: ``timestamp := 0``
+#. Next sync: use ``timestamp`` stored in last successful sync.
+#. Fetch current collection timestamp ``HEAD /buckets/<bid>/collections/<cid>/records``
+   in ``ETag`` response header and store its value in ``start``.
+#. Fetch ``GET /buckets/<bid>/collections/<cid>/records?_sort=-last_modified&_before=<start>&_since=<timestamp>``
+#. If response is ``200 OK``, stack the obtained list of remote changes.
+#. If response has ``Next-Page`` header, follow full URL in header.
+#. If list of changes is empty, **done** → handle the stack of remote changes
+   and update the timestamp: ``timestamp := MAX(local_records['last_modified'])``
```
michielbdejong: ``timestamp := MAX(local_records['last_modified']) - 1`` (see above)

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+
+.. image:: ../../images/sync-newest.svg
+
+With this approach, the main algorithm is rather simple but if an error occurs
+between the first and the last step, the client must redownload every page obtained
+from *step 1* until it succeeds to fetch every page of the sync.
```
michielbdejong: I don't think this is true?

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+
+Strategy #3 — Newest then oldest
+--------------------------------
+
+For very large collections, it could be interesting to perform a first partial
+synchronisation, and then fetch old records in background.
+
+When a new client wants to sync, it gets some recent records using the newest
+strategy, and then fetches the old records in a second phase, in background
+for example.
```
michielbdejong: What's the advantage of this?

docs/api/1.x/synchronisation.rst
Natim: > because it may be that two or more changes happened during the same timestamp

This is not possible on server side. Even if two records happens at the same timestamp, they will have different timestamps.

See http://kinto.readthedocs.org/en/latest/api/1.x/cliquet/timestamps.html?highlight=timestamp

Natim: >  in FxSync you sometimes see 1000 records with the exact same timestamp, so you shouldn't discard 950 of them after fetching only the first 50.

This is a real problem for syncto because Kinto guarantees that it won't happend. If Sync doesn't it means that we have a problem in syncto.

docs/api/1.x/synchronisation.rst
Natim: However a default limit can be set in the configuration.

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,129 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+be set in the server settings, clients must handle pagination when polling for
+changes.
 
-It basically consists in fetching the list until the ``Next-Page`` is present.
+In order to reduce the size of response payloads, the client can also force the
+pagination using the ``?_limit=<nb>`` querystring parameter.
 
-Since changes can occur between the first and the last page, the synchronisation
-process is a bit more complex.
+Pagination basically consists in fetching the list until the ``Next-Page`` response header
```
Natim: > the list until the ``Next-Page`` response header is present

the list while the ``Next-Page`` response header is present ?

docs/api/1.x/synchronisation.rst
Natim: We probably need to update the start parameter at some point.

leplatrem: I don't think we need this `-1`, since the `_since` and `_before` are both exclusive

n1k0:
Very clear and insightful, answers a bunch of questions I had; r+.

docs/api/1.x/synchronisation.rst
michielbdejong: But we need inclusive, right? Say the records have been changes at `t=1,2,3` in a sequence `11122 22233` and your fetch limit is 5. Then your first page should be `_since=0` and the second page should be `_since=1` (so that you don't skip the `222` records in the second page).

leplatrem:
> why would you do it this way instead of just keep going in reverse chronological order?

Yeah, I changed it. In the third approach, they both go backwards. One fetches the old records with `_before=MIN(...)` and the other one fetches the new records with `_since=MAX()`

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,133 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+
+When a new client wants to sync, instead of syncing hundreds of pages on the
+first synchronization, two distinct synchronization processes can be combined.
+
+For example, start with some recent records in order to populate a UI,
+and then fetch older records in background.
+
+#. Obtain a few pages of recent records using the *newest first* strategy from above
+#. In background, fetch old records using ``_sort=-lastmodified`` and ``_before=MIN(local_records[last_modified])``
+#. Recent changes can be obtained using ``_sort=-lastmodified`` and ``_since=MAX(local_records[last_modified])``
```
michielbdejong: I still think you'll skip records with this strategy unless you can be sure that each record has a unique `lastmodified` timestamp. As soon as you have 1 local record with a given timestamp, all other remote records with that same timestamp will be ignored, right?

docs/api/1.x/synchronisation.rst
leplatrem: In Kinto the timestamps are guaranteed to be unique at each write operation. So there will never be two records with the same timestamp! (there is a database constraint for that by the way)

docs/api/1.x/synchronisation.rst
```diff
@@ -13,67 +13,133 @@ This section describes the basic aspects of synchronisation using *Kinto*.
...
+synchronisation, and then fetch old records in background.
+
+When a new client wants to sync, instead of syncing hundreds of pages on the
+first synchronization, two distinct synchronization processes can be combined.
+
+For example, start with some recent records in order to populate a UI,
+and then fetch older records in background.
+
+#. Obtain a few pages of recent records using the *newest first* strategy from above
+#. In background, fetch old records using ``_sort=-lastmodified`` and ``_before=MIN(local_records[last_modified])``
```
Natim: ``_sort=-last_modified`` isn't it?

docs/api/1.x/synchronisation.rst
michielbdejong: ok, I remember now, sorry. So then it's safe for Kinto. We could add a note saying that for Syncto, this constraint is not true, and records will actually get skipped when using this approach (in my test account, I had hundreds of records with identical timestamp, because the FxSync sets the first sync time as the timestamp for pre-existing bookmarks).

leplatrem: No problem, thanks for reviewing this thoroughly!

> I had hundreds of records with identical timestamp, because the FxSync sets the first sync time as the timestamp for pre-existing bookmarks).

@Natim we've got a problem then! :) I opened issue https://github.com/mozilla-services/syncto/issues/82

Issue #440, opened 2016-02-12
Postgres Encoding


chrismbeckett:
Trying to do a fresh setup of Kinto using the Python packages with Postgres. I am running into an issue where running ``kinto migrate`` is giving me the error:

```sh
/home/ubuntu/workspace/kinto/local/lib/python2.7/site-packages/cliquet/storage/postgresql/client.py:86: UserWarning: Reuse existing PostgreSQL connection. Parameters permission_* will be ignored.
  warnings.warn(msg)
/home/ubuntu/workspace/kinto/local/lib/python2.7/site-packages/cliquet/storage/postgresql/client.py:86: UserWarning: Reuse existing PostgreSQL connection. Parameters cache_* will be ignored.
  warnings.warn(msg)
Traceback (most recent call last):
  File "/home/ubuntu/workspace/kinto/bin/kinto", line 11, in <module>
    sys.exit(main())
  File "/home/ubuntu/workspace/kinto/local/lib/python2.7/site-packages/kinto/__main__.py", line 79, in main
    cliquet.init_schema(env)
  File "/home/ubuntu/workspace/kinto/local/lib/python2.7/site-packages/cliquet/scripts/cliquet.py", line 31, in init_schema
    getattr(registry, backend).initialize_schema()
  File "/home/ubuntu/workspace/kinto/local/lib/python2.7/site-packages/cliquet/storage/postgresql/__init__.py", line 93, in initialize_schema
    self._check_database_encoding()
  File "/home/ubuntu/workspace/kinto/local/lib/python2.7/site-packages/cliquet/storage/postgresql/__init__.py", line 144, in _check_database_encoding
    raise AssertionError('Unexpected database encoding %s' % encoding)
AssertionError: Unexpected database encoding sql_ascii
```
I have been able to reproduce this multiple times. I am running on Ubuntu Trusty (14.04 LTS) on Cloud9 (c9.io), with Postgres 9.5.1. The default database encoding (as shown below) is SQL_ASCII.

![psql](https://cloud.githubusercontent.com/assets/6036093/13018957/70255512-d181-11e5-934a-7ce2bcf22140.png)

I am not sure why my default encoding is SQL_ASCII instead of UTF-8, but since the Postgres docs say you can set the encoding for a new database when creating it, I don't think that Kinto should fail if the default encoding is not what is expected.

ayusharma:
open `pgsql` 

and change your database encoding:

`
[...]

PR #453, opened 2016-02-23
Add how to troubleshoot psql encoding problems.

almet:
Solution mentioned in #440.

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -71,3 +71,11 @@ Make repository::
 
 Also, make sure the user that runs uwsgi can access /var/run/uwsgi and can
 write in the uwsgi directory.
+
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.rst
```diff
@@ -71,3 +71,11 @@ Make repository::
 
 Also, make sure the user that runs uwsgi can access /var/run/uwsgi and can
 write in the uwsgi directory.
+
+Unexpected database encoding sql_ascii
+======================================
+
```
Natim: If possible I would add the exact message returned by postgresql. So that looking for the solution would come up in the search engine.

Also it is possible other encoding would raise a similar error.

almet: updated.

PR #463, opened 2016-02-25
parser for kinto command

ayusharma:
Kinto --version command

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -8,7 +8,7 @@ from six.moves import input
 from cliquet.scripts import cliquet
 from pyramid.scripts import pserve
 from pyramid.paster import bootstrap
-
+from kinto import __version__
 from kinto.config import init
 
 CONFIG_FILE = 'config/kinto.ini'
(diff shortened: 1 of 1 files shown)


kinto/__main__.py
```diff
@@ -30,6 +30,8 @@ def main(args=None):
                         dest='backend',
                         required=False,
                         default=None)
+    parser.add_argument('--version', action='version', version=__version__,
+                            help='Print the Kinto version and exit.')
```
almet: flake8 reports that ``kinto/__main__.py:34:29: E127 continuation line over-indented for visual indent``

almet:
Thanks, that looks great, you just need to address the little nitpick and we're good to merge!

kinto/__main__.py
```diff
@@ -30,6 +30,8 @@ def main(args=None):
                         dest='backend',
                         required=False,
                         default=None)
+    parser.add_argument('--version', action='version', version=__version__,
```
Natim: Should we also support the sort '-v' option?

    parser.add_argument('-v', '--version', action='version', ...

almet: I believe this will do it by default. In which case it could be good to define it explicitely here, you're right.

Natim:
Fixed and merged in 55de173

Natim:
Thank you for that :+1:

PR #464, opened 2016-02-25
Add a custom id generator tutorial.

almet:
r? @chrismbeckett

### DIFF
--- a/docs/tutorials/custom-id-generator.rst
+++ b/docs/tutorials/custom-id-generator.rst
@@ -34,11 +34,22 @@ Let's define a "name generator", which will output 8 random characters
 Note that you actually need to define both the generator **and** the validator,
 which can be provided as a regexp.
 
+.. important::
+
+    The id generator should be random enough to avoid collisions.
+
+    The random approach here serves the purpose of this tutorial, but for a production
+    ready implementation, have a look at :github:`dylang/shortid` for example.
+
+
+Enable in configuration
+-----------------------
+
 Okay, once we've got this new generator, we can put it in a file named
-``name_generator.py``, and then configure the kinto server to use this
+``name_generator.py``, and then configure the Kinto server to use this
 generator to actually generate the record IDs.
 
-In your `kinto.ini` configuration file, set the ``kinto.id_generator``
+In your :file:`config/kinto.ini` configuration file, set the ``kinto.id_generator``
 setting:
 
 .. code-block:: ini
@@ -49,7 +60,7 @@ And then run your server as usual
 
 .. code-block:: bash
 
-    $ pserve config/kinto.ini
+    $ kinto --ini config/kinto.ini start
 
 Now, if you try to generate a new record, you should have the new ID used,
 well done!
@@ -80,3 +91,16 @@ well done!
             ]
         }
     }
+
+.. note::
+
+    Currently the configuration of the custom generator applies to every resources
+    (buckets, groups, collections, records). This tiny limitation can easily be
+    fixed, don't hesitate to :ref:`get in touch with us <communication_channels>`!
+
+
+Kinto.js client
+---------------
+
+The offline-first client :github:`Kinto/kinto.js` can also be configured to
+generate custom ids. See `its documentation <https://kintojs.readthedocs.org/en/latest/api/#custom-id-generation-using-an-id-schema>`_ for more details.
--- a/docs/tutorials/custom-id-generator.rst
+++ b/docs/tutorials/custom-id-generator.rst
@@ -1,0 +1,82 @@
+.. _tutorial-id-generator:
+
+How to define and use a custom ID generator?
+============================================
+
+By default, Kinto uses UUID4 as ``id`` for its records. In some cases, it might
+not be what you want. But you're not out of luck, there is a way to configure
+Kinto to use a different ``id_generator`` than the default one.
+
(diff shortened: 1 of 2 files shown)


chrismbeckett:
Yeah, that's me :)

------ Original Message ------

docs/tutorials/custom-id-generator.rst
```diff
@@ -0,0 +1,82 @@
...
+            "id": "fm64GtdA",
+            "last_modified": 1456438400719,
+            "yeah": "oh"
+        },
+        "permissions": {
+            "write": [
+                "basicauth:2025f72e6967625e3e878288b55d8946839e51968d11991b8a7dd0f040d4b6f0"
+            ]
+        }
+    }
```
leplatrem: I would add a note here to say that *currently the configuration of the custom generator applies to every resources*

It's a tiny limitation and could easily be fixed

docs/tutorials/custom-id-generator.rst
```diff
@@ -0,0 +1,82 @@
...
+            ascii_letters = ('abcdefghijklmopqrstuvwxyz'
+                             'ABCDEFGHIJKLMOPQRSTUVWXYZ')
+            alphabet = ascii_letters + string.digits + '-_'
+            letters = [random.choice(ascii_letters + string.digits)]
+            letters += [random.choice(alphabet) for x in range(7)]
+            return ''.join(letters)
+
+
+Note that you actually need to define both the generator **and** the validator,
+which can be provided as a regexp.
```
leplatrem: There is something important I guess: the id generator should be *random enough* to avoid *id collisions*

docs/tutorials/custom-id-generator.rst
```diff
@@ -0,0 +1,82 @@
...
+
+    import random
+    import string
+
+    from cliquet.storage import generators
+
+
+    class NameGenerator(generators.Generator):
+
+        regexp = r'^[a-zA-Z0-9-_]{8}$'
```
Natim: I would rather go for a base64 of ``os.urandom(9)`` bytes implementation.

Natim:
```python
>>> import os, base64
>>> base64.b64encode(os.urandom(9))
'TKmj0DI4wcLh'
>>> len(base64.b64encode(os.urandom(9)))
12
```

docs/tutorials/custom-id-generator.rst
Natim: To avoid id collisions within the scope of the bucket collection.

docs/tutorials/custom-id-generator.rst
Natim:
```
>>> len(base64.b64encode(os.urandom(6)))
8
```

Natim: We could even make the number of bytes a settings of the generator.

almet: ``os.urandom`` is probably not a good idea because it's suppose to be random. From the python docs:

>Return a string of n random bytes suitable for cryptographic use. This function returns random bytes from an OS-specific randomness source. The returned data should be unpredictable enough for cryptographic applications.

almet: nvm.

Natim: I did a pull request on https://github.com/corpix/shortid/pull/2 about this.

chrismbeckett:
Moving the discussion between @Natim and I here, instead of having it over in the [Corpix/shortid](https://github.com/corpix/shortid/issues/1#issuecomment-190329759).

I am not really sure that the merits of *ShortID* needed to get pulled into this - the example provided here using urandom() is great for an example. I think people would like to use their own ID generation strategies for all sorts of reasons. The main use case would be they want to use a common ID format and strategy between different systems/services. But, since I was originally asking how to integrate ShortID specifically into Kinto in the original IRC chat, I thought I would elaborate on the requirements I personally use regarding ID generation on my own projects, and why I initiated this request in the first place. Here is what I look for:

1. Polyglot (works in multiple languages, since I may need to use it across different micro-services)
2. Isomorphic (works across client and server so I can generate my record ID's anywhere)
3. Crypto-based strategy that reasonably ensures the required degree of nearly-true randomness
4. URL friendly (works well across HTTP boundaries)
4. Popular/Proven (leveraging the wisdom of crowds/proven in production to ensure it is sound)

The following are a number of strategies that I have found that fit this bill:

1. UUID - globally unique, proven, but large and less user friendly if you don't need *globally* unique

For strategies that don't require global uniqueness, for collection ID's I have used the following that fit the above requirements:

2. ObjectID - strategy taken from MongoDB, and now ported widely. Has the benefit of being time ordered.
3. [FirebaseID]((https://www.firebase.com/blog/2015-02-11-firebase-unique-identifiers.html) - url friendly, scalable, works, and has the benefit of supporting both time and lexigraphic ordered)
4. [ShortID](https://github.com/dylang/shortid) - growing in popularity, proven, being ported.

almet:
I concur with this: the tutorial should not rely on external dependencies if there is no need to. ShortID could be a great addition, but that's a feature, and out of the scope of this tutorial.

chrismbeckett:
Gratz :) Thank you both for helping me with this. I think this example will be a big help to many people.

PR #474, opened 2016-02-29
Fix secret in ini on Python 3 (fixes #341)

leplatrem:
@Natim r?

### DIFF
--- a/kinto/config/__init__.py
+++ b/kinto/config/__init__.py
@@ -1,5 +1,4 @@
 import os
-import binascii
 import codecs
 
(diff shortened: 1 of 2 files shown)


Natim:
r+

Natim:
It is hard to grasp what got fixed without a test :)

leplatrem:
Since no test existed for that part, I felt lazy to start them :]

Natim:
I think if we are fixing a bug there we should add one.

leplatrem:
Will do.

PR #481, opened 2016-03-01
Mention special case of SyncTo for synchronisation

leplatrem:


Natim:
WFM

docs/api/1.x/synchronisation.rst
```diff
@@ -138,6 +138,11 @@ and then fetch older records in background.
 
 .. image:: ../../images/sync-both.svg
 
+.. warning::
```
almet: I'm not sure we should add mozilla's specific doc here (this should probably leave in syncto instead)

leplatrem: Yep, I think you're right :)

almet:
Let's close this then!

Issue #270, opened 2015-11-12
Expose the experimental json-schema feature in the public settings on the hello_page.


Natim:
https://kinto.dev.mozaws.net/v1/

Natim:
I wonder if this list of public exposed settings should not be configurable using the configuration file.

Natim:
Now that we have capabilities it should be exposed as a capability.

Here is an example of how a capability can be exposed: https://github.com/Kinto/kinto-attachment/blob/master/kinto_attachment/__init__.py#L19-L22

Here is the place where we could add the capability: https://github.com/Kinto/kinto/blob/8cc9b87b7776b68da391790f99a9b8b0aa65195f/kinto/__init__.py#L55

Here as some example of the two tests that could be written:  https://github.com/Kinto/kinto-attachment/blob/a3b531662b6468dd2585e92a238e99551439703e/kinto_attachment/tests/test_plugin_setup.py

Natim:
@lavish205 said on IRC he wanted to work on this bug.

Natim:
You need to expose the capability if the settings is set to true:

Here is the way to do it:
```
config.add_api_capability("schema",
                          description="Enforce a schema for collection records.",
                          url="http://kinto.readthedocs.org/en/latest/api/1.x/"
                              "collections.html?highlight=schema#collection-json-schema")
```

lavish205:
@Natim can you assign this bug to me?

PR #486, opened 2016-03-03
Expose the experimental json-schema feature in the public settings on the hello_page

lavish205:
Refs: https://github.com/Kinto/kinto/issues/270

### DIFF
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -54,12 +54,15 @@ def main(global_config, config=None, **settings):
     config.registry.public_settings.add('cliquet.batch_max_requests')
 
     # Expose capability
-    schema_enabled = asbool(settings.get('experimental_collection_schema_validation'))
+    schema_enabled = asbool(
+        settings.get('experimental_collection_schema_validation')
+    )
     if schema_enabled:
-        config.add_api_capability("schema",
-                                  description="Enforce a schema for collection records.",
-                                  url="http://kinto.readthedocs.org/en/latest/api/1.x/"
-                                      "collections.html#collection-json-schema")
+        config.add_api_capability(
+            "schema",
+            description="Enforce a schema for collection records.",
+            url="http://kinto.readthedocs.org/en/latest/api/1.x/"
+            "collections.html#collection-json-schema")
 
     # Scan Kinto views.
     kwargs = {}
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -55,14 +55,14 @@ def main(global_config, config=None, **settings):
 
     # Expose capability
     schema_enabled = asbool(
-        settings.get('experimental_collection_schema_validation')
+        settings['experimental_collection_schema_validation']
     )
     if schema_enabled:
         config.add_api_capability(
             "schema",
-            description="Enforce a schema for collection records.",
+            description="Validates collection records with JSON schemas.",
             url="http://kinto.readthedocs.org/en/latest/api/1.x/"
-            "collections.html#collection-json-schema")
+                "collections.html#collection-json-schema")
 
     # Scan Kinto views.
     kwargs = {}
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -53,6 +53,14 @@ def main(global_config, config=None, **settings):
     # Retro-compatibility with first Kinto clients.
     config.registry.public_settings.add('cliquet.batch_max_requests')
 
+    # Expose capability
+    schema_enabled = asbool(settings.get('experimental_collection_schema_validation'))
+    if schema_enabled:
+        config.add_api_capability("schema",
+                                  description="Enforce a schema for collection records.",
+                                  url="http://kinto.readthedocs.org/en/latest/api/1.x/"
+                                      "collections.html#collection-json-schema")
+
     # Scan Kinto views.
     kwargs = {}
     flush_enabled = asbool(settings.get('flush_endpoint_enabled'))
--- a/kinto/tests/test_views_hello.py
+++ b/kinto/tests/test_views_hello.py
@@ -28,18 +28,11 @@ class HelloViewTest(BaseWebTest, unittest.TestCase):
         self.assertEqual(response.json['user']['bucket'],
                          '23bb0efc-e80d-829e-6757-79d41e16640f')
 
-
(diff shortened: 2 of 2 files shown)


kinto/__init__.py
```diff
@@ -53,6 +53,13 @@ def main(global_config, config=None, **settings):
     # Retro-compatibility with first Kinto clients.
     config.registry.public_settings.add('cliquet.batch_max_requests')
 
+    # Expose capability
+    schema_enabled = asbool(settings.get('experimental_collection_schema_validation'))
+    if schema_enabled:
+        config.add_api_capability("attachments",
```
Natim: It should be schema right?

Something like that maybe:

```python
config.add_api_capability("schema",
                          description="Enforce a schema for collection records.",
                          url="http://kinto.readthedocs.org/en/latest/api/1.x/"
                              "collections.html#collection-json-schema")
```

almet: I confirm that this should be "schema" (or maybe "experimental.schema" ?)

leplatrem: I wouldn't mention experimental here

Natim:
Thank you for that it looks great.

Maybe we could add the tests in this existing file: https://github.com/Kinto/kinto/blob/master/kinto/tests/test_views_hello.py ?

kinto/tests/test_views_hello.py
```diff
@@ -27,3 +27,39 @@ def test_returns_bucket_id_and_url_if_authenticated(self):
...
+        settings = super(HelloViewSettingFalseTest, self).get_app_settings(
+            additional_settings
+        )
+        settings['experimental_collection_schema_validation'] = False
+        return settings
+
+    def test_capability_is_exposed_if_setting_is_not_set(self):
+        resp = self.app.get('/')
+        capabilities = resp.json['capabilities']
+        self.assertEqual(dict(), capabilities)
```
Natim: Maybe ``self.assertNotIn('schema', capabilities)``

Natim:
Thank you for this it is great. I will let @leplatrem review it for the wording part however the feature and tests seems to be there for me.

You may want to tacle #487 which is more or less the same idea.

Natim:
Ok I have pushed my reviews.

kinto/__init__.py
```diff
@@ -53,6 +53,17 @@ def main(global_config, config=None, **settings):
...
     config.registry.public_settings.add('cliquet.batch_max_requests')
 
+    # Expose capability
+    schema_enabled = asbool(
+        settings.get('experimental_collection_schema_validation')
+    )
+    if schema_enabled:
+        config.add_api_capability(
+            "schema",
+            description="Enforce a schema for collection records.",
```
almet: Add some validation to pushed records using a json schema

kinto/__init__.py
```diff
@@ -53,6 +53,17 @@ def main(global_config, config=None, **settings):
...
+            url="http://kinto.readthedocs.org/en/latest/api/1.x/"
+            "collections.html#collection-json-schema")
```
almet: this should be aligned with the line before it.

kinto/tests/test_views_hello.py
```diff
@@ -27,3 +27,39 @@ def test_returns_bucket_id_and_url_if_authenticated(self):
...
         self.assertEqual(response.json['user']['bucket'],
                          '23bb0efc-e80d-829e-6757-79d41e16640f')
+
+
+class HelloViewSchemaCapabilityExposedTest(BaseWebTest, unittest.TestCase):
+
+    def get_app_settings(self, additional_settings=None):
+        settings = super(
+            HelloViewSchemaCapabilityExposedTest,
+            self).get_app_settings(additional_settings)
```
leplatrem: inconsistent indentation with test below

almet:
looks good to me, added a few nitpicks.

kinto/tests/test_views_hello.py
```diff
@@ -27,3 +27,39 @@ def test_returns_bucket_id_and_url_if_authenticated(self):
...
+        self.assertIn('schema', capabilities)
+        expected = {
+            "description": "Enforce a schema for collection records.",
+            "url": "http://kinto.readthedocs.org/en/latest/api/1.x/"
+                   "collections.html#collection-json-schema",
+        }
+        self.assertEqual(expected, capabilities['schema'])
+
+
+class HelloViewSchemaCapabilityTest(BaseWebTest, unittest.TestCase):
```
leplatrem: instead of having two tests suite to vary the setting value, you can do:
```python
app = self.make_app(settings={...})
resp = app.get('/')
```

kinto/__init__.py
```diff
@@ -53,6 +53,17 @@ def main(global_config, config=None, **settings):
     # Retro-compatibility with first Kinto clients.
     config.registry.public_settings.add('cliquet.batch_max_requests')
 
+    # Expose capability
+    schema_enabled = asbool(
+        settings.get('experimental_collection_schema_validation')
```
leplatrem: nit: since default settings are loaded, you can do  `settings['experimental...]`

kinto/__init__.py
leplatrem: «Validates collection records with JSON schemas»

kinto/__init__.py
lavish205: Acc. to idiomatic python, .get() is the best way to access value of any particular key from dict.

kinto/tests/test_views_hello.py
lavish205: but we need different settings for both the test cases. It has been discussed with @Natim

Natim:
@lavish205 this is great thank you :+1:

kinto/__init__.py
```diff
@@ -53,6 +53,17 @@ def main(global_config, config=None, **settings):
     # Retro-compatibility with first Kinto clients.
     config.registry.public_settings.add('cliquet.batch_max_requests')
 
+    # Expose capability
+    schema_enabled = asbool(
+        settings['experimental_collection_schema_validation']
```
almet: apparently, @leplatrem is saying that we don't need to convert to bool here since this was done already.

leplatrem: nope it is not converted :)

leplatrem:
Thanks @lavish205 for your patience!

Issue #490, opened 2016-03-04
kinto start takes too much time.


leplatrem:
It can sometimes take more than 2 seconds, as reported by @n1k0 on JS clients integration tests.

I investigated a bit and found out, that on my machine, loading the entrypoint takes more than 1 second already:
```
$ time python -c "from pkg_resources import load_entry_point; load_entry_point('kinto', 'console_scripts', 'kinto')"
python -c   0,96s user 0,16s system 99% cpu 1,132 total
```

In comparison, `pserve` takes 200msec: 
```
$ time python -c "from pkg_resources import load_entry_point; load_entry_point('pyramid', 'console_scripts', 'pcreate')"
python -c   0,18s user 0,09s system 98% cpu 0,272 total
```

I realized that moving `import requests` from `cliquet.initialization` imports [PR](https://github.com/mozilla-services/cliquet/pull/674), and remove `import pip` from `kinto.__main__` I could reduce by half:
```
$ time python -c "from pkg_resources import load_entry_point; load_entry_point('kinto', 'console_scripts', 'kinto')"
python -c   0,36s user 0,18s system 98% cpu 0,543 total
```

I knew this was not going to speed up the `kinto start` command too much. I tracked down and noticed the `__main__:main` was executed twice because of `--reload` argument.

n1k0:
:+1: (please)

PR #491, opened 2016-03-04
Speed-up startup (ref #490)

leplatrem:
@n1k0 r?

(this would explain why the time had increased recently: #367)

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -3,7 +3,6 @@ import argparse
 import os
 import sys
 
-import pip
 from six.moves import input
 from cliquet.scripts import cliquet
 from pyramid.scripts import pserve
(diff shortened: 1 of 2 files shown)


n1k0:
I'm lacking proper context when it comes to the amount of performance impact this has, though importing `pip` only when necessary sounds always good to get anyway :)

r+

kinto/__main__.py
```diff
@@ -76,6 +75,7 @@ def main(args=None):
             try:
                 import psycopg2  # NOQA
             except ImportError:
+                import pip
```
almet: does this really speedup the startup time? If yes, do you know why? (importing pip shouldn't take that long!)

PR #493, opened 2016-03-05
Add date time and version to config file (fixes #488)

### LINKED ISSUE #488: Add creation date and current kinto version to generated config file
Using comments for example:

```ini
# Created at Thu, 03 Mar 2016 17:02:37 +0100
# Using Kinto version 1.11.2

[server:main]
use = egg:waitress#main
host = 0.0.0.0
port = 8888

```

ayusharma:
Refs  #488

### DIFF
--- a/kinto/config/__init__.py
+++ b/kinto/config/__init__.py
@@ -1,9 +1,11 @@
 import os
 import codecs
 from time import strftime
+
 from cliquet import utils as cliquet_utils
-from kinto import __version__
+
 from kinto import logger
+from kinto import __version__
 
 HERE = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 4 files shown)


kinto/config/__init__.py
```diff
@@ -1,8 +1,8 @@
 import os
 import codecs
-
```
almet: The spaces are actually legitimate here: in python, you:

- first import standard library modules (hint: time is one)
- then import 3rd party library modules
- then import the modules from you project (here kinto).

Between each block you need an empty line.

almet:
Travis CI is showing the tests that fail. You should see this fail also on your machine (when running `make test`)

kinto/tests/test_config.py
```diff
@@ -68,7 +74,9 @@ def test_init_postgresql_values(self, mocked_render_template):
             'permission_backend': 'cliquet.permission.postgresql',
             'storage_url': postgresql_url,
             'cache_url':  postgresql_url,
-            'permission_url': postgresql_url
+            'permission_url': postgresql_url,
+            'kinto_version': __version__,
+            'config_file_timestamp': strftime('%a, %d %b %Y %H:%M:%S %z')
```
leplatrem: nitpick: for the tests you can use a hard-coded version if you don't test its output

kinto/tests/test_configuration/test.ini
```diff
@@ -1,3 +1,7 @@
+# Created at config_file_timestamp
```
leplatrem: Why do you have this `config_file_timestamp` string here ? I thought the file content was compared somewhere else, no ?

Natim:
Thank you @ayusharma :)

kinto/tests/test_configuration/test.ini
Natim: I tried to change that and tests were failing.

kinto/tests/test_config.py
Natim: This is part of testing that the current date is correctly passed to the template.

PR #495, opened 2016-03-07
Possibility to exclude default bucket plugin (fixes #466)

### LINKED ISSUE #466: Explain in the doc how to disable default_buckets and bucket creation from authenticated users.
From a discussion with @chrismbeckett on IRC:

1) Anyone can register for my application and get an auth token that allows them to login to Kinto. With personal buckets, someone could figure out how to start storing anything they want. I am concerned I may need to *turn off* personal buckets since they may be a security risk.

2) Same for buckets. I am probably going to lock down creating buckets so it is done by my auth server when an account is created.

woovar:
To solve issue https://github.com/Kinto/kinto/issues/466, in the configuration the 'default-bucket' plugin (personal buckets) can be excluded.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,9 @@ This document describes changes between each past release.
 - Allow buckets to store arbitrary properties. (#462)
 - Speed-up startup (ref #490)
 
+**Breaking changes**
+- kinto.plugins.default_bucket plugin is no longer assumed. We invite users to check that the kinto.plugins.default_bucket is present in the includes setting if they expect it. (ref #495)
+
 
 1.11.2 (2016-02-03)
 ===================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ This document describes changes between each past release.
 - Speed-up startup (ref #490)
 
 **Breaking changes**
-- kinto.plugins.default_bucket plugin is no longer assumed. We invite users to check that the kinto.plugins.default_bucket is present in the includes setting if they expect it. (ref #495)
+- ``kinto.plugins.default_bucket`` plugin is no longer assumed. We invite users to check that the ``kinto.plugins.default_bucket`` is present in the ``includes`` setting if they expect it. (ref #495)
 
 
 1.11.2 (2016-02-03)
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -438,12 +438,6 @@ to create your own plugin.
 
 .. _configuring-notifications:
 
-It is also possible to exclude the default bucket plugin:
-
-.. code-block:: ini
-
-    kinto.excludes = kinto.plugins.default_bucket
-
 Notifications
 =============
 
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -438,6 +438,12 @@ to create your own plugin.
 
 .. _configuring-notifications:
 
+It is also possible to exclude the default bucket plugin:
+
+.. code-block:: ini
+
+    kinto.excludes = kinto.plugins.default_bucket
+
 Notifications
 =============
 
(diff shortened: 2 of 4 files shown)


kinto/__init__.py
```diff
@@ -47,7 +47,8 @@ def main(global_config, config=None, **settings):
 
     # In Kinto API 1.x, a default bucket is available.
     # Force its inclusion if not specified in settings.
-    if 'kinto.plugins.default_bucket' not in settings['includes']:
+    if 'kinto.plugins.default_bucket' not in settings['includes'] \
```
almet: I would assign the key you're looking for into a variable first, and then use it in the comparisons.

almet:
Thanks for proposing this. It misses tests that show that this is working like it should.

Natim:
I wonder if we should add an exclude option rather than just removing the fact that default_bucket is added if not present in ``kinto.includes``.

What you did is is a nice way to not change the behavior and still let people deactivate the default bucket.

But I think we shouldn't add the default_bucket if it is not present in the kinto.includes.

leplatrem:
Thanks @woovar !

Indeed, as @Natim  said, that's the best way to keep backward compatibility for users who expect the default bucket to work without having it in their configuration.

But since the plugin is present in the default configuration file (output with `kinto init`) I think we can even get rid of these lines and avoid the `exclude` setting.

It would require to add a *breaking changes* entry in the changelog, to invite users to check that the `kinto.plugins.default_bucket` is present in the `includes` setting if they expect it.

woovar:
@Natim @leplatrem That would be great and is much cleaner. I assumed since those lines were in the code, we absolutely wanted backwards compatibility. But if we can get rid of the requirement, it will become much clearer. Just removing the default bucket plugin from the configuration will stop the default bucket behavior. I can write tests for that and remove the appropriate code in __init__.py in this PR.

docs/configuration/settings.rst
```diff
@@ -438,6 +438,12 @@ to create your own plugin.
 
 .. _configuring-notifications:
 
+It is also possible to exclude the default bucket plugin:
+
+.. code-block:: ini
+
+    kinto.excludes = kinto.plugins.default_bucket
+
```
leplatrem: This can now be removed :)

kinto/tests/test_views_disable_default.py
```diff
@@ -0,0 +1,47 @@
...
+                      MINIMALIST_RECORD)
+
+
+class DisableDefaultBucketViewTest(BaseWebTest, unittest.TestCase):
+
+    test_url = '/buckets/default'
+
+    def setUp(self):
+        super(DisableDefaultBucketViewTest, self).setUp()
+        self.events = []
```
leplatrem: you don't need to override `setUp`, `tearDown`, `_get_test_app` or `get_app_settings` since you don't change their behaviour

CHANGELOG.rst
```diff
@@ -13,6 +13,9 @@ This document describes changes between each past release.
 - Allow buckets to store arbitrary properties. (#462)
 - Speed-up startup (ref #490)
 
+**Breaking changes**
+- kinto.plugins.default_bucket plugin is no longer assumed. We invite users to check that the kinto.plugins.default_bucket is present in the includes setting if they expect it. (ref #495)
```
leplatrem: nitpick : use ` ```` ` around setting names and values

leplatrem:
Great ! Thanks a lot for this contribution !

There is a conflict in the README file. I can merge and rebase this branch for you if you don't want to bother with that. Let me know :)

woovar:
@leplatrem If you could do that, that would be great. Thanks!

Natim:
I have rebased it here: https://github.com/Kinto/kinto/compare/woovar-495?expand=1

Natim:
Tests are there: https://travis-ci.org/Kinto/kinto/builds/114451183

Natim:
Superseeded by #499

PR #496, opened 2016-03-07
flush_endpoint capability added if flush_enabled (fixes #496)

lavish205:


Natim:
Thanks you @lavish205 for this. I will let @leplatrem review it :+1:

kinto/__init__.py
```diff
@@ -67,7 +67,16 @@ def main(global_config, config=None, **settings):
     # Scan Kinto views.
     kwargs = {}
     flush_enabled = asbool(settings.get('flush_endpoint_enabled'))
-    if not flush_enabled:
+
+    if flush_enabled:
+        config.add_api_capability(
+            "flush_endpoint",
```
leplatrem: nitpick: I wonder if we should name it `flush` only (?)

Natim: With @almet we though of ``flush``, ``flush_endpoint``, ``flush_endpoint_enabled``.
``flush_endpoint`` seemed more explicit that just ``flush``

In the future we should define a way to phrase:

- The kinto server has got the capability to have ``attachments``
- The kinto server has got the capability to have ``schema``
- The kinto server has got the capability to have ``changes``
- The kinto server has got the capability to have ``flush_endpoint``

leplatrem: Makes sense.

PR #461, opened 2016-02-25
Kinto creates blank record on Content-Type:text/plain

Natim:
As Kinto is JSON Datastore and whenever we POST text/plain instead of JSON it creates a blank record in DB instead of rejecting text/plain requests. 

Request: 
```
echo '{"data": {"description": "xyz"}}' | http POST http://localhost:8080/v1/buckets/default/collections/test-collection/records Content-Type:text/plain  -v --auth 'token:my-secret'
```

```
POST /v1/buckets/default/collections/test-collection/records HTTP/1.1
Accept: application/json
Accept-Encoding: gzip, deflate
Authorization: Basic dG9rZW46bXktc2VjcmV0
Connection: keep-alive
Content-Length: 33
Content-Type: text/plain
Host: localhost:8080
User-Agent: HTTPie/0.9.2

{"data": {"description": "xyz"}}

``` 

Response :

```
HTTP/1.1 201 Created
Access-Control-Expose-Headers: Retry-After, Content-Length, Alert, Backoff
Connection: keep-alive
Content-Length: 187
Content-Type: application/json; charset=UTF-8
Date: Thu, 25 Feb 2016 10:46:57 GMT
Server: nginx/1.8.0

{
    "data": {
        "id": "1227cf0c-66a5-4620-a178-255e7f479f28",
        "last_modified": 1456417017195
    },
    "permissions": {
        "write": [
            "basicauth:157beded4ea96ad4a21ed7fd91fff2a92a5daf47f054877d2e9768259e61600c"
        ]
    }
}
```

PR #498, opened 2016-03-07
Text plain body should be rejected with an error

leplatrem:
Revamp of #461 

r? @Natim


Natim:
It works for me if tests passes :+1:

PR #503, opened 2016-03-08
Add default bucket capability

leplatrem:
@Natim r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,8 +20,6 @@ This document describes changes between each past release.
 - Server now returns 415 error response if client cannot accept JSON response (#461, mozilla-services/cliquet#667)
 - Server now returns 415 error response if client does not send JSON request (#461, mozilla-services/cliquet#667)
 - Add the ``__lbheartbeat__`` endpoint, for load balancer membership test.
-- Added the ``flush_endpoint``, ``schema`` and ``default_bucket`` capabilities
-  if enabled in settings
 
 Protocol is now version 1.4. See `API changelog <http://kinto.readthedocs.org/en/latest/api/>`_.
 
@@ -56,6 +54,8 @@ Protocol is now version 1.4. See `API changelog <http://kinto.readthedocs.org/en
   `See more details <http://cliquet.readthedocs.org/en/latest/reference/notifications.html>`_.
 - Track execution time on StatsD for each authentication sub-policy (mozilla-services/cliquet#639)
 - Default console log renderer now has colours (mozilla-service/cliquet#671)
+- Added the ``flush_endpoint``, ``schema`` and ``default_bucket`` to the capabilities
+  if enabled in settings
 
 **Bug fixes**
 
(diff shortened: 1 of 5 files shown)


kinto/plugins/default_bucket/test_plugin.py
```diff
@@ -207,6 +207,19 @@ def test_formatted_error_are_passed_through(self):
             self.assertEqual(resp.body, response.body)
 
 
+class HelloViewTest(BaseWebTest, unittest.TestCase):
+
+    def test_returns_bucket_id_and_url_if_authenticated(self):
+        response = self.app.get('/', headers=self.headers)
+        self.assertEqual(response.json['user']['bucket'],
```
Natim: Should we move that information in the capability itself?

Natim:
r+ (comment should not block this as it is the current behavior)

CHANGELOG.rst
```diff
@@ -20,6 +20,8 @@ This document describes changes between each past release.
 - Server now returns 415 error response if client cannot accept JSON response (#461, mozilla-services/cliquet#667)
 - Server now returns 415 error response if client does not send JSON request (#461, mozilla-services/cliquet#667)
 - Add the ``__lbheartbeat__`` endpoint, for load balancer membership test.
+- Added the ``flush_endpoint``, ``schema`` and ``default_bucket`` capabilities
```
almet: I'm not sure this is part of the protocol (the protocol keeps unchanged here, just more capabilities have been added to it)

docs/api/index.rst
```diff
@@ -24,6 +24,8 @@ Changelog
 - Return 415 error response if client cannot accept JSON response
 - Return 415 error response if client does not send JSON request
 - Add the ``GET /v1/__lbheartbeat__`` endpoint, for load balancer membership test.
+- Add the ``flush_endpoint``, ``schema`` and ``default_bucket`` capabilities
```
almet: ditto.

CHANGELOG.rst
Natim: Well it has been added as a capability while it wasn't the case before so we can now look for it here.
For me it is fine to say the protocol changed.

almet: But it hasn't! The protocol isn't about what's inside the resources, but how to access them, and the mean of access is exactly the same as before.

Natim: > what's inside the resources, but how to access them, and the mean of access is exactly the same as before.

I mean the protocol of the home page changed and how we are getting the information of it ``default_bucket``, ``flush_endpoint`` and ` schema`` are activated

Natim: It doesn't hurt anyone to add this information in the CHANGELOG.

almet: I agree, but It's just not in the right section: the protocol is unchanged.

almet:
Thanks!

kinto/plugins/default_bucket/test_plugin.py
leplatrem: It is not a bad idea, but it would be a protocol breaking change. #lazy

CHANGELOG.rst
leplatrem: apparently @n1k0 does not agree with what we said here. So I will update the CHANGELOG in the release and keep a mention in the protocol changes.

n1k0: Yes, thank you! :heart:

almet: Can you be a bit more explicit ? Why should we consider this a protocol change?

PR #504, opened 2016-03-10
Rename ProtectedResource in ShareableResource.

almet:
The former is now deprecated.

r? @leplatrem @magopian


leplatrem:
Good catch!

r+

PR #512, opened 2016-03-17
Recreate objects after deletion fails

Natim:
kinto.py is struggeling recreating collection, buckets, groups and records with the ``safe=True`` flag if they have been deleted.

These are the test cases that are covering the matter.

### DIFF
--- a/kinto/tests/test_views_groups.py
+++ b/kinto/tests/test_views_groups.py
@@ -158,6 +158,13 @@ class GroupManagementTest(BaseWebTest, unittest.TestCase):
         self.assertEquals(self.permission.user_principals('mat'),
                           {group_url})
 
+    def test_groups_can_be_created_after_deletion(self):
+        self.app.delete(self.group_url, headers=self.headers)
+        headers = self.headers.copy()
+        headers['If-None-Match'] = '*'
+        self.app.put_json(self.group_url, MINIMALIST_GROUP,
+                          headers=headers, status=201)
+
 
 class InvalidGroupTest(BaseWebTest, unittest.TestCase):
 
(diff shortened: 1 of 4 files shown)


leplatrem:
I could fix the problem easily :)

kinto/tests/test_views_records.py
```diff
@@ -227,3 +227,12 @@ def test_records_should_accept_client_accept(self):
                      MINIMALIST_RECORD,
                      headers=headers,
                      status=200)
+
+    def test_records_can_be_created_after_deletion(self):
+        self.app.delete(self.record_url,
+                        headers=self.headers,
+                        status=200)
```
leplatrem: nit: 200 is superfluous

kinto/tests/test_views_groups.py
```diff
@@ -158,6 +158,13 @@ def test_group_member_removal_updates_user_principals_with_patch(self):
         self.assertEquals(self.permission.user_principals('mat'),
                           {group_url})
 
+    def test_groups_can_be_created_after_deletion(self):
+        self.app.delete(self.group_url, headers=self.headers)
```
leplatrem: The group is not created in setup, so you should create it before deleting it.

Natim:
The problem have been merged into Cliquet master.

Natim:
Tests are passing with cliquet-master. We can/should wait for a cliquet release before merging.

almet:
Refs https://github.com/mozilla-services/cliquet/commit/2c2f0337923272a52caa3a0e32a1bfd23f8e374b

Natim:
cliquet 3.1.2 have been released with this fix.

Natim:
Tests are now green :+1:

Issue #515, opened 2016-03-25
It doesn't seems possible to configure Redis event listener from environment variables.


happy-tanuki:
I tried to [get-stated](http://kinto.readthedocs.org/en/latest/get-started.html#using-docker-compose) and It's fine. kinto is great. 

Next I want to use the redis listener in the [Notifications section](http://kinto.readthedocs.org/en/latest/configuration/settings.html#notifications).
The some lines were added to docker-compose.yml.

    redis:
        image: redis
    web:
        ....
        links:
            - db
            - redis
        ....
        environment:
            ....
            KINTO_EVENT_LISTENERS: redis
            KINTO_EVENT_LISTENERS_REDIS_USE: cliquet.events.redis
            KINTO_EVENT_LISTENERS_REDIS_URL: redis://redis:6379/0
            KINTO_EVENT_LISTENERS_REDIS_POOL_SIZE: 5
            KINTO_EVENT_LISTENERS_REDIS_LISTNAME: queue

When `docker-compose up`, the error occurred.

    web_1   | Traceback (most recent call last):
    web_1   |   File "/usr/local/lib/python3.5/dist-packages/cliquet/initialization.py", line 399, in setup_listeners
    web_1   |     listener = listener_mod.load_from_config(config, prefix)
    web_1   | AttributeError: module 'redis' has no attribute 'load_from_config'

When the value of `KINTO_EVENT_LISTENERS` was changed to `cliquet.listeners.redis`, kinto was launched. (`redis` or `cliquet.events.redis` in the document are wrong?).

When I update the value, the another error occurred.

    web_1   | 2016-03-25 02:01:13,648 ERROR [cliquet.initialization][waitress] "POST  /v1/buckets/.../records" 201 (11 ms) Unable to send the payload to Redis agent=curl/7.43.0 authn_type=basicauth collection_id=record collection_timestamp=1458871273636 errno=None exception=Traceback (most recent call last):
    ...
    web_1   |   File "/usr/local/lib/python3.5/dist-packages/redis/connection.py", line 538, in send_packed_command
    web_1   |     self.connect()
    web_1   |   File "/usr/local/lib/python3.5/dist-packages/redis/connection.py", line 442, in connect
    web_1   |     raise ConnectionError(self._error_message(e))
    web_1   | redis.exceptions.ConnectionError: Error 111 connecting to localhost:6379. Connection refused.

[...]

PR #516, opened 2016-03-25
Fix the module name of Redis event listener

happy-tanuki:
In `Notifications` section I fixed the value of `kinto.event_listeners.redis.use`. #515


Natim:
Thanks @happy-tanuki :)

PR #523, opened 2016-03-31
The API is v1.x, and not 1.1

almet:
pair= @magopian


almet:
This breaks the links for people who have bookmarks on 1.1 links.

leplatrem:
> This breaks the links 

It only breaks the anchor right?

PR #537, opened 2016-04-07
Document how to upgrade Kinto

magopian:
@natim r?

### DIFF
--- a/docs/get-started.rst
+++ b/docs/get-started.rst
@@ -182,6 +182,25 @@ Then install the package using the default configuration:
 The server should now be running on http://localhost:8888
 
 
+Upgrading Kinto
+---------------
+
+Whenever you need to upgrade Kinto, first make sure that the server is stopped,
+and that the virtualenv is activated:
+
+::
+
+    source env/bin/activate
+
+Now upgrade Kinto (and its dependencies) using the following command:
+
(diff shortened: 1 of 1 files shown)


docs/get-started.rst
```diff
@@ -182,6 +182,25 @@ Then install the package using the default configuration:
...
+
+::
+
+    source env/bin/activate
+
+Now upgrade Kinto (and its dependencies) using the following command:
+
+::
+
+    pip install --upgrade kinto
```
Natim: It doesn't work if you want to upgrade cliquet does it?

magopian: Yes it does because [Cliquet's version isn't pinned](https://github.com/Kinto/kinto/blob/master/setup.py#L21). I've double checked by running this command, and it does indeed upgrade all the dependencies that don't have pinned versions.

Natim:
r+

leplatrem:
@magopian Thanks for taking care of this :)

Something crucial is missing: a call to `migrate` in case the db schema was changed. Maybe a mention about semver idea of breaking changes would be a plus: like read changelog... ;)

Also, a nit, but should the upgrade docs be in *get started* ? Maybe *production* ?

/cc @Natim

PR #541, opened 2016-04-12
Allow groups to store arbitrary properties. (fixes #469)

### LINKED ISSUE #469: Metadata on Groups
Similarily to how you can store extra properties (metadata) on a collection, it would be useful to be able to do this with groups.

In my applications, almost everything is dynamic. Users can create groups on the fly, rename them, etc., so I tend to use generated ID's for everything. It would be nice to be able to set a title and description on groups for UI presentation.

Right now I have to create a collection for storing group metadata separately from the actual group.

leplatrem:
@magopian r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ This document describes changes between each past release.
 
 - Relax content-type validation when no body is posted (fixes #507)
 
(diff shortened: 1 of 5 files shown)


almet:
Neat! LGTM, r+

PR #552, opened 2016-04-24
small typo fix

enkidulan:


docs/get-started.rst
```diff
@@ -138,7 +138,7 @@ On Debian / Ubuntu based systems::
 
 On RHEL-derivatives::
 
-    apt-get install libffi-devel openssl-devel python-devel python-virtualenv
+    dnf install libffi-devel openssl-devel python-devel python-virtualenv
```
almet: For those who don't know what DNF is:

> DNF was introduced in Fedora 18, and it has been the default package manager for Fedora since version 22
> — Wikipedia

almet:
Thanks for the contribution!

Natim:
Apparently dnf is not present on CentOS 6 :confused: 

@enkidulan any idea why you choose dnf rather than yum here?

enkidulan:
@Natim that is a good point, I guess it may have sense to mention both in docs. But keep in mind that
`dnf` will replace `yum` in all RHEL-derivatives eventually as it is a next generation of `yum`. Fedora uses it as a default package manager at least for 3 last releases, not sure about other distributives.

PR #553, opened 2016-04-26
Add link to English version of the article

glasserc:
I didn't even realize this article was already available in English and was just reading it in French.


almet:
Thanks for detecting this. I believe we can just link to the english version since the docs are in english. Also there is a note at the very top of the articles when it's available in french, so I believe that should be enough.

PR #576, opened 2016-05-02
kinto executable fails if no arg is passed to it

Natim:
```
$ .venv/bin/pip install kinto
$ .venv/bin/kinto --ini test/kinto.ini
Traceback (most recent call last):
  File ".venv/bin/kinto", line 11, in <module>
    sys.exit(main())
  File "/home/niko/work/kinto.js/.venv/lib/python3.5/site-packages/kinto/__main__.py", line 58, in main
    if args['which'] == 'init':
KeyError: 'which'
```

Versions:

```
$ .venv/bin/kinto --version
2.1.1
$ .venv/bin/python --version
Python 3.5.1+
```

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -54,16 +54,16 @@ def main(args=None):
                               default=False)
     parser_start.set_defaults(which='start')
 
-    args = vars(parser.parse_args(args))
+    parsed_args = vars(parser.parse_args(args))
 
-    config_file = args['ini_file']
+    config_file = parsed_args['ini_file']
 
-    if args['which'] == 'init':
+    if parsed_args['which'] == 'init':
         if os.path.exists(config_file):
             print("%s already exists." % config_file, file=sys.stderr)
             return 1
 
-        backend = args['backend']
+        backend = parsed_args['backend']
         if not backend:
             while True:
                 prompt = ("Select the backend you would like to use: "
@@ -86,13 +86,13 @@ def main(args=None):
                 import pip
                 pip.main(['install', "cliquet[postgresql]"])
 
-    elif args['which'] == 'migrate':
+    elif parsed_args['which'] == 'migrate':
         env = bootstrap(config_file)
         cliquet.init_schema(env)
 
-    elif args['which'] == 'start':
+    elif parsed_args['which'] == 'start':
         pserve_argv = ['pserve', config_file]
-        if args['reload']:
+        if parsed_args['reload']:
             pserve_argv.append('--reload')
         pserve.main(pserve_argv)
 
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -52,13 +52,21 @@ def main(args=None):
                               default=False)
     parser_start.set_defaults(which='start')
 
-    args = vars(parser.parse_args())
+    try:
+        args = vars(parser.parse_args(args))
+    except SystemExit:  # pragma: no cover for python2
+        return 10
+
     config_file = args['ini_file']
+
+    if 'which' not in args:  # pragma: no cover
+        parser.print_help(sys.stderr)
+        return 10
 
     if args['which'] == 'init':
         if os.path.exists(config_file):
-            print("%s already exist." % config_file, file=sys.stderr)
-            sys.exit(1)
+            print("%s already exists." % config_file, file=sys.stderr)
+            return 1
 
         backend = args['backend']
         if not backend:
@@ -77,9 +85,9 @@ def main(args=None):
 
         # Install postgresql libraries if necessary
         if backend == "postgresql":
-            try:
+            try:  # pragma: no cover
                 import psycopg2  # NOQA
-            except ImportError:
+            except ImportError:  # pragma: no cover
                 import pip
                 pip.main(['install', "cliquet[postgresql]"])
 
@@ -93,6 +101,4 @@ def main(args=None):
             pserve_argv.append('--reload')
         pserve.main(pserve_argv)
 
-
-if __name__ == "__main__":
-    main()
+    return 0
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -36,7 +36,9 @@ def main(args=None):
 
     subparsers = parser.add_subparsers(title='subcommands',
                                        description='valid subcommands',
+                                       dest='subcommand',
                                        help='init/start/migrate')
+    subparsers.required = True
 
     parser_init = subparsers.add_parser('init')
     parser_init.set_defaults(which='init')
@@ -58,10 +60,6 @@ def main(args=None):
         return 10
 
     config_file = args['ini_file']
-
-    if 'which' not in args:  # pragma: no cover
-        parser.print_help(sys.stderr)
-        return 10
 
     if args['which'] == 'init':
         if os.path.exists(config_file):
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -54,10 +54,7 @@ def main(args=None):
                               default=False)
     parser_start.set_defaults(which='start')
 
-    try:
-        args = vars(parser.parse_args(args))
-    except SystemExit:  # pragma: no cover for python2
-        return 10
+    args = vars(parser.parse_args(args))
 
     config_file = args['ini_file']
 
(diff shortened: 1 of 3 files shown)


kinto/__main__.py
```diff
@@ -77,9 +85,9 @@ def main(args=None):
 
         # Install postgresql libraries if necessary
         if backend == "postgresql":
-            try:
+            try:  # pragma: no cover
```
Natim: This is [supposed to be tested](https://github.com/Kinto/kinto/pull/576/files#diff-6e50bc18504011fcabd48bb239812b9fR60) but I don't know why the coverage is not calculated.

Natim: @leplatrem If you have feedback on this I would be interested.

kinto/__main__.py
```diff
@@ -52,13 +52,21 @@ def main(args=None):
...
 
-    args = vars(parser.parse_args())
+    try:
+        args = vars(parser.parse_args(args))
+    except SystemExit:  # pragma: no cover for python2
+        return 10
+
     config_file = args['ini_file']
 
+    if 'which' not in args:  # pragma: no cover
```
Natim: This is not supposed to append because parse_args should take care of it.

glasserc: It looks like this is caused by a bug in the argparse module: http://bugs.python.org/issue9253#msg186387

The workaround is to set a `dest` and set `subparsers.required = True`. I have done this in https://github.com/Kinto/kinto/pull/576/commits/84f74af1bd32e9cda251e64a38ae51ce813ddc80.

kinto/__main__.py
```diff
@@ -52,13 +54,17 @@ def main(args=None):
                               default=False)
     parser_start.set_defaults(which='start')
 
-    args = vars(parser.parse_args())
+    try:
+        args = vars(parser.parse_args(args))
+    except SystemExit:  # pragma: no cover for python2
+        return 10
```
Natim: Maybe we don't need this anymore then.

glasserc: I didn't understand the reason that this got added in the first place, so I'm not qualified to comment on that.. isn't this for the purposes of testing or something?

kinto/__main__.py
```diff
@@ -93,6 +99,4 @@ def main(args=None):
             pserve_argv.append('--reload')
         pserve.main(pserve_argv)
 
-
-if __name__ == "__main__":
-    main()
```
glasserc: Did we intend to remove this?

Natim: Yes, it is hard to test and because we are using entry_points I guess we don't need it anymore.

kinto/tests/test_main.py
```diff
@@ -0,0 +1,116 @@
...
+    def tearDown(self):
+        try:
+            os.remove('/tmp/kinto.ini')
+        except OSError:
+            pass
+
+    def test_cli_init_generate_configuration(self):
+        res = main(['--ini', '/tmp/kinto.ini', '--backend', 'memory', 'init'])
+        assert res == 0
+        assert os.path.exists('/tmp/kinto.ini')
```
leplatrem: nit: use constant

kinto/tests/test_main.py
```diff
@@ -0,0 +1,116 @@
...
+        with mock.patch('sys.stderr', new_callable=StringIO) as mock_stderr:
+            with pytest.raises(SystemExit) as excinfo:
+                main(['--ini', '/tmp/kinto.ini'])
+            assert 'INI_FILE' in mock_stderr.getvalue()
+            assert excinfo.value.code == 2
+
+    def test_cli_init_install_postgresql_dependencies_if_needed(self):
+        realimport = builtins.__import__
+
+        def myimport(name, *args, **kwargs):
```
leplatrem: nit: rename psycopg2_missing

kinto/tests/test_main.py
```diff
@@ -0,0 +1,116 @@
...
+            assert 'redis' in content
+
+    def test_fails_if_not_enough_args(self):
+        with mock.patch('sys.stderr', new_callable=StringIO) as mock_stderr:
+            with pytest.raises(SystemExit) as excinfo:
+                main(['--ini', '/tmp/kinto.ini'])
+            assert 'INI_FILE' in mock_stderr.getvalue()
+            assert excinfo.value.code == 2
+
+    def test_cli_init_install_postgresql_dependencies_if_needed(self):
```
leplatrem: nit: install**s**

kinto/tests/test_main.py
```diff
@@ -0,0 +1,116 @@
...
+            assert res == 1
+            content = mock_stderr.getvalue()
+            assert '/tmp/kinto.ini already exists.' in content
+
+    def test_cli_init_ask_for_backend_if_not_specified(self):
+        with mock.patch("kinto.__main__.input", create=True, return_value="2"):
+            res = main(['--ini', '/tmp/kinto.ini', 'init'])
+            assert res == 0
+
+    def test_cli_init_ask_until_backend_is_valid(self):
```
leplatrem: nit: ask**s**

kinto/__main__.py
glasserc: I was using it to test locally using `python kinto/__main__.py`, which is a little easier than running `pip install` every time I make a change..

kinto/tests/test_main.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class TestMain(unittest.TestCase):
+    def tearDown(self):
+        try:
+            os.remove('/tmp/kinto.ini')
+        except OSError:
+            pass
+
+    def test_cli_init_generate_configuration(self):
```
leplatrem: nit: generate**s**

kinto/__main__.py
```diff
@@ -52,13 +54,14 @@ def main(args=None):
                               default=False)
     parser_start.set_defaults(which='start')
 
-    args = vars(parser.parse_args())
+    args = vars(parser.parse_args(args))
```
n1k0: Note: /me wonders how it could even work before =)

Nit: Maybe we could create another var instead of overriding the existing `args` one? eg. `parsed_args`

leplatrem:
Glad you tackled this! Thanks!

The tests are a bit indigest, but I can't see any better way.

LGTM

kinto/tests/test_main.py
```diff
@@ -0,0 +1,114 @@
...
+            res = main(['--ini', '/tmp/kinto.ini',
+                        '--backend', 'memory', 'init'])
+            assert res == 1
+            content = mock_stderr.getvalue()
+            assert '/tmp/kinto.ini already exists.' in content
+
+    def test_cli_init_ask_for_backend_if_not_specified(self):
+        with mock.patch("kinto.__main__.input", create=True, return_value="2"):
+            res = main(['--ini', '/tmp/kinto.ini', 'init'])
+            assert res == 0
```
glasserc: Should this test check that the mock actually got called?

n1k0:
Looks good though many mocks everywhere; I suppose we don't have many alternative choices with argparse?

kinto/__main__.py
Natim: If you use ``pip install -e .`` then it should update when updating the file without having to restart ``pip install`` in between.

kinto/__main__.py
Natim: Actually by default parse_args use sys.argv automatically.

kinto/__main__.py
glasserc: This is awesome! Thanks!

kinto/__main__.py
leplatrem: shouldn't the `no cover` go to the `except` block ?

kinto/tests/test_main.py
Natim: Can I test if the content of the file actually contains the right backend?

Natim:
> Looks good though many mocks everywhere

Yes I use them to basically to validate the interaction with the user.
I guess it is still better than no tests in that case.

glasserc:
r+

Natim:
Tests are passing locally. (Travis dans les choux)

PR #544, opened 2016-05-02
a GET on "/v1" returns a 404, maybe it should return a 307 like on "/"?

Natim:
If you try to GET the url `/`, it gives you a `307 temporary redirect`. However, if you GET the url `/v1`, it simply returns a `404 not found`.

Is that expected?


Natim:
This should be merged after mozilla-services/cliquet#700 is released in the next cliquet version.

PR #578, opened 2016-05-02
get rid of javascript tutorial

### LINKED ISSUE #562: retitle javascript.rst "Guide to implementing a client"
The `javascript.rst` tutorial in this documentation is basically a low-level guide to implementing a client. I don't think it's useful considering the `HTTPie` examples strewn throughout the documentation, but if we're going to keep it, we should make it extremely clear that this code is meant for library developers and NOT for JavaScript application developers.

glasserc:
This covers stuff that is already better covered in JavaScript-specific
projects such as Kinto.js and kinto-client.

We may want to write a document that covers best practices in writing a
client, but this document isn't that, and it just distracts from actual
existing clients, so let's delete it.

Fixes #562.

r? @n1k0


n1k0:
r+!

PR #584, opened 2016-05-03
Mention Next-Page in CDN headers to let pass

leplatrem:
When I reach a paginated list through the CDN, I shall obtain a `Next-Page` header containing the full URL.

This pull-request adds the `Next-Page` to the list of headers to let pass on the CDN.

@Natim r?


Natim:
Good point.

PR #585, opened 2016-05-04
Rename get-started to install and polish first-steps doc

glasserc:
As someone with a CouchDB background, it took a while for me to figure out what the actors in a Kinto ecosystem were and where they lived. (Unlike in CouchDB, there's only one Kinto "database", although the clients act sort of like databases themselves.) This is meant to help clarify this and generally flesh out the first few pages of documentation that a new user will see.


glasserc:
While I was working on this, I started to wonder if a more useful "first steps" document might use a client library and hide this complexity from the user. That change was more invasive than what I was looking to do today, so I left it as-is for now, but I'm open to making that kind of change later if it seems like a good idea. (In that case, I guess the "first steps" page would become something like a tour of the Kinto API, perhaps for client implementors.)

docs/api/1.x/index.rst
```diff
@@ -17,7 +17,6 @@ Full detailed API documentation:
    records
    groups
    permissions
-   synchronisation
```
leplatrem: Is this a mistake or do you think the sync docs are not relevant ? https://github.com/Kinto/kinto/blob/master/docs/api/1.x/synchronisation.rst

docs/tutorials/index.rst
```diff
@@ -6,8 +6,9 @@ Tutorials
 .. toctree::
    :maxdepth: 1
 
-   get-started
+   install
    first-steps
+   synchronisation
```
leplatrem: Ok got it. We could move the file in the tutorials folder then

glasserc: Yes, I did that in https://github.com/Kinto/kinto/pull/585/commits/894f7c18df3abe636bca40bfc0349ec505a2c48a.

leplatrem:
> I started to wonder if a more useful "first steps" document might use a client library and hide this complexity from the user.

Yes good idea!

> [...] perhaps for client implementors

I don't think client implementors are the first target. (*Well, actually they would if this doc would be focused on the server itself and if we had some eco-system docs elsewhere.*).
An idea would be to focus on front-end/mobtile devs that don't want to bother with the server just yet. Maybe some recipes for some common use-cases ?

leplatrem:
r+ for me, if I'm not the only one to agree on the renaming of *get started* to *install* you can merge ;)

Thanks a lot!

almet:
This is great, thanks!

Issue #550, opened 2016-04-22
500 when validation fails with an integer value instead of string 


leplatrem:
The way we transform the validation errors of `jsonschema` to `cornice` is not rock-solid.
It looks like values and fields are sometimes mixed-up.
When an integer reaches the `name` parameter of `raise_invalid()` it fails miserably.

```diff
         except jsonschema_exceptions.ValidationError as e:
             field = e.path.pop() if e.path else e.validator_value.pop()
-            raise_invalid(self.request, name=field, description=e.message)
+            raise_invalid(self.request, name=str(field), description=e.message)
```

To reproduce this, define a schema with an enum of strings. And provide a record with a value as integer for the field.

leplatrem:
Removed easy-pick label since I could not reproduce.

leplatrem:
Closing until we reproduce.

PR #590, opened 2016-05-12
[WIP] Try to fix crash when validation type mismatch

leplatrem:
Could not reproduce bug #550


leplatrem:
Closing because the bug could not be reproduced.

PR #454, opened 2016-05-19
Authenticated user should be allowed to browse her buckets by default

Natim:
Kinto config:

```
[app:main]
use = egg:kinto

# Required by integration tests
kinto.flush_endpoint_enabled = true

# Required by basic auth
kinto.userid_hmac_secret = a-secret-string

[server:main]
use = egg:waitress#main
host = 0.0.0.0
port = 8888
```

Starting server

```
./.venv/bin/pserve test/kinto.ini
Starting server in PID 3230.
serving on http://0.0.0.0:8888
```

Listing buckets:

```
$ http :8888/v1/buckets -a a:
HTTP/1.1 403 Forbidden
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
Content-Length: 95
Date: Tue, 23 Feb 2016 08:58:07 GMT
Server: waitress

{
    "code": 403, 
    "errno": 121, 
    "error": "Forbidden", 
    "message": "This user cannot access this resource."
}
```

This is quite surprising, as one might expect being able to at least list his own default bucket in there.

If you create a custom bucket, then you're suddenly allowed to browser buckets:

```
$ http PUT :8888/v1/buckets/foo -a a:
HTTP/1.1 201 Created
Content-Length: 154
Date: Tue, 23 Feb 2016 09:03:09 GMT
Etag: "1456218189534"
Last-Modified: Tue, 23 Feb 2016 09:03:09 GMT
Server: waitress

{
    "data": {
        "id": "foo", 
        "last_modified": 1456218189534
    }, 
    "permissions": {
        "write": [
            "basicauth:a2125d3ab6f6b9509e1bc87932010af71cebfd50424e972934a0e4b5a1047f42"
        ]
    }
}

$ http :8888/v1/buckets -a a:        
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
Cache-Control: no-cache
Content-Length: 53
Date: Tue, 23 Feb 2016 09:03:17 GMT
Etag: "1456218189534"
Last-Modified: Tue, 23 Feb 2016 09:03:09 GMT
Server: waitress
Total-Records: 1

{
    "data": [
        {
            "id": "foo", 
            "last_modified": 1456218189534
        }
    ]
}
```

TL;DR: I think authenticated users should be able to list the buckets they have access to by default.

### DIFF
--- a/kinto/authorization.py
+++ b/kinto/authorization.py
@@ -145,4 +145,7 @@ class AuthorizationPolicy(core_authorization.AuthorizationPolicy):
 
 
 class RouteFactory(core_authorization.RouteFactory):
-    pass
+    def __init__(self, request):
+        super(RouteFactory, self).__init__(request)
+        if self.on_collection and self.resource_name == 'bucket':
+            self.force_empty_list = True
--- a/kinto/authorization.py
+++ b/kinto/authorization.py
@@ -1,5 +1,5 @@
 from kinto.core import authorization as core_authorization
-from pyramid.security import IAuthorizationPolicy
+from pyramid.security import IAuthorizationPolicy, Authenticated
 from zope.interface import implementer
 
 
@@ -145,7 +145,16 @@ class AuthorizationPolicy(core_authorization.AuthorizationPolicy):
 
 
 class RouteFactory(core_authorization.RouteFactory):
-    def __init__(self, request):
-        super(RouteFactory, self).__init__(request)
-        if self.on_collection and self.resource_name == 'bucket':
-            self.force_empty_list = True
+    pass
+
+
+class BucketRouteFactory(RouteFactory):
+    def fetch_shared_records(self, perm, principals, get_bound_permissions):
+        """Buckets list is authorized even if no object is accessible for
+        the current principals.
+        """
+        shared = super(BucketRouteFactory, self).fetch_shared_records(
+            perm, principals, get_bound_permissions)
+        if shared is None and Authenticated in principals:
+            self.shared_ids = []
+        return self.shared_ids
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -86,6 +86,9 @@ class AuthorizationPolicy(object):
                 principals,
                 get_bound_permissions=self.get_bound_permissions)
             allowed = len(shared_records) > 0
+            if not allowed and context.force_empty_list:
+                allowed = True
+                context.forced_empty_list = True
 
         return allowed
 
@@ -101,6 +104,7 @@ class RouteFactory(object):
     permission_object_id = None
     current_record = None
     get_shared_ids = None
+    force_empty_list = False
 
     method_permissions = {
         "head": "read",
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -85,10 +85,7 @@ class AuthorizationPolicy(object):
                 permission,
                 principals,
                 get_bound_permissions=self.get_bound_permissions)
-            allowed = len(shared_records) > 0
-            if not allowed and context.force_empty_list:
-                allowed = True
-                context.forced_empty_list = True
+            allowed = shared_records is not None
 
         return allowed
 
@@ -104,7 +101,6 @@ class RouteFactory(object):
     permission_object_id = None
     current_record = None
     get_shared_ids = None
-    force_empty_list = False
 
     method_permissions = {
         "head": "read",
@@ -121,7 +117,7 @@ class RouteFactory(object):
         self._check_permission = request.registry.permission.check_permission
 
         # Partial collections of ShareableResource:
-        self.shared_ids = []
+        self.shared_ids = None
 
         # Store service, resource, record and required permission.
         service = utils.current_service(request)
@@ -173,12 +169,25 @@ class RouteFactory(object):
         return self._check_permission(self.permission_object_id, *args, **kw)
 
     def fetch_shared_records(self, perm, principals, get_bound_permissions):
-        ids = self.get_shared_ids(
-            permission=perm,
-            principals=principals,
-            get_bound_permissions=get_bound_permissions)
-        # Store for later use in ``ShareableResource``.
-        self.shared_ids = [self.extract_object_id(id_) for id_ in ids]
+        """Fetch records that are readable or writable for the current
+        principals.
+
+        If no record is shared, it returns None.
+
+        .. warning::
+            This sets the ``shared_ids`` attribute to the context with the
+            return value. The attribute is then read by
+            :class:`kinto.core.resource.ShareableResource`
+        """
+        ids = self.get_shared_ids(permission=perm,
+                                  principals=principals,
+                                  get_bound_permissions=get_bound_permissions)
+        if len(ids) > 0:
+            # Store for later use in ``ShareableResource``.
+            self.shared_ids = [self.extract_object_id(id_) for id_ in ids]
+        else:
+            self.shared_ids = None
+
         return self.shared_ids
 
     def get_permission_object_id(self, request, record_id=None):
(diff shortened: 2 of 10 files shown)


Natim:
Ok I am not glad I add to change all this.
Feel free to comment. r? @leplatrem

kinto/core/resource/__init__.py
```diff
@@ -1128,6 +1128,9 @@ def _extract_filters(self, queryparams=None):
         filters = super(ShareableResource, self)._extract_filters(queryparams)
 
         ids = self.context.shared_ids
+        if getattr(self.context, "forced_empty_list", False):
+            ids = ['UNKNOWN']
```
Natim: Maybe there is a better way here so that we don't even bother triggering a storage query since we want an empty list here.

kinto/authorization.py
```diff
@@ -145,4 +145,7 @@ def get_bound_permissions(self, *args, **kwargs):
 
 
 class RouteFactory(core_authorization.RouteFactory):
-    pass
+    def __init__(self, request):
+        super(RouteFactory, self).__init__(request)
+        if self.on_collection and self.resource_name == 'bucket':
+            self.force_empty_list = True
```
leplatrem: This attribute is not defined if the condition is not met.

Instead, you could define another RouteFactory (e.g.  `BucketRouteFactory` with a class attribute like `allow_empty_list`)

kinto/core/authorization.py
```diff
@@ -86,6 +86,9 @@ def permits(self, context, principals, permission):
                 principals,
                 get_bound_permissions=self.get_bound_permissions)
             allowed = len(shared_records) > 0
+            if not allowed and context.force_empty_list:
+                allowed = True
+                context.forced_empty_list = True
```
leplatrem: Why do you mute the context here?

kinto/core/resource/__init__.py
```diff
@@ -245,6 +245,12 @@ def collection_get(self):
         self._raise_412_if_modified()
 
         headers = self.request.response.headers
+
+        if getattr(self.context, "forced_empty_list", False):
+            # Return an empty list.
+            headers['Total-Records'] = encode_header('0')
+            return self.postprocess([])
+
```
leplatrem: I would have preferred to keep this close to the existing (and only) link between context and resource : https://github.com/Kinto/kinto/blob/master/kinto/core/resource/__init__.py#L1122-L1135
For example, extract_filters could return something like (last_modified == -1) to return 0 records.

kinto/authorization.py
Natim: It is defined in the ``core_authorization.RouteFactory``

kinto/core/authorization.py
```diff
@@ -86,6 +86,9 @@ def permits(self, context, principals, permission):
                 principals,
                 get_bound_permissions=self.get_bound_permissions)
             allowed = len(shared_records) > 0
+            if not allowed and context.force_empty_list:
```
leplatrem: force is `allow` here

kinto/core/resource/__init__.py
```diff
@@ -245,6 +245,12 @@ def collection_get(self):
         self._raise_412_if_modified()
 
         headers = self.request.response.headers
+
+        if getattr(self.context, "forced_empty_list", False):
```
leplatrem: Oh I see, you have `force` and `forced`... hum, we should be able to simplify that

kinto/core/resource/__init__.py
Natim: Yes it is what I did in the first commit but it makes a query to the storage while we don't want to make one in that case.

kinto/views/buckets.py
```diff
@@ -1,7 +1,7 @@
 from kinto.core import resource
 from kinto.core.events import ResourceChanged, ACTIONS
 from pyramid.events import subscriber
-
+from kinto.authorization import RouteFactory
```
leplatrem: `from kinto.authorization import BucketRouteFactory`

kinto/core/resource/viewset.py
```diff
@@ -226,7 +228,7 @@ def get_view_arguments(self, endpoint_type, resource_cls, method):
 
     def get_service_arguments(self):
         args = super(ShareableViewSet, self).get_service_arguments()
-        args['factory'] = authorization.RouteFactory
+        args['factory'] = self.factory
```
leplatrem: :+1:

kinto/core/authorization.py
Natim: We need to detect that it was allowed while there were no shared_records.
If we don't know that we will endup having all the records as if we had the read permissions on buckets.

kinto/views/buckets.py
Natim: We thouht about doing that, but it is more the KintoRouteFactory than the BucketRouteFactory.

I think we shouldn't have multiple RouteFactory for a given project, it makes it harder to follow.

kinto/core/resource/viewset.py
Natim: Actually I think we should read that from the config.root_factory, I did a hack to get there but I wish we could fix it using the config one.

kinto/views/buckets.py
leplatrem: > we shouldn't have multiple RouteFactory for a given project

That's not a problem if their code is small. Having tests like `if self.resource_name == 'bucket'` bothers me.

Look nothing says here that we shouldn't have one per route:
http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#route-factories

kinto/core/resource/__init__.py
leplatrem: Yes, but it seems that in order to save a query we introduce some additional indirection :(

leplatrem:
We should work on this patch to simplify the approache/remove indirections...

kinto/tests/test_views_buckets.py
```diff
@@ -34,10 +34,11 @@ def test_buckets_names_can_have_underscores(self):
...
 
-    def test_nobody_can_list_buckets_by_default(self):
-        self.app.get(self.collection_url,
-                     headers=get_user_headers('alice'),
-                     status=403)
+    def test_everybody_can_list_buckets_by_default(self):
+        resp = self.app.get(self.collection_url,
+                            headers=get_user_headers('alice'),
+                            status=200)
+        assert resp.json['data'] == []
```
leplatrem:
```py

    def test_no_anonymous_can_list_buckets_by_default(self):
        self.app.get(self.collection_url, status=403)

    def test_anybody_can_list_buckets_by_default(self):
        resp = self.app.get(self.collection_url,
                            headers=get_user_headers('alice'))
        # But it contains no record if no read permission.
        self.assertEqual(len(resp.json['data']), 0)
```

kinto/views/buckets.py
```diff
@@ -11,9 +11,9 @@ class Options:
 
 
 @resource.register(name='bucket',
-                   collection_methods=('GET', 'POST', 'DELETE'),
```
leplatrem: Why this change ?

leplatrem:
@Natim, I reworked your PR here https://github.com/Kinto/kinto/commit/cbcb9bb02f8c3909866a75b1ff633b61f9208080

If you agree with the approach, I can push it here :)

kinto/views/buckets.py
Natim: Because it is back to default value.

Natim:
Yes it is much better thanks!

leplatrem:
pair=@Natim

n1k0:
o/

PR #597, opened 2016-05-19
Fix crash when a cache expires setting is set for a specific bucket or collection

leplatrem:
I'll have to backport this to 2.X :)

@Natim r?

### DIFF
--- a/kinto/views/records.py
+++ b/kinto/views/records.py
@@ -107,9 +107,8 @@ class Record(resource.ShareableResource):
 
         cache_expires = self._collection.get('cache_expires')
         if cache_expires is None:
-            by_bucket = 'kinto.%s_record_cache_expires_seconds' % (
-                self.bucket_id)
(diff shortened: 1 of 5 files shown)


kinto/views/records.py
```diff
@@ -109,11 +109,11 @@ def _handle_cache_expires(self, response):
         if cache_expires is None:
             by_bucket = 'kinto.%s_record_cache_expires_seconds' % (
                 self.bucket_id)
-            by_collection = '%s_%s_record_cache_expires_seconds' % (
+            by_collection = 'kinto.%s_%s_record_cache_expires_seconds' % (
```
Natim: Didn't we say this was stripped during conf loading?

leplatrem: Oh maybe that's the other way around, remove the prefix above

Natim: Yes it is what I though.

Natim:
r+

PR #599, opened 2016-05-20
Support old settings

glasserc:
Fixes #596.

PR #611, opened 2016-05-20
Add another setting

glasserc:
Fixes the remaining setting that was left out of #599.

After this, we can cut a minor or patch version release.


leplatrem:
r+

PR #616, opened 2016-05-21
Merge some of docs/core into docs

glasserc:
There's still a lot of stuff that needs to be deduplicated, rewritten, or rethought, but I wanted to make sure that all the code was available through some toctree, and then I went a bit overboard. Highlights in this PR include:

- merging docs/core/api into docs/api and eliminating some of the duplication
- fixing the logo in the front page and deleting the cliquet logo
- merging docs/core/troubleshooting into docs/troubleshooting
- linking docs/core/index from docs/index, so there's now a Kinto-Core toplevel section (after all the other sections)

I suspect there's still more duplication in e.g. docs/core/contributing vs. docs/community, docs/core/ecosystem vs. docs/tutorial/write-plugin. Someone also should reread docs/core/rationale and update it according to our modern worldview (which is: Kinto as long as possible, then build a service on top of kinto.core).


leplatrem:
Thanks Ethan for going further with this!

Can you explicit the changes you made to the resources docs please?

Also, I would be in favor of hiding the kinto-core docs until they are not ready. I read them quickly and I think they bring more confusion than clarity. So as you said this will have to be tackled later

docs/api/1.x/index.rst
```diff
@@ -12,18 +12,22 @@ Full detailed API documentation:
    :maxdepth: 2
 
    authentication
+   resource
```
leplatrem: I'm not sure the resource docs are ready to be exposed so much :)

docs/api/1.x/resource.rst
```diff
@@ -427,18 +200,6 @@ A ``details`` attribute in the response provides the offending record and
...
-
-When a record is created, the timestamp of the collection is incremented.
-
-It is possible to force the timestamp if the specified record has a
-``last_modified`` field.
-
-If the specified timestamp is in the past, the collection timestamp does not
-take the value of the created record but is bumped into the future as usual.
-
-
```
leplatrem: Why did you choose to remove this?

docs/api/1.x/sorting.rst
```diff
@@ -0,0 +1,19 @@
+Sorting
+#######
+
+When requesting elements (such as records) from a plural endpoint
+(such as a collection), the results can be sorted using an additional
```
leplatrem: We could repeat the formulae `Plural endpoints (such as collections) support XXX`

docs/api/1.x/pagination.rst
```diff
@@ -0,0 +1,29 @@
+Paginate
+########
+
+If the ``_limit`` parameter is provided, the number of records returned is limited.
```
leplatrem: For example: Plural endpoints (such as collections) support *limitation* of number returned objects

docs/api/1.x/index.rst
```diff
@@ -12,18 +12,22 @@ Full detailed API documentation:
    :maxdepth: 2
```
leplatrem: This may be too much, one level could help to shrink the full reference TOC
capture d ecran de 2016-05-23 10-17-10

docs/api/1.x/resource.rst
glasserc: This information is already present in the standalone `timestamps.rst` file.

docs/api/1.x/index.rst
glasserc: My hope is that we will feel embarrassed by their exposure and that will make us fix them :) Also, other people looking at them might improve them, but if they're hidden, they probably won't be improved.

glasserc:
The changes I made to `resources.rst` are all detailed in full detail in the individual commits. I extracted sections of it as `sorting`, `pagination` (which includes counting), and `selecting_fields`. I also made a minor grammatical change to an explanation of what the 304 response means on a GET.

docs/api/1.x/resource.rst
leplatrem: I am truely sorry if you had explained it already :( I usually review from the github UI and don't read the commit messages... my fault.

Regarding the change itself, shouldn't we add link like `For more info about timestamps behaviour, check out...` ? (genuine question, too much links can sometimes be confusing)

docs/api/1.x/index.rst
leplatrem: > My hope is that we will feel embarrassed by their exposure and that will make us fix them :) 

That can be a strategy indeed! ... audacious but it may work :)

leplatrem:
r+ on this. We could always go further, but this is a huge step forward! Thanks Ethan!

docs/api/1.x/resource.rst
glasserc: You're a better man than I if you can follow a giant documentation merge just by looking at one diff :)

The contents of `resource.rst` are intentionally general and try to set out principles that all endpoints follow. I think it might be better to just remove the entire file and describe the actual endpoints in buckets, collections and records files. In that case, my goal is to eventually delete this entire file, so I don't want to put a link here. But I'm not sure, and I would be open to the argument that having a generic file like this is valuable. Let's iterate on it.

PR #618, opened 2016-05-21
Update kinto.core.storage.StorageBase methods definitions.

Lothiraldan:
They were a little outdated, I copied the Postgresql storage
methods definitions for reference.


leplatrem:
Nice!
Thanks for taking the time!

Seems like you're doing something cool ;)

PR #630, opened 2016-05-23
Fix enabling flush enpoint with KINTO_FLUSH_ENDPOINT_ENABLED environment variable (fixes #588)

### LINKED ISSUE #588: Enabling the flush endpoint through env vars does not seem to work
I'm running Kinto + postgres with docker-compose (using the example docker-compose.yml in the repo). 

Adding ```KINTO_FLUSH_ENDPOINT_ENABLED``` to the environment section in docker-compose.yml does not enable the flush endpoint for me. I instead had to add ```kinto.flush_endpoint_enabled = true``` to a custom ini file, that worked.

Can the flush endpoint be enabled through an env var like this?

leplatrem:
@Natim r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ This document describes changes between each past release.
 - Fix internal storage filtering when an empty list of values is provided.
 - Authenticated users are now allowed to obtain an empty list of buckets on
   ``GET /buckets`` even if no bucket is readable (#454)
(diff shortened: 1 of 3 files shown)


Natim:
Do you think we could add a test to validate that?

leplatrem:
> Do you think we could add a test to validate that?

Although we have generic tests in kinto.core for the `load_default_settings()` mecanism, I don't think we have tests for its usage on particular kinto settings (each one defined in `kinto.__init__:DEFAULT_SETTINGS`)

Natim:
r+

Issue #628, opened 2016-05-23
PrintLogger' object has no attribute 'isEnabledFor'


leplatrem:
```
kinto start                                                          
event='Backend settings referring to cliquet are DEPRECATED. Please update your kinto.cache_backend setting to kinto.core.cache.memory (was: cliquet.cache.memory).'
event='Backend settings referring to cliquet are DEPRECATED. Please update your kinto.permission_backend setting to kinto.core.permission.memory (was: cliquet.permission.memory).'
event='Backend settings referring to cliquet are DEPRECATED. Please update your kinto.storage_backend setting to kinto.core.storage.memory (was: cliquet.storage.memory).'
Traceback (most recent call last):
  File "/home/mathieu/Code/Mozilla/kinto/.venv/bin/kinto", line 9, in <module>
    load_entry_point('kinto', 'console_scripts', 'kinto')()
  File "/home/mathieu/Code/Mozilla/kinto/kinto/__main__.py", line 97, in main
    pserve.main(pserve_argv)
  File "/home/mathieu/Code/Mozilla/kinto/.venv/local/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 60, in main
    return command.run()
  File "/home/mathieu/Code/Mozilla/kinto/.venv/local/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 371, in run
    global_conf=vars)
  File "/home/mathieu/Code/Mozilla/kinto/.venv/local/lib/python2.7/site-packages/pyramid/scripts/pserve.py", line 406, in loadapp
    return loadapp(app_spec, name=name, relative_to=relative_to, **kw)
  File "/home/mathieu/Code/Mozilla/kinto/.venv/lib/python2.7/site-packages/PasteDeploy-1.5.2-py2.7.egg/paste/deploy/loadwsgi.py", line 247, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/home/mathieu/Code/Mozilla/kinto/.venv/lib/python2.7/site-packages/PasteDeploy-1.5.2-py2.7.egg/paste/deploy/loadwsgi.py", line 272, in loadobj
    return context.create()
  File "/home/mathieu/Code/Mozilla/kinto/.venv/lib/python2.7/site-packages/PasteDeploy-1.5.2-py2.7.egg/paste/deploy/loadwsgi.py", line 710, in create
    return self.object_type.invoke(self)
[...]

PR #633, opened 2016-05-23
Clear the logger when you configure it

glasserc:
This is another ugly solution to #628 that relies on knowledge of the internals of the structlog library. The fundamental problem in https://github.com/hynek/structlog/issues/71 is that calling `configure` updates the `processors` variable globally, but the log proxy object has already set a `_logger` attribute based on a previous value of the `logger_class` parameter. If we clear that _logger variable, it will generate a new one on the next time around.

### DIFF
--- a/kinto/core/logs.py
+++ b/kinto/core/logs.py
@@ -10,19 +10,6 @@ from kinto.core import utils
 logger = structlog.get_logger()
 
 
-def reset_logger():
-    """Hack to work around https://github.com/hynek/structlog/issues/71.
-
-    This clears a magic field in the logger so that it will regenerate
-    the right kind of logger on its next use.
-
-    Do this when you call structlog.configure(). Otherwise, some parts
-    of the logger will be outdated and you may get strange behavior.
-
-    """
-    logger._logger = None
-
-
 def decode_value(value):
     try:
         return six.text_type(value)
--- a/kinto/core/logs.py
+++ b/kinto/core/logs.py
@@ -10,6 +10,19 @@ from kinto.core import utils
 logger = structlog.get_logger()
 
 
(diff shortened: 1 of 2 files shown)


tarekziade:
Why don't we use https://github.com/Kinto/kinto/issues/628#issuecomment-220983946 ?

That fixes the issue and Hynek said that's the way to go

tarekziade:
oh sorry I just saw you had a lengthy discussion on hynek's github :)) 

ignore my comment

kinto/core/logs.py
```diff
@@ -10,6 +10,19 @@
 logger = structlog.get_logger()
 
 
+def reset_logger():
+    """Hack to work around https://github.com/hynek/structlog/issues/71.
+
+    This clears a magic field in the logger so that it will regenerate
+    the right kind of logger on its next use.
+
+    Do this when you call structlog.configure(). Otherwise, some parts
```
tarekziade: Maybe we could provide our own `configure` function that includes `logger._logger = None` because I don't see a use case in kinto for using `reset_logger` besides configuring the logger.

glasserc: OK. Since we call `configure` in `initialization.py:setup_logging`, I've just moved this code there in 9f83849eb4067c1514abfc91c1afea5306aad7cc.

Natim:
r+

Natim:
Apparently https://github.com/hynek/structlog/pull/72 have been merged @glasserc I will let you merge this if you want or wait for a release of structlog.

leplatrem:
:+1: 

We can merge like this and add a dedicated issue `Remove structlog initialization hack when new version is released`

PR #634, opened 2016-05-23
[WIP] Add the preliminary code for working on mongodb data storage

Lothiraldan:
Still WIP but should be sufficient to work on with @Natim

### DIFF
--- a/kinto/core/storage/mongo/__init__.py
+++ b/kinto/core/storage/mongo/__init__.py
@@ -1,0 +1,245 @@
+from __future__ import absolute_import, unicode_literals
+from functools import wraps
+
+import pymongo
+from six.moves.urllib import parse as urlparse
+
+from kinto.core import utils, logger
+from kinto.core.storage import (
+    exceptions, DEFAULT_ID_FIELD,
+    DEFAULT_MODIFIED_FIELD, DEFAULT_DELETED_FIELD)
+from kinto.core.storage import StorageBase
+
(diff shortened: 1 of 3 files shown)


Natim:
Thank you for putting this here.

leplatrem:
Hey guyz! Cool to see you here ;)

Before you start implementing the bodies of this scaffold, can you detail a bit the value added by this new backend?

For example, I remember a while ago that AlwaysData was stuck with an old Postgresql version, and Mongo was mentioned.

Since it will add some setup/integration «*overhead*» we can list pros/cons and see if it is relevant to put it in the core or as a community-plugin!

Thanks for starting this!

Natim:
The plan is to make it so that we can deploy Kinto in AlwaysData.
The plan is also to makes it a community backend. We started working on this here to make the setup easier (plugin tests and stuff) but we will extract it at some point.

As you may have seen this is just WIP PR so that we don't loose the work that have been done so far nothing more.

tarekziade:
Very nice ! 

Maybe we could have this new backend implemented in its own repository ? e.g. kinto-mongodb. 

Keeping kinto as small as possible is making it easier to deploy when you don't use mongodb - and gives us a less complex requirements handling. For instance, no need to add a bunch of Skips in the tests.

tarekziade:
Closing this, feel free to add a link to a repo in case you start back this work. Thanks

PR #642, opened 2016-05-24
Simplify logger definition

leplatrem:
@glasserc r?

### DIFF
--- a/kinto/config/kinto.tpl
+++ b/kinto/config/kinto.tpl
@@ -123,7 +123,7 @@ kinto.includes = kinto.plugins.default_bucket
 # kinto.logging_renderer = kinto.core.logs.ClassicLogRenderer
 
 [loggers]
-keys = root
+keys = root, kinto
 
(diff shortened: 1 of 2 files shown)


kinto/config/kinto.tpl
```diff
@@ -123,7 +123,7 @@ kinto.includes = kinto.plugins.default_bucket
 # kinto.logging_renderer = kinto.core.logs.ClassicLogRenderer
 
 [loggers]
-keys = root, kinto_core
+keys = root
```
Natim: I think you should rename kinto_core by kinto and log all kinto errors here.

Natim: Here we lose the fact that we may want the DEBUG log level for kinto, while the INFO level for other package is fine.

glasserc:
I agree with @Natim :)

leplatrem:
Final approval ?

glasserc:
r+

PR #643, opened 2016-05-24
Prevent client to cache and validate authenticated requests (fixes #635)

### LINKED ISSUE #635: There's still a problem with browser caching
It seems adding the `Vary` header to responses didn't entirely solve the issue we had with cached authenticated responses in the browser.

[This fiddle](https://jsfiddle.net/n1k0/5axpdbbj/3/) reproduces the issue against the kinto dev server, where you can see by running the script at least twice that the second time the responses are cached and served to distinct authenticated requests.

The issue is reproducible in both Firefox and Chrome.

First user response for `GET /buckets`:

Second user response  for `GET /buckets`:

Notice the two distinctly used Authorization headers sent, and the the two equivalent responses obtained. In the second capture, the second user received a cached response containing data from the first user. That's rather annoying.

leplatrem:
* [x] Check that `no-store` header solves it. @n1k0 feedback ?
* [x] Revert patches for `Vary` header. @Natim feedback/opinion?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ This document describes changes between each past release.
 - Fix reading settings for events listeners from environment variables (fixes #515)
 - Fix principal added to ``write`` permission when a publicly writable object
   is created/edited (fixes #645)
+- Prevent client to cache and validate authenticated requests (fixes #635)
 
 **Documentation**
 
--- a/docs/configuration/production.rst
+++ b/docs/configuration/production.rst
@@ -160,7 +160,7 @@ It might also be relevant to set your main server :ref:`as readonly <configurati
 In the configuration of the CDN service, you should also:
 
 - Allow ``OPTIONS`` requests (CORS)
-- Pass through cache and concurrency control headers: ``ETag``, ``Last-Modified``, ``Expire``, ``Vary``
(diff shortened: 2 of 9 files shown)


Natim:
I feel completly lost with this PR,

Why would you revert patches on the Vary header? Why would you prevent the app for catching?

I think we need to describe what we are trying to achieve:

 - We do not want to serve catched value to another user.

That's what the ``Vary`` header is for. If it doesn't we need to investigate why. Preventing user from catching is a hack but not really what we are trying to do here.

Natim:
Ok then I read: https://github.com/Kinto/kinto/issues/635#issuecomment-221246471

Natim:
Did you understand why Vary: Authorization doesn't work?

Natim:
According to Google article I would rather add ``Cache-Control:  private`` and keep the Vary header.

Natim:
After discussion this is a good solution.

- The client should not handle the cache automatically. 
- The kinto protocol will continue to work (with cache) even with no-store because the client (kinto-client) will handle it.

Natim:
To phrase it differently:

By default the server explicitly tells clients to not store cache entries for authenticated responses.

If you plan on overriding this behavior, your client calls should explicitly expose the appropriate cache headers.

Natim:
r+

PR #649, opened 2016-05-24
Improved documentation about running in production with uWSGI (fixes #543, fixes #545)

### LINKED ISSUE #543: "Running in production" doesn't point to the app.wsgi file
The [running in production with uwsgi doc](http://kinto.readthedocs.org/en/latest/configuration/production.html#running-with-uwsgi) talks about an `app.wsgi` which is provided, but
1/ it's not "there" in the project folder if kinto was installed using pip
2/ it's not in the pip package I think (it's not in https://github.com/Kinto/kinto/blob/master/MANIFEST.in)

So maybe the documentation should be modified to link to the app.wsgi in the repository? (like https://github.com/Kinto/kinto/blob/master/app.wsgi)

### LINKED ISSUE #545: The documentation for running in production stores the kinto.sock in a tmpfs
Following the configuration for [running in production with uwsgi](http://kinto.readthedocs.org/en/latest/configuration/production.html#running-with-uwsgi), it advises to store the `kinto.sock` file in `/var/run/uwsgi/kinto.sock` which, at least on some distributions, is a tmpfs (temporary filesystem).

This makes uwsgi complain: `bind(): Permission denied [core/socket.c line 185]`

See also http://stackoverflow.com/questions/12778646/uwsgi-socket-permissions

leplatrem:
fixes #543, fixes #545

@glasserc r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,10 @@ This document describes changes between each past release.
   ``GET /buckets`` even if no bucket is readable (#454)
 - Fix enabling flush enpoint with ``KINTO_FLUSH_ENDPOINT_ENABLED`` environment variable (fixes #588)
 
+**Documentation**
+
+- Improved documentation about running in production with uWSGI (#543, #545)
+
 
 3.0.1 (2016-05-20)
 ==================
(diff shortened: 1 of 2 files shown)


glasserc:
r+

PR #650, opened 2016-05-24
Fix example filename in core docs (fixes #655)

### LINKED ISSUE #655: Relax record id validation
Do we really need to ensure that posted record ids match a uuid regex?

We can generate a uuid when a record without id is posted, and leave the usage of uuid in our official clients.
But is there any reason to use a different regex that collection and bucket names?

edit: The usecase is the Web sync extension chrome.storage.sync: since any key is accepted, it takes the md5 of the key to "generate" UUIDs. Instead we could let the client push any key as record id.
 
* Related #140

leplatrem:
@leplatrem r?

### DIFF
--- a/docs/core/ecosystem.rst
+++ b/docs/core/ecosystem.rst
@@ -29,7 +29,7 @@ as long as the replacement follows the original component API.
 
 .. code-block:: ini
 
-    # myproject.ini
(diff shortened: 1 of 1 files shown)


leplatrem:
r+!

PR #651, opened 2016-05-24
Added the GET /contribute.json endpoint for open-source information (fixes #607)

### LINKED ISSUE #607: [cliquet] Implement a contribute.json endpoint
Original: https://github.com/mozilla-services/cliquet/issues/334

oh! mozilla has a micro-protocol https://www.contributejson.org/

Let's implement that view

leplatrem:
I wonder where to put the API endpoint docs ? in Utilities ?

@Natim @glasserc r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,13 @@ This document describes changes between each past release.
 3.0.2 (unreleased)
 ==================
 
+**Protocol**
+
+- Added the ``GET /contribute.json`` endpoint for open-source information (fixes #607)
+
+Protocol is now in version **1.6**. See `API changelog <http://kinto.readthedocs.io/en/latest/api/>`_.
+
+
 **Bug fixes**
 
 - Fix internal storage filtering when an empty list of values is provided.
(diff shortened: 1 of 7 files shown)


CHANGELOG.rst
```diff
@@ -6,6 +6,13 @@ This document describes changes between each past release.
 3.0.2 (unreleased)
 ==================
 
+**Protocol**
+
+- Added the ``GET /contribute.json`` endpoint for open-source information (fixes #607)
+
+Protocol is now in version **1.6**. See `API changelog <http://kinto.readthedocs.io/en/latest/api/>`_.
```
glasserc: I would probably write that a protocol is "at" a version or just "is version 1.6". To me, "in version 1.6" sounds like it's contained in the version, like a feature of the version.

glasserc:
Otherwise, r+

leplatrem:
I think we should still document it somewhere in the API docs, no ?

glasserc:
Well, Kinto's documentation is very thorough, so I guess so. But I would also understand not documenting it since it isn't part of our public API.

Natim:
This is great :+1:

PR #653, opened 2016-05-25
Add permissions endpoint (fixes #600)

### LINKED ISSUE #600: Add an endpoint to list objects and permissions of current user
Context: The kinto-admin where we need to render a tree of buckets and collections the authenticated user has access to.

If you don't have read permission to the bucket containing a collection you have read access for, you can't see that bucket listed anywhere it seems. In an ideal world, I'd expect that if I have access to `/buckets/buck/collections/coll`:

- I could query `/buckets` and see `buck` listed;
- I could query `/buckets/buck/collections` and see `coll` listed.

Silly thoughts: 

- Maybe we need a new `enumerate` permission type?
- Maybe we need a new endpoint (possibly through a plugin) allowing to retrieve the tree of available resources for a given authenticated user? eg.

```
$ http http://kinto/v1/access
{
  "buckets": [
    {
      id: "buck",
      permissions: {
        write: [...], read: [...]
      },
      collections: [
        {
          id: "coll",
          permissions: {
            write: [...], read: [...]
          }
        },
        ...
      ]
    },
    ...
  ]
}
```

If that's not possible for whatever reason, how should we retrieve the list of resources an authenticated user has access to?

Note that another possible answer might be that this is just matter of configuring all this manually, by setting permissions individually per resource in the tree; though that would mean giving read permission to containers we don't actually want to fully expose to the user.

leplatrem:
I wanted to give a try to my proposition in #600.

This **can be thrown away** if the approach is wrong. It allowed me to figure out what pieces of code were necessary, especially regarding the permission backend API.

edit: ~~For example, the current code only works with `write` permissions.~~

> I renamed `test_views_permissions.py` to `test_views_objects_permissions.py` and it screwed the diff...

### Test it

```
$ KINTO_EXPERIMENTAL_PERMISSIONS_ENDPOINT=true kinto start --reload
```

```
$ echo {} | http PUT :8888/v1/buckets/default/collections/test --auth user:pass
```

```
$ http  :8888/v1/permissions --auth user:pass
```

```http
HTTP/1.1 200 OK
Content-Length: 431
Date: Wed, 25 May 2016 16:20:34 GMT
Server: waitress

{
    "data": [
        {
            "bucket_id": "2f9b1aaa-552d-48e8-1b78-371dd08688b3", 
            "collection_id": "test", 
            "permissions": [
                "write", 
                "read", 
                "record:create"
            ], 
            "resource_name": "collection", 
            "uri": "/buckets/2f9b1aaa-552d-48e8-1b78-371dd08688b3/collections/test"
        }, 
        {
            "bucket_id": "2f9b1aaa-552d-48e8-1b78-371dd08688b3", 
            "permissions": [
                "write", 
                "read", 
                "collection:create", 
                "group:create"
            ], 
            "resource_name": "bucket", 
            "uri": "/buckets/2f9b1aaa-552d-48e8-1b78-371dd08688b3"
        }
    ]
}

```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,12 @@ This document describes changes between each past release.
 3.3.0 (unreleased)
 ==================
 
-- Nothing changed yet.
+**Protocol**
+
+- Add new endpoint ``GET /v1/permissions`` to retrieve the list of permissions
+  granted on every kind of object (#600).
+
+Protocol is now at version **1.8**. See `API changelog <http://kinto.readthedocs.io/en/latest/api/>`_.
 
 
 3.2.0 (2016-06-14)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,8 +8,9 @@ This document describes changes between each past release.
 
 **Protocol**
 
-- Add new endpoint ``GET /v1/permissions`` to retrieve the list of permissions
+- Add new *experimental* endpoint ``GET /v1/permissions`` to retrieve the list of permissions
   granted on every kind of object (#600).
+  Requires setting ``kinto.experimental_permissions_endpoint`` to be set to ``true``.
 
 Protocol is now at version **1.8**. See `API changelog <http://kinto.readthedocs.io/en/latest/api/>`_.
 
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -28,7 +28,8 @@ DEFAULT_SETTINGS = {
     'bucket_create_principals': Authenticated,
     'multiauth.authorization_policy': (
         'kinto.authorization.AuthorizationPolicy'),
-    'experimental_collection_schema_validation': 'False',
+    'experimental_collection_schema_validation': False,
+    'experimental_permissions_endpoint': False,
     'http_api_version': HTTP_API_VERSION
 }
 
@@ -60,7 +61,7 @@ def main(global_config, config=None, **settings):
     # Scan Kinto views.
     kwargs = {}
 
-    flush_enabled = asbool(settings.get('flush_endpoint_enabled'))
+    flush_enabled = asbool(settings['flush_endpoint_enabled'])
     if flush_enabled:
         config.add_api_capability(
             "flush_endpoint",
@@ -69,7 +70,14 @@ def main(global_config, config=None, **settings):
             url="http://kinto.readthedocs.io/en/latest/configuration/"
                 "settings.html#activating-the-flush-endpoint")
     else:
-        kwargs['ignore'] = 'kinto.views.flush'
+        kwargs['ignore'] = ['kinto.views.flush']
+
+    # Permissions endpoint enabled if permission backend is setup.
+    permissions_endpoint_enabled = (
+        asbool(settings['experimental_permissions_endpoint']) and
+        hasattr(config.registry, 'permission'))
+    if not permissions_endpoint_enabled:
+        kwargs.setdefault('ignore', []).append('kinto.views.permissions')
 
     config.scan("kinto.views", **kwargs)
 
--- a/kinto/views/permissions.py
+++ b/kinto/views/permissions.py
@@ -1,9 +1,8 @@
 from cornice import Service
-from pyramid.request import Request
-from pyramid.interfaces import IRoutesMapper
 from pyramid.security import NO_PERMISSION_REQUIRED, Authenticated
 
 from kinto.authorization import PERMISSIONS_INHERITANCE_TREE
+from kinto.core import utils as core_utils
 
 
 permissions = Service(name='permissions',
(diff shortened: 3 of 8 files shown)


kinto/views/permissions.py
```diff
@@ -0,0 +1,64 @@
...
+        object_uris = backend.principals_accessible_objects(principals, perm)
+        for object_uri in object_uris:
+            perms_by_object_uri.setdefault(object_uri, []).append(perm)
+
+    entries = []
+    for object_uri, perms in perms_by_object_uri.items():
+        # Obtain associated resource from object URI
+        resource_name, matchdict = core_utils.view_lookup(request, object_uri)
+        # XXX: do we want bucket_id, record_id or just id ?
+        matchdict['%s_id' % resource_name] = matchdict.pop('id')
```
leplatrem: @n1k0 : 
this (consistent with notifications payload format):
```js
{
    "data": [
        {
            "bucket_id": "2f9b1aaa-552d-48e8-1b78-371dd08688b3", 
            "collection_id": "test", 
            "permissions": [
                "write", 
                "read", 
                "record:create"
            ], 
            "resource_name": "collection", 
            "uri": "/buckets/2f9b1aaa-552d-48e8-1b78-371dd08688b3/collections/test"
        }, 
        {
            "bucket_id": "2f9b1aaa-552d-48e8-1b78-371dd08688b3", 
            "permissions": [
                "write", 
                "read", 
                "collection:create", 
                "group:create"
            ], 
            "resource_name": "bucket", 
            "uri": "/buckets/2f9b1aaa-552d-48e8-1b78-371dd08688b3"
        }
    ]
}

```

or that (always an `id` attribute, whatever the resource type):
```js
{
    "data": [
        {
            "bucket_id": "2f9b1aaa-552d-48e8-1b78-371dd08688b3", 
            "id": "test", 
            "permissions": [
                "write", 
                "read", 
                "record:create"
            ], 
            "resource_name": "collection", 
            "uri": "/buckets/2f9b1aaa-552d-48e8-1b78-371dd08688b3/collections/test"
        }, 
        {
            "id": "2f9b1aaa-552d-48e8-1b78-371dd08688b3", 
            "permissions": [
                "write", 
                "read", 
                "collection:create", 
                "group:create"
            ], 
            "resource_name": "bucket", 
            "uri": "/buckets/2f9b1aaa-552d-48e8-1b78-371dd08688b3"
        }
    ]
}


```

kinto/views/permissions.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # Expand implicit permissions using descending tree.
+        permissions = set(perms)
+        for perm in perms:
+            obtained = perms_descending_tree[resource_name][perm]
+            # Related to same resource only (not every sub-objects).
+            permissions |= obtained[resource_name]
+
+        entry = dict(uri=object_uri,
+                     resource_name=resource_name,
```
leplatrem: @n1k0 `resource_name` (consistent with notifications payloads) or just `resource` ?

kinto/views/permissions.py
n1k0: I was thinking of `parent_id` but it might introduce more confusion than anything else. I'm fine with both of your suggestions with a slight preference for the second option where I appreciate being able to always rely on the `id` property describing the resource id.

kinto/views/permissions.py
n1k0: If consistency is a priority (and I think it should) I'm okay to follow existing conventions :)

leplatrem:
@Natim r?
@n1k0 feedback?

n1k0:
feedback+++

leplatrem:
> feedback+++

So far it's behing an `experimental_` setting that is false by default. I suggest that we wait for the implementation to be done in kinto-admin to enable it officially (since we may discover flaws...)

kinto/__init__.py
```diff
@@ -28,7 +28,8 @@
     'bucket_create_principals': Authenticated,
     'multiauth.authorization_policy': (
         'kinto.authorization.AuthorizationPolicy'),
-    'experimental_collection_schema_validation': 'False',
+    'experimental_collection_schema_validation': False,
```
n1k0: huh! does that mean it was always enabled?

n1k0:
>  I suggest that we wait for the implementation to be done in kinto-admin to enable it officially

wfm :+1:

CHANGELOG.rst
```diff
@@ -6,7 +6,13 @@ This document describes changes between each past release.
 3.3.0 (unreleased)
 ==================
 
-- Nothing changed yet.
+**Protocol**
+
+- Add new *experimental* endpoint ``GET /v1/permissions`` to retrieve the list of permissions
+  granted on every kind of object (#600).
+  Requires setting ``kinto.experimental_permissions_endpoint`` to be set to ``true``.
```
Natim: Shouldn't we remove experimental from config names? It doesn't add any value and we are upgrading the protocol nevertheless... I am not sure it adds any value.

Natim:
I'd like to make it return only bucket and collection by default and records if we ask for it or at least let the ability to select the kind of records we want to list.

CHANGELOG.rst
leplatrem: I like the fact that it warns the user that its enabling an experimental feature. Let's remove that when kinto-admin pull-request is ready

leplatrem:
> I'd like to make it return only bucket and collection by default and records if we ask for it or at least let the ability to select the kind of records we want to list.

I guess your concern is optimization. Let's keep filtering and optimization for another issue/PR and keep this experimental feature small and subject to breaking changes (hence the name experimental)

leplatrem:
final r+ ?

Natim:
Ok then.

leplatrem:
Would you prefer to put in a plugin while it is not used in kinto-admin (like `kinto.plugins.permissions_endpoint`) ?

/cc @n1k0

almet:
Really cool,

Having stuff as plugins seems to be a good idea, so that people can disable the stuff really easily, and a proof that our plugin system is extensible enough, so I would say yes, make it a plugin and activate it by default?

Natim:
Yes you can even do a core plugin as for default_buckets

n1k0:
First feedback: we need to enable support for the `OPTIONS` verb for the `/permissions` endpoint:

```
➜  kinto-admin git:(master) ✗ http OPTIONS :8888/v1/permissions -a toto:toto
HTTP/1.1 405 Method Not Allowed
Allow: GET, HEAD
Content-Length: 102
Date: Tue, 28 Jun 2016 09:06:33 GMT
Server: waitress

{
    "code": 405, 
    "errno": 115, 
    "error": "Method Not Allowed", 
    "message": "Method not allowed on this endpoint."
}
```

n1k0:
Successful experiment achieved in kinto-admin here https://github.com/Kinto/kinto-admin/pull/161

n1k0:
\o/

PR #656, opened 2016-05-27
Heartbeat timeout (fixes #601)

### LINKED ISSUE #601: Heartbeat max timeout
Let's make sure the heartbeat times out before 30 seconds. Otherwise uWSGI hara kiris the web head and we don't have a chance to send back the result of the heartbeat where it tells us which backend fails.

I suggest 15 seconds max by defaut + an option.

leplatrem:
* [x] POC using signals r? @tarekziade 
* [x] Write test (using very low timeout and time.sleep?)

### DIFF
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -60,6 +60,9 @@ Feature settings
 |                                                 |              | endpoint: ``/v1`` redirects to ``/v1/`` and ``/buckets/default/``        |
 |                                                 |              | to ``/buckets/default``. No redirections are made when turned off.       |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
+| kinto.heartbeat_timeout_seconds                 | ``5``        | The maximum duration of the heartbeat endpoint, which pings every        |
+|                                                 |              | backend.                                                                 |
++-------------------------------------------------+--------------+--------------------------------------------------------------------------+
 
 
 .. _configuration-backends:
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -60,9 +60,7 @@ Feature settings
 |                                                 |              | endpoint: ``/v1`` redirects to ``/v1/`` and ``/buckets/default/``        |
 |                                                 |              | to ``/buckets/default``. No redirections are made when turned off.       |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
-| kinto.heartbeat_timeout_seconds                 | ``10``       | The maximum duration of each heartbeat entry. Depending of the amount of |
-|                                                 |              | heartbeat entries (default: 3), this should be adjusted so that the      |
-|                                                 |              | maximum total remains below the WSGI worker timeout (usually 30 seconds) |
+| kinto.heartbeat_timeout_seconds                 | ``10``       | The maximum duration of each heartbeat entry, in seconds.                |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
 
 
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -60,7 +60,7 @@ Feature settings
 |                                                 |              | endpoint: ``/v1`` redirects to ``/v1/`` and ``/buckets/default/``        |
 |                                                 |              | to ``/buckets/default``. No redirections are made when turned off.       |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
-| kinto.heartbeat_timeout_seconds                 | ``2``        | The maximum duration of each heartbeat entry. Depending of the amount of |
+| kinto.heartbeat_timeout_seconds                 | ``10``       | The maximum duration of each heartbeat entry. Depending of the amount of |
 |                                                 |              | heartbeat entries (default: 3), this should be adjusted so that the      |
 |                                                 |              | maximum total remains below the WSGI worker timeout (usually 30 seconds) |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -60,8 +60,9 @@ Feature settings
 |                                                 |              | endpoint: ``/v1`` redirects to ``/v1/`` and ``/buckets/default/``        |
 |                                                 |              | to ``/buckets/default``. No redirections are made when turned off.       |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
-| kinto.heartbeat_timeout_seconds                 | ``5``        | The maximum duration of the heartbeat endpoint, which pings every        |
-|                                                 |              | backend.                                                                 |
+| kinto.heartbeat_timeout_seconds                 | ``2``        | The maximum duration of each heartbeat entry. Depending of the amount of |
+|                                                 |              | heartbeat entries (default: 3), this should be adjusted so that the      |
+|                                                 |              | maximum total remains below the WSGI worker timeout (usually 30 seconds) |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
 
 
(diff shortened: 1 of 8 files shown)


Natim:
I personalty like the approach.

kinto/core/views/heartbeat.py
```diff
@@ -10,10 +12,11 @@
 def get_heartbeat(request):
     """Return information about server health."""
     status = {}
-
     heartbeats = request.registry.heartbeats
-    for name, callable in heartbeats.items():
-        status[name] = callable(request)
+    seconds = float(request.registry.settings['heartbeat_timeout_seconds'])
+    with timeout(seconds):
```
leplatrem: nit: missing comment that it could raise TimeoutException

Natim: Can we add a test with the error message displayed to the user?

Natim: I would probably go for a global timeout as well as a timeout per heartbeat too.

leplatrem: That would be reimplementing harakiri :(

leplatrem:
> I personalty like the approach.

Too bad: 
```
$ kinto start
...
File "/home/mathieu/Code/Mozilla/kinto/.venv/local/lib/python2.7/site-packages/timeoutcontext-1.0.0-py2.7.egg/timeoutcontext/_timeout.py", line 68, in _replace_alarm_handler
    raise_timeout)
ValueError: signal only works in main thread lang=None uid=None
2016-05-27 12:05:03,265 INFO  [kinto.core.initialization][waitress] "GET   /v1/__heartbeat__" 500 (2 ms) request.summary agent=HTTPie/0.9.2 authn_type=None errno=999 lang=None time=2016-05-27T12:05:03 uid=None
```

tarekziade:
You should do it the other way 

1/ call each heartbeat function into a separate thread. 
2/ a successful result flip a flag
3/ on timeout (or before) return with the result.

use concurrent.futures.ThreadPoolExecutor

tarekziade:
also: r-

docs/configuration/settings.rst
```diff
@@ -60,6 +60,10 @@ Feature settings
 |                                                 |              | endpoint: ``/v1`` redirects to ``/v1/`` and ``/buckets/default/``        |
 |                                                 |              | to ``/buckets/default``. No redirections are made when turned off.       |
 +-------------------------------------------------+--------------+--------------------------------------------------------------------------+
+| kinto.heartbeat_timeout_seconds                 | ``2``        | The maximum duration of each heartbeat entry. Depending of the amount of |
```
tarekziade: 2 seconds seems very low. Since we're running in parallel I think we can use 10 seconds by default maybe ?

kinto/core/__init__.py
```diff
@@ -54,6 +54,7 @@
         'kinto.core.events.setup_transaction_hook',
     ),
     'event_listeners': '',
+    'heartbeat_timeout_seconds': 5,
```
tarekziade: I am confused, I've read 2, 3 and 5 seconds in various places

kinto/core/views/heartbeat.py
```diff
@@ -1,20 +1,37 @@
...
 @heartbeat.get(permission=NO_PERMISSION_REQUIRED)
 def get_heartbeat(request):
     """Return information about server health."""
-    status = {}
 
+    # Start executing heartbeats concurrently.
     heartbeats = request.registry.heartbeats
+    pool = ThreadPoolExecutor(max_workers=len(heartbeats.keys()))
+    futures = []
     for name, callable in heartbeats.items():
```
tarekziade: shadowing the builtin, let's use function or func or callable_

kinto/core/views/heartbeat.py
```diff
@@ -1,20 +1,37 @@
...
-        status[name] = callable(request)
+        futures.append((name, pool.submit(callable, request)))
 
+    # Wait the results, with timeout.
+    seconds = float(request.registry.settings['heartbeat_timeout_seconds'])
+    status = {}
+    for name, future in futures:
+        try:
+            status[name] = future.result(timeout=seconds)
```
tarekziade: you are running threads, but here you are doing a sequential loop. We're losing the benefits of the threads and making the loop duration dependant on the number of callables.

The pattern is to just return avec n seconds and let the futures flip the flag. e.g. a general wait with a timeout (see the pool API)

leplatrem: Hum, I don't think so. The callable starts running when it is submitted to the pool, that's why I have two loops!

leplatrem: Ok, sorry, I think I know what you mean. Will rework.

kinto/core/views/heartbeat.py
```diff
@@ -11,10 +14,28 @@ def get_heartbeat(request):
...
+
+    # Wait the results, with timeout.
+    seconds = float(request.registry.settings['heartbeat_timeout_seconds'])
+    done, not_done = wait(futures, timeout=seconds)
+    if len(not_done) > 0:
+        name = not_done.pop().__heartbeat_name
+        error_msg = "'%s' heartbeat has exceeded timeout of %s seconds."
+        logger.error(error_msg % (name, seconds))
+
+    # If any has failed, return a 503 error response.
```
tarekziade: there's one thing missing: when one (or several) heartbeat(s) fails, we need to catch the future(s) exception(s) and log them here with `logger.exception` so we can track down what happens. The future object holds that exception, so we just need to iterate on them and collect the TB

tarekziade: another option is to catch them in `heartbeat_check` and push them in an `exceptions` list

tarekziade: also: make sure you call `logger.exception` in the main thread sequentially on the collected tracebacks, otherwise you might have mangled exceptions since two functions can fail in parallel at the same instant

kinto/core/views/heartbeat.py
```diff
@@ -11,10 +14,35 @@ def get_heartbeat(request):
...
+        future.__heartbeat_name = name  # For logging purposes.
+        futures.append(future)
+
+    # Wait for the results, with timeout.
+    seconds = float(request.registry.settings['heartbeat_timeout_seconds'])
+    done, not_done = wait(futures, timeout=seconds)
+
+    # A heartbeat is supposed to return True or False, and never raise.
+    # Just in case, go though results to spot any potential exception.
+    for future in done:
```
tarekziade: if we raise here we're getting a 500 I think. I think what you want is (not tested):

```python
for future in done:
    exc = future.exception()
    if exc is not None:
        logger.error("<xxx> failed")
        logger.error(exc)
```

leplatrem: Yes, I did not change the previous behaviour regarding heartbeats (the `try/except` is managed there)

tarekziade: ok I see, so there's the convention that the heartbeat functions should catch all errors. But we never know what might happen in external heartbeat functions.

Is that a behaviour we want to keep ? 

e.g. do we want the global heartbeat to completely fail when one heartbeat is producing an error, or do we want to lg that error and flag that backend to false in the result ?

kinto/core/views/heartbeat.py
```diff
@@ -34,7 +34,10 @@ def heartbeat_check(name, func):
     # A heartbeat is supposed to return True or False, and never raise.
     # Just in case, go though results to spot any potential exception.
     for future in done:
-        future.result()  # Will re-raise.
+        exc = future.exception()
+        if exc is not None:
+            logger.error("'%s' heartbeat failed." % future.__heartbeat_name)
```
tarekziade: nit: user %r instead of '%s'

tarekziade:
Looks great r+

kinto/core/views/heartbeat.py
leplatrem: aaah that's why you used `%r`! :/

PR #659, opened 2016-05-27
Renamed some permission backend methods for consistency with other classes (fixes #608)

### LINKED ISSUE #608: [cliquet] Uniformize the naming of getters in our classes
original: https://github.com/mozilla-services/cliquet/issues/290

Looks like the coding style is not uniform on our classes. That bug proposes to use `get_`. People agreed. 

So the work left to be done is a pass on the code and see if we use `get_` everywhere

leplatrem:
@Natim r?


Natim:
LGTM

PR #660, opened 2016-05-30
Fix 400 with data.id in default bucket

leplatrem:
It can look hacky, but it remains confined in the default plugin, and prevents clients to have code like `if id == "default"`

@Natim r?

### DIFF
--- a/kinto/core/utils.py
+++ b/kinto/core/utils.py
@@ -254,8 +254,12 @@ def build_request(original, dict_obj):
     path = path.encode('utf-8')
 
     method = dict_obj.get('method') or 'GET'
+
     headers = dict(original.headers)
     headers.update(**dict_obj.get('headers') or {})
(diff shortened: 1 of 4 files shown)


kinto/core/utils.py
```diff
@@ -254,8 +254,11 @@ def build_request(original, dict_obj):
     path = path.encode('utf-8')
 
     method = dict_obj.get('method') or 'GET'
+
     headers = dict(original.headers)
     headers.update(**dict_obj.get('headers') or {})
+    headers.pop('Content-Length', None)
```
Natim: Cannot we remote it just for the default bucket?

leplatrem: You're talking about `id`? You'd prefer to remove it if equal to `default`?

Natim:
r+

PR #667, opened 2016-06-01
Allow record ids to be any string instead of just UUIDs (fixes #655)

### LINKED ISSUE #655: Relax record id validation

leplatrem:
@glasserc r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ This document describes changes between each past release.
 3.1.1 (unreleased)
 ==================
 
+**Protocol**
+
+- Allow record ids to be any string instead of just UUIDs (fixes #655).
+
+Protocol is now at version **1.7**. See `API changelog <http://kinto.readthedocs.io/en/latest/api/>`_.
+
 **New features**
 
 - ``kinto start`` now accepts a ``--port`` option to specify which port to listen to.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 **Protocol**
 
-- Allow record ids to be any string instead of just UUIDs (fixes #655).
+- Allow record IDs to be any string instead of just UUIDs (fixes #655).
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.rst
```diff
@@ -6,6 +6,12 @@ This document describes changes between each past release.
 3.1.1 (unreleased)
 ==================
 
+**Protocol**
+
+- Allow record ids to be any string instead of just UUIDs (fixes #655).
```
Natim: nit: ``IDs `

docs/api/index.rst
```diff
@@ -11,7 +11,12 @@ API
 Changelog
 ---------
 
-1.6 (unreleased)
+1.7 (unreleased)
+''''''''''''''''
+
+- Allow record ids to be any string instead of just UUID.
```
Natim: nit: `IDs`

kinto/core/storage/generators.py
```diff
@@ -51,9 +51,5 @@ class UUID4(Generator):
     duplicate would be about 50% (`source <http://en.wikipedia.org/wiki/\
 Universally_unique_identifier#Random_UUID_probability_of_duplicates>`_).
     """
-    regexp = (r'^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-'
```
Natim: I don't really like that change :(

Shouldn't we actually inherit from the kinto-core UUID4 generator in kinto to override the regexp with a more general one?

Natim:
r- about removing the regexp from the Generator

leplatrem:
Updated. I also added some tests about posting with existing `id`. I was sure we had some!

Natim:
Oh yeah :)

glasserc:
LGTM

PR #670, opened 2016-06-02
Mention that group principals are URIs

leplatrem:
@n1k0 r?


n1k0:
Indeed that's nice to know ;) r+

PR #678, opened 2016-06-09
Faq: add a bunch of questions

enguerran:
r?

Feel free to rephrase. I did not invent neither the question nor the answers, I take them from slack discussion. /see https://kinto.slack.com/files/ticabri/F1FDXLU58/FAQ_Kinto

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -173,3 +173,17 @@ There are three conflict resolution strategies:
 
 There is, of course, a convenient API to handle conflict one by one
 http://kintojs.readthedocs.io/en/latest/api/#resolving-conflicts-manually
+
+Would you recommend Redis or PostgreSQL?
+--------------------------------------
+
+You can use both of them:
+
+* Redis will let you start easily and you will have a database running in memory which
+means your database should be smaller than your server RAM. It is a good solution for
+experimentation and you will also be able to use a Redis cluster to scale in production.
+* PostgreSQL is a good solution for a Kinto server and will let you use all the power of
+PostgreSQL and its tooling.
+
+Do not hesitate to mix both if you can, for instance you can use PostgreSQL for the
+storage backend and Redis for the permission and cache backends.
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -195,7 +195,7 @@ No, and it will not. This is something that should be done on top of Kinto, with
 ElasticSearch for instance. In order to do this, you could listen to the events that
 Kinto triggers and send the data to your ElasticSearch cluster.
 `There is a tutorial <http://kinto.readthedocs.io/en/latest/tutorials/write-plugin.html>`_
-for that on the documentation.
+for that in the documentation.
 
 Say I wanted to move all my Kinto data out of the database, would the best way to be via the backend?
 -----------------------------------------------------------------------------------------------------
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -161,3 +161,15 @@ I am seeing an Exception error, what's wrong?
 
 Have a look at the :ref:`Troubleshooting section <troubleshooting>` to
 see what to do.
+
+If two users modify the same collection offline, how does that conflict get resolved?
+-------------------------------------------------------------------------------------
+
+There are three conflict resolution strategies:
+
+* SERVER_WINS: local changes are overridden by remote ones ;
+* CLIENT_WINS: remote changes are overriden by local one ;
+* MANUAL (default): handle them on your own.
+
+There is, of course, a convenient API to handle conflict one by one
+http://kintojs.readthedocs.io/en/latest/api/#resolving-conflicts-manually
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -187,3 +187,12 @@ PostgreSQL and its tooling.
 
 Do not hesitate to mix both if you can, for instance you can use PostgreSQL for the
 storage backend and Redis for the permission and cache backends.
+
+What about aggregation/reporting around data, is Kinto ready for that?
+----------------------------------------------------------------------
+
+No, and it will not. This is something that should be done on top of Kinto, with
+ElasticSearch for instance. In order to do this, you could listen to the events that
+Kinto triggers and send the data to your ElasticSearch cluster.
+`There is a tutorial <http://kinto.readthedocs.io/en/latest/tutorials/write-plugin.html>`_
+for that on the documentation.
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -161,3 +161,40 @@ I am seeing an Exception error, what's wrong?
 
 Have a look at the :ref:`Troubleshooting section <troubleshooting>` to
 see what to do.
+
+If two users modify the same collection offline, how does that conflict get resolved?
+-------------------------------------------------------------------------------------
+
+It depends on which client you use. If you use kinto.js, for instance, you can set up
```
Natim: We shouldn't talk about client here, all client can and should implement the three thing. You can even implement it at the HTTP level.

We should rephrase like that:

    There are three conflict resolution strategies:

docs/faq.rst
```diff
@@ -161,3 +161,40 @@ I am seeing an Exception error, what's wrong?
...
+three conflict resolution strategies:
+
+* SERVER_WINS: local changes are overridden by remote ones ;
+* CLIENT_WINS: remote changes are overriden by local one ;
+* MANUAL (default): handle them on your own.
+
+There is, of course, a convenient API to handle conflict one by one
+http://kintojs.readthedocs.io/en/latest/api/#resolving-conflicts-manually
+
+Would you recommend redis or postgres?
```
Natim: Redis, PostgreSQL

docs/faq.rst
```diff
@@ -161,3 +161,40 @@ I am seeing an Exception error, what's wrong?
...
+--------------------------------------
+
+We recommend using postgres which is most focused lately, but you can still use redis.
```
Natim: We can rephrase like:

You can use both of them: 

- Redis will let you start easily and you will have a database running in memory which means your database should be smaller than your server RAM. It is a good solution for experimentation and you will also be able to use a Redis cluster to scale in production.

- PostgreSQL is a good solution for a Kinto server and will let you use all the power of PostgreSQL and its tooling.

Do not hesitate to mix both if you can, for instance you can use PostgreSQL for the storage backend and Redis for the permission and cache backends.

docs/faq.rst
```diff
@@ -161,3 +161,40 @@ I am seeing an Exception error, what's wrong?
...
+The app we've extracted Kinto from
```
Natim: We should remove that it is not true.

docs/faq.rst
```diff
@@ -161,3 +161,40 @@ I am seeing an Exception error, what's wrong?
...
+`was using redis <https://github.com/mozilla-services/readinglist/blob/4745ff1ae31b2e4dc22487fe6d96cc97a083c7fc/config/readinglist.ini#L4>`_.
+
+What about aggregation/reporting around data, is Kinto ready for that?
+----------------------------------------------------------------------
+
+No, and it will not. This is something that should be done on top of Kinto, with
+ElasticSearch for instance. In order to do this, you could listen to the events that
+Kinto triggers and send the data to your ElasticSearch cluster. There is a tutorial for
```
Natim: We can add a link to the tutorial: http://kinto.readthedocs.io/en/latest/tutorials/write-plugin.html

docs/faq.rst
```diff
@@ -161,3 +161,40 @@ I am seeing an Exception error, what's wrong?
...
+that on the documentation.
+
+Say I wanted to move all my Kinto data out of the database, would the best way to be via the backend?
+-----------------------------------------------------------------------------------------------------
+
+It really depends on how you setup things, and what kind of data is there. One really
```
Natim: I would say that the Kinto protocol is build in order for you to sync things.

Therefore you can use the protocol to sync two databases.

Natim:
Thank you @enguerran for starting this, could you give me the write access on your fork so that I can a commit with my changes? You can also take them into account directly.

enguerran:
@Natim I am not familiar with PR, I added you as contributor of my fork. Nevertheless I can modify my commits regarding your remarks: may I have to keep one commit by question/answer or may I let modification visible in the PR by creating one new commit for each modification you ask?

Natim:
I think one commit will do :)

enguerran:
I updated questions/answers with a beautiful `push -f` to keep one commit by question.
But I see you speak about _one commit_: do you mean _one and only one commit_ for the PR?

Natim:
No I was thinking that you could just add a commit with my changes and a flag ``@natim review`` that is [usually what we do](https://github.com/mozilla-services/servicedenuages.fr/pull/100/commits). I am not a big fan of rewritting history but your solution is totally fine :)

docs/faq.rst
```diff
@@ -161,3 +161,50 @@ I am seeing an Exception error, what's wrong?
 
 Have a look at the :ref:`Troubleshooting section <troubleshooting>` to
 see what to do.
+
+If two users modify the same collection offline, how does that conflict get resolved?
```
Natim: modifies

docs/faq.rst
```diff
@@ -161,3 +161,50 @@ I am seeing an Exception error, what's wrong?
...
+-------------------------------------------------------------------------------------
+
+There are three conflict resolution strategies:
```
Natim: conflicts

docs/faq.rst
Natim: Actually I am not 100% sure about this one either :)

docs/faq.rst
```diff
@@ -161,3 +161,50 @@ I am seeing an Exception error, what's wrong?
...
+storage backend and Redis for the permission and cache backends.
+
+What about aggregation/reporting around data, is Kinto ready for that?
+----------------------------------------------------------------------
+
+No, and it will not. This is something that should be done on top of Kinto, with
+ElasticSearch for instance. In order to do this, you could listen to the events that
+Kinto triggers and send the data to your ElasticSearch cluster.
+`There is a tutorial <http://kinto.readthedocs.io/en/latest/tutorials/write-plugin.html>`_
+for that on the documentation.
```
Natim: in the documentation

enguerran:
Alright, I'll do one commit by review for the next one. Thanks a lot for your patience and your time.
My last correction is coming.

docs/faq.rst
enguerran: I think it is like _if they modify_

Natim:
Ok thanks, I will merge it and we can iterate later :+1:

docs/faq.rst
glasserc: Yes, @enguerran is correct, "one user modifies", "two users modify"

PR #681, opened 2016-06-14
Align text in logo

leplatrem:
I did as little change as possible, based on @elioqoshi feedback. 
We should probably adjust text size but I didn't want to do it without the vectorial sources...


leplatrem:
not unanimous feedback on this. Let's wait for Aymeric feedback :)

PR #684, opened 2016-06-16
Do not forget to update the API release date

leplatrem:
@Natim r?

I released 3.2 without setting the release date of protocol 1.7 :(((


Natim:
r+

Issue #638, opened 2016-05-24
add a --dry-run mode in migrate


tarekziade:
original: https://github.com/mozilla-services/cliquet/issues/685

a --dry-run mode will let us see what's the current schema and what will be the operations to perform

leplatrem:
Fixed in #685

PR #685, opened 2016-06-20
--dry-run mode added in migrate

lavish205:
#638

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -2,7 +2,8 @@ from __future__ import print_function
 import argparse
 import os
 import sys
-
+import logging
+import logging.config
 from six.moves import input
 from kinto.core import scripts
 from pyramid.scripts import pserve
@@ -44,8 +45,14 @@ def main(args=None):
     parser_init.set_defaults(which='init')
 
     parser_migrate = subparsers.add_parser('migrate')
+    parser_migrate.add_argument('--dry-run',
+                                action='store_true',
+                                help='Simulate the migration operations '
+                                     'and show information',
+                                dest='dry',
+                                required=False,
+                                default=False)
     parser_migrate.set_defaults(which='migrate')
-
     parser_start = subparsers.add_parser('start')
     parser_start.add_argument('--reload',
                               action='store_true',
@@ -92,8 +99,10 @@ def main(args=None):
                 pip.main(['install', "kinto[postgresql]"])
 
     elif parsed_args['which'] == 'migrate':
+        dry = parsed_args['dry']
+        logging.config.fileConfig(config_file)
         env = bootstrap(config_file)
-        scripts.migrate(env)
+        scripts.migrate(env, dry)
 
     elif parsed_args['which'] == 'start':
         pserve_argv = ['pserve', config_file]
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -74,34 +74,41 @@ class Storage(StorageBase):
         self._max_fetch_size = max_fetch_size
 
     def _execute_sql_file(self, filepath):
-        here = os.path.abspath(os.path.dirname(__file__))
-        schema = open(os.path.join(here, filepath)).read()
+        schema = open(filepath).read()
         # Since called outside request, force commit.
         with self.client.connect(force_commit=True) as conn:
             conn.execute(schema)
 
-    def initialize_schema(self):
+    def initialize_schema(self, is_dry=False):
         """Create PostgreSQL tables, and run necessary schema migrations.
 
         .. note::
 
             Relies on JSONB fields, available in recent versions of PostgreSQL.
         """
+        here = os.path.abspath(os.path.dirname(__file__))
+
         version = self._get_installed_version()
         if not version:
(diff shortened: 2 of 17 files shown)


Natim:
This seems to be a great start indeed.

kinto/__main__.py
```diff
@@ -44,8 +45,14 @@ def main(args=None):
     parser_init.set_defaults(which='init')
 
     parser_migrate = subparsers.add_parser('migrate')
+    parser_migrate.add_argument('--dry-run',
+                                action='store_true',
+                                help='Information regarding version'
+                                     ' of the schema in user',
```
leplatrem: Help would be better like `Simulate the migration operations and show information`

lavish205:
@leplatrem Do i need to modify test cases as well?

kinto/core/storage/postgresql/__init__.py
```diff
@@ -111,11 +118,17 @@ def initialize_schema(self):
...
-            self._execute_sql_file(os.path.join('migrations', filepath))
-
-        logger.info('Schema migration done.')
+            logger.info('Migrate PostgreSQL storage schema from'
+                        ' version %s to %s.' % migration)
+            filename = 'migration_%03d_%03d.sql' % migration
+            filepath = os.path.join(here, 'migrations', filename)
+            if is_dry:
+                logger.info("Execute PostgreSQL storage migration"
+                            " from " % filepath)
```
leplatrem: missing `%s`

kinto/core/storage/postgresql/__init__.py
```diff
@@ -111,11 +118,17 @@ def initialize_schema(self):
...
+            else:
+                self._execute_sql_file(filepath)
+        logger.info('PostgreSQL storage schema migration ' +
+                    ("simulated." if is_dry else "done."))
```
leplatrem: nitpick: choose between single quote or double quote but not both mixed

leplatrem:
> Do i need to modify test cases as well?

I guess the tests will fail if the [code coverage](https://en.wikipedia.org/wiki/Code_coverage) won't be 100%.
You'll have to at least add a new test that calls the migration methods with `is_dry=True`

leplatrem:
>  <lavish> leplatrem, what to do next for that issue? (--dry-run one)

First, the tests should pass. 

Then, you need to add a couple of tests where you call initialize_schema with `is_dry=True`.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -74,34 +74,41 @@ def __init__(self, client, max_fetch_size, *args, **kwargs):
...
             Relies on JSONB fields, available in recent versions of PostgreSQL.
         """
+        here = os.path.abspath(os.path.dirname(__file__))
+
         version = self._get_installed_version()
         if not version:
+            filepath = os.path.join(here, 'schema.sql')
+            if is_dry:
+                logger.info("Create PostgreSQL storage schema at version "
+                            "%s from %s" % (self.schema_version, filepath))
```
leplatrem: I believe that this message should not be within `is_dry` but always in output instead

leplatrem:
@lavish205 Are you stuck ? Do you wish we'd take over ? Don't hesitate to tell us :)

Natim:
Yes apparently tests are passing it is just a matter of 2 lines missing in the coverage.

lavish205:
@leplatrem @Natim I'm unable to figure out what is missing?

Natim:
According to the test coverage results:

missing-lines

These are the lines that are not tested:

We should add a test to try the cache migration in ``--dry-mode``
- https://github.com/lavish205/kinto/blob/issue_638/kinto/core/cache/postgresql/__init__.py#L80-L82

Same thing for the permission backend:
- https://github.com/lavish205/kinto/blob/issue_638/kinto/core/permission/postgresql/__init__.py#L74-L76

Same thing for the storage backend:
- https://github.com/lavish205/kinto/blob/issue_638/kinto/core/storage/postgresql/__init__.py#L94-L96
- https://github.com/lavish205/kinto/blob/issue_638/kinto/core/storage/postgresql/__init__.py#L125-L127

lavish205:
@Natim I have added one test case: https://github.com/lavish205/kinto/blob/49b6a50a28bf2a47976e169e21ad3ada8aa83b46/kinto/tests/core/test_scripts.py#L27
Where I'm calling storage, cache and permission migration. What else I have to do?

Natim:
Your test doesn't cover the lines because you are mocking initialize_schema.
The missing tests are to actually call the initialize_schema with dry_run.

leplatrem:
As I said above, you should leverage the code you've changed, using the tests cases (the coverage results give you the lines that were not executed).

Some lines will be very easy to reach, some others will not be trivial because you'll have to go through each step of the code to go through every «`if`».

We can help for the hardest ones, but I'm sure you can do some of them :)
But if you feel discouraged, just tell us, most of the work is done already, it would be awesome to have this feature in version 3.4 ;)

Good luck!

leplatrem:
I merged it and added the couple of necessary tests :)

Thanks Lavish for your work on this! This is a really useful feature for production stacks.

lavish205:
Thanks @leplatrem  and @Natim

PR #686, opened 2016-06-20
Pin SQLAlchemy to <1.1

leplatrem:
Builds fail because the transaction adapter for SQLAlchemy is not compatible with SQLAlchemy 1.1 (currently beta)

See https://github.com/zopefoundation/zope.sqlalchemy/issues/15

@Natim r?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ if installed_with_pypy:
     # We install psycopg2cffi instead of psycopg2 when dealing with pypy
     # Note: JSONB support landed after psycopg2cffi 2.7.0
     POSTGRESQL_REQUIRES = [
-        'SQLAlchemy<1.1',  # zopefoundation/zope.sqlalchemy#15
+        'SQLAlchemy==1.0',  # zopefoundation/zope.sqlalchemy#15
         'psycopg2cffi>2.7.0',
         'zope.sqlalchemy',
     ]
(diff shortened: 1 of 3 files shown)


leplatrem:
@Natim do you have an idea why it installs `SQLAlchemy==1.1.0b1` with `<1.1` in setup?

almet:
Resources that might be useful https://www.python.org/dev/peps/pep-0440/#handling-of-pre-releases and https://www.python.org/dev/peps/pep-0440/#exclusive-ordered-comparison.

setup.py
```diff
@@ -37,15 +37,15 @@ def read_file(filename):
     # We install psycopg2cffi instead of psycopg2 when dealing with pypy
     # Note: JSONB support landed after psycopg2cffi 2.7.0
     POSTGRESQL_REQUIRES = [
-        'SQLAlchemy',
+        'SQLAlchemy==1.0',  # zopefoundation/zope.sqlalchemy#15
```
Natim: I am not sure this would take 1.0.13. Also the requirements.txt file seems to say the contrary so if you built the requirements.txt file after this change just keep it like that. I would probably change it with ``<=1.1.0`` since ``1.0.13`` might be the last but we are not sure about it.

Natim:
Totally r+ thanks for fixing that :)

PR #690, opened 2016-06-28
Convert readthedocs links to HTTPS

adamchainz:
Because security.

Also I was just running my `.org` -> `.io` converter but then realized you've already done that bit, just not the HTTPS.


Natim:
r+

Issue #230, opened 2015-10-23
History of changes


tarekziade:
**issue pending a use case feedback**

Let's add a history of changes in Kinto.

Slightly different from the requests logged into heka, we're only logging here changes made to the database.

Features:
- log each change made to collections : who/when/what (hey, we have notifications for this now ;))
- provide a simple API to retrieve the history  
- maximum history size in days through configuration

leplatrem:
That's a great use case for writing a kinto plugin indeed :)

ametaireau:
Notifications for the win!

leplatrem:
The implementation of history tracking itself is very simple, it is very similar to what was done in kinto-changes (BTW I think this feature should land there).
We could even store a diff between old and new on update etc. 

Among the use-cases we have in mind, there is:

* What was changed in this collection since last time?
* When was this record created?
* What is the history of the collection json schema?


History must thus track and allow filtering/sorting by:

* `resource_name`: every kind of object (bucket, collection, records, ...)
* `prefixed_userid`: userid prefixed by authentication policy
* `action`: create/update/delete
* `uri`: object URI without version prefix

The history is actually a resource whose records are the notifications payloads, that can be filtered, sorted etc. like any other resource but readonly.

But while I was thinking about the implementation, I realized the main challenge was about permissions (especially if we store diffs).

* **Minimalist**: use a setting to specify which principals are allowed to read
* **Per bucket/collection**: readable if read on bucket/collection
* **Filtered per permission**: show only the history entries where the related object is readable. This would probably give the best experience, but would be relatively complex to implement ('cuz of perfs, reminder: permissions can be in redis, history records in postgresql)

Do you have any feedback on the preferred approach? I personnally like the second one as a first step.

n1k0:
> The history is actually a resource whose records are the notifications payloads

Just to be sure, would that mean we could access the resource data itself? Ideally it would be nice to have the previous data (before the action is executed) and the new data (after the action is executed). Maybe this could be rendered conditionally (eg. with `?full=true`) to avoid large entries to bloat the output.

Natim:
About the permission layer I would create the history collection in the same bucket so that we can give permissions to the same groups of the bucket. However we shouldn't automatically detect permissions and just let them be defined in the config as for kinto-changes.

leplatrem:
> Ideally it would be nice to have the previous data and the new data

Yes, this is trivial to implement.

[...]

PR #694, opened 2016-07-04
History of changes plugin (ref #230)

leplatrem:
I wanted to try out the usage of a `kinto.core` resource for this... 

I had to add a couple of safety checks (poc) when a resource is registered without its instance endpoint. 

* [x] Declare new resource `history`
* [x] Register without `record_path`
* [x] History of resources
* [x] History of default bucket (#704)
* [x] Passing tests of permissions behaviour
* [x] Fix bug in memory permission backend (#708)
* [x] Fix behaviour in permission backends when bound permissions is empty (#709)
* [x] Implement efficient permissions on history entries (new methods in permission backend) (#714)
* [x] Implement efficient bucket deletion (#711)
* [x] Use instance_uri()
* [ ] Validate the payloads (maybe implement POC in kinto-admin)
* [x] Write documentation


Natim:
f+

leplatrem:
#704 has become a blocker for this.

leplatrem:
Okay my friends! The code is ready! You can review it @Natim !
Meanwhile I'll update the docs.

kinto/plugins/history/__init__.py
```diff
@@ -0,0 +1,17 @@
+from kinto.core.events import ResourceChanged
+
+from .listener import on_resource_changed
+
+
+def includeme(config):
+    config.add_api_capability('history',
```
Natim: Can we list here the buckets for which the history is activated?

kinto/plugins/history/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+                              description='Track changes on data.',
+                              url='https://kinto.readthedocs.io')
+
+    # Activate end-points.
+    config.scan('kinto.plugins.history.views')
+
+    # Listen to every resources (except history)
```
Natim: I wish we could select in the config the bucket that we want the history to be kept for as we did for kinto-changes.

Natim: btw this would be automatic if we used a Listener rather than a subscriber.

kinto/plugins/history/listener.py
```diff
@@ -0,0 +1,94 @@
+from kinto.core.utils import instance_uri
+import copy
+from datetime import datetime
+
+
+def on_resource_changed(event):
```
Natim: Could we use a Listener there? https://github.com/Kinto/kinto-changes/blob/master/kinto_changes/listener.py#L11

kinto/plugins/history/__init__.py
leplatrem: The history is activated for every buckets. Spawn an issue to add a setting in order to limit the buckets where it's enabled ?

kinto/plugins/history/listener.py
leplatrem: Interesting, I'll look into that.

kinto/plugins/history/__init__.py
Natim: No, it should be part of the MVP, if we use a Listener this will be already covered.

kinto/plugins/history/__init__.py
leplatrem: Actually listeners can setup for certain kinds of objects, but not for a list of buckets/collections ids. Let's keep that for another issue. Because we might prefer have a blacklist instead etc.
From the point of view of kinto-admin, it makes it harder to implement if the tracked buckets are limited. Let's start simple.

kinto/plugins/history/listener.py
leplatrem: Well actually, I've look and it makes the configuration more verbose for users (plugin+listener), and does not bring anything concrete... So I might want to leave it like that.

Natim: Ok let's create an issue to address that then :'(

leplatrem:
Final r+ ?

Natim:
r+

docs/api/1.x/history.rst
```diff
@@ -0,0 +1,123 @@
...
+                        }
+                    },
+                    "timestamp": 1469006098757,
+                    "uri": "/buckets/blog/collections/articles/records/b3b76c56-b6df-4195-8189-d79da4a128e1",
+                    "user_id": "basicauth:43181ac0ae7581a23288c25a98786ef9db86433c62a04fd6071d11653ee69089",
+                }
+            ]
+        }
+
+As other plural endpoints, the entries can be filtered and sorted using the querystring.
```
Natim: other plural endpoints → other endpoints ?

kinto/plugins/history/__init__.py
Natim: Apparently it is not covered by the use of Listener.

Natim:
Great job \o/

PR #696, opened 2016-07-04
Fix crash unknown collection and readonly (fixes #693)

### LINKED ISSUE #693: Inconsistency with 404 response on empty collections
* Set `read_only` to true
* Give readonly access to the postgresql user
* Give `read` permission to everyone on a bucket `foo`
* Going to `/buckets/foo/collections/unknown` gives 404
* Going to `/buckets/foo/collections/unknown/records` gives 503

Listing the records of an unknown collection should definitely give 404, except with the `default` bucket plugin.

Very related to https://github.com/Kinto/kinto/issues/558

leplatrem:
@Natim r?


Natim:
GG.

PR #520, opened 2016-03-29
[WIP] Using route_path to generate url (fixes #103)

vsham20:
Refs #103

PR #702, opened 2016-07-05
Using route_path to generate parent_id. — Fixes #103

### LINKED ISSUE #103: pyramid.route_path to build internal route_path
We are building URIs by hand at multiple places in the code (e.g. parent id, default subrequests) and we should probably use ``request.route_path()`` for that.

http://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_path

Natim:
Revamp of #520

### DIFF
--- a/kinto/core/utils.py
+++ b/kinto/core/utils.py
@@ -383,3 +383,9 @@ def view_lookup(request, uri):
     resource_name = route.name.replace('-record', '')\
                               .replace('-collection', '')
     return resource_name, matchdict
+
+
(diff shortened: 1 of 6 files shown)


leplatrem:
Instead of duplicating the imports and the call to strip prefix, it would be a lot nicer to have something like:

```py
from kinto.core.utils import instance_uri, list_uri

instance_uri(request, 'collection', id='toto')
list_uri(request, 'record')
```

leplatrem:
r+

PR #711, opened 2016-07-07
Delete records from storage by parent_id

leplatrem:
* [x] Delete by parent_id in postgresql
* [x] Delete by parent_id in memory
* [x] Delete by parent_id in redis
* [x] Purge deleted by parent_id in postgresql
* [x] Purge deleted by parent_id in memory
* [x] Purge deleted by parent_id in redis
* [x] Use deletion and purge by parent_id for bucket deletion

### DIFF
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -40,7 +40,7 @@ class MemoryBasedStorage(StorageBase):
                    for r in records]
         return deleted
 
-    def strip_deleted_record(self, resource, parent_id, record,
+    def strip_deleted_record(self, collection_id, parent_id, record,
                              id_field=DEFAULT_ID_FIELD,
                              modified_field=DEFAULT_MODIFIED_FIELD,
                              deleted_field=DEFAULT_DELETED_FIELD):
@@ -50,7 +50,6 @@ class MemoryBasedStorage(StorageBase):
         deleted = {}
         deleted[id_field] = record[id_field]
         deleted[modified_field] = record[modified_field]
-
         deleted[deleted_field] = True
         return deleted
 
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -320,16 +320,16 @@ class Storage(MemoryBasedStorage):
                 collections = {collection_id: collections[collection_id]}
             for collection, colrecords in collections.items():
                 for r in colrecords.values():
-                    records.append(dict(__col_id__=collection,
-                                        __pid__=pid,
+                    records.append(dict(__collection_id__=collection,
+                                        __parent_id__=pid,
                                         **r))
 
         records, count = self.extract_record_set(records,
                                                  filters, None,
                                                  id_field, deleted_field)
 
-        deleted = [self.delete(r.pop('__col_id__'),
-                               r.pop('__pid__'),
+        deleted = [self.delete(r.pop('__collection_id__'),
+                               r.pop('__parent_id__'),
                                r[id_field],
                                id_field=id_field, with_deleted=with_deleted,
                                modified_field=modified_field,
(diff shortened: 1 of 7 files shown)


leplatrem:
@Natim I haven't touched the Redis backend... I think I need your support :) Don't hesitate to push commits, I won't force push...

Natim:
Do you need to grab the deleted when doing the * ?

leplatrem:
> Do you need to grab the deleted when doing the * ?

There is a `with_deleted` parameter.

Deleting the tombstones is done with `purge_deleted`

kinto/core/storage/redis.py
```diff
@@ -342,11 +356,14 @@ def delete_all(self, collection_id, parent_id, filters=None,
...
                    deleted_field=DEFAULT_DELETED_FIELD,
                    auth=None):
+
         records, count = self.get_all(collection_id, parent_id,
                                       filters=filters,
                                       id_field=id_field,
                                       modified_field=modified_field,
                                       deleted_field=deleted_field)
+        # XXX: Here we need to have the correct coolection_id and
+        # parent_id for the record id
```
leplatrem: I don't think you can do it via `get_all()`, since for each deleted record you'll need to figure out which parent_id and collection_id it was stored into. We might have to rewrite `delete_all()` :/

Natim: Yeah, I am currently blocked there, I have two solutions, either I am finding a way to give collection_id and parent_id to delete, either I am rewritting ``delete_all``

Would you like to chat a little bit about it with me?

Natim: I am wondering if we want the list of deleted item at all for that particular use case.

Natim:
After thinking this through I wonder if we cannot simply add a backend method to do : `Use deletion and purge by parent_id for bucket deletion` rather than changing existing ones.

leplatrem:
> we should probably add a new storage method rather than to change the existing ones so that we can do the delete all without having to return previous records or to create tombstones for instance.
>  
> I wonder if we cannot simply add a backend method
>

The current implementation of `delete_all()` in Redis is really poor, especially in terms of performance. Let's first try to improve it, I don't see it as impossible!

We can pair and take the time necessary, there is no hurry anyway!

kinto/core/storage/redis.py
```diff
@@ -268,31 +268,51 @@ def purge_deleted(self, collection_id, parent_id, before=None,
...
-        keys = ['{0}.{1}.{2}.deleted'.format(collection_id, parent_id,
-                                             _id.decode('utf-8'))
-                for _id in ids]
+        if collection_id is None:
+            collection_id = '*'
+
+        keys_pattern = '{0}.{1}.deleted'.format(collection_id, parent_id)
+
+        collections_keys = [key.decode('utf-8') for key in
+                            self._client.scan_iter(match=keys_pattern)]
```
Natim: Better than `keys()` :D

kinto/core/storage/memory.py
```diff
@@ -275,45 +258,84 @@ def delete(self, collection_id, parent_id, object_id,
...
+                        if parent_id_match.match(pid)}
+
+        records = []
+        for pid, collections in by_parent_id.items():
+            if collection_id is not None:
+                collections = {collection_id: collections[collection_id]}
+            for collection, colrecords in collections.items():
+                for r in colrecords.values():
+                    records.append(dict(__col_id__=collection,
+                                        __pid__=pid,
```
Natim: We could probably use `parent_id` instead of `pid`

kinto/core/storage/redis.py
```diff
@@ -330,14 +350,66 @@ def get_all(self, collection_id, parent_id, filters=None, sorting=None,
...
+            if len(ids) == 0:
+                continue
+
+            records_keys = ['{0}.{1}.{2}.records'.format(collection_id,
+                                                         parent_id,
+                                                         _id.decode('utf-8'))
+                            for _id in ids]
+            results = self._client.mget(records_keys)
+            collection_records = [dict(__col_id__=collection_id,
+                                       __pid__=parent_id,
```
Natim: ditto

Natim:
This is going to land soon :+1:

PR #714, opened 2016-07-12
Improvements on core.authorization and permission backend

leplatrem:
* [x] Refactor and simplify permissions methods to get rid of the callback `get_bound_permissions`
* [x] Rework `get_accessible_objects()` to obtain every objects in a single call (c.f. GET /permissions endpoint)
* [x] Add `get_objects_permissions()` to retrieve the permissions set for a list of objects ids
* [x] Update changelog
* [x] Make tests pass

### DIFF
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -70,22 +70,21 @@ class AuthorizationPolicy(object):
         if context.allowed_principals:
             allowed = bool(set(context.allowed_principals) & set(principals))
         else:
-            allowed = context.check_permission(
-                permission,
-                principals,
-                get_bound_permissions=self.get_bound_permissions)
+            object_id = context.permission_object_id
+            if self.get_bound_permissions is None:
+                bound_perms = [(object_id, permission)]
+            else:
+                bound_perms = self.get_bound_permissions(object_id, permission)
+            allowed = context.check_permission(principals, bound_perms)
 
         # If not allowed on this collection, but some records are shared with
         # the current user, then authorize.
         # The ShareableResource class will take care of the filtering.
         is_list_operation = (context.on_collection and
-                             'create' not in permission)
+                             not permission.endswith('create'))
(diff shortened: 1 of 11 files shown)


kinto/core/authorization.py
```diff
@@ -115,6 +114,8 @@ def __init__(self, request):
         self.get_prefixed_userid = functools.partial(prefixed_userid, request)
 
         self._check_permission = request.registry.permission.check_permission
+        permission = request.registry.permission
```
leplatrem: nit: move above

kinto/core/authorization.py
```diff
@@ -179,9 +183,17 @@ def fetch_shared_records(self, perm, principals, get_bound_permissions):
...
         """
-        ids = self.get_shared_ids(permission=perm,
-                                  principals=principals,
-                                  get_bound_permissions=get_bound_permissions)
+        bound_permissions = []
+        if get_bound_permissions is not None:
+            bound_permissions = [(o, p) for (o, p) in
+                                 get_bound_permissions(self._object_id_match,
+                                                       perm)
+                                 if o.endswith(self._object_id_match)]
```
leplatrem: Although this was move from every permission backend implementation to here, I believe it is not properly tested / specified

leplatrem:
@Natim feedback?

Natim:
I've read the code I understand that rather than using the get_bound_permissions function you prepare it and then you pass its result.

I think I would need some more explanation but it seems a good refactor to me.

leplatrem:
> using the get_bound_permissions function you prepare it and then you pass its result.

yes, that is it! actually, each backend implementation was resolving the result of this callback the same way! so it makes much more sense to call it before leveraging the backends. It makes the code simpler to read IMO.

Plus, it gives us a consistent way of fetching the permissions for several objects, necessary for #653 and #694 :) Win-Win :)

Natim:
r+

PR #715, opened 2016-07-13
Error if we create a record without deleting its tombstone

Natim:
I know that this is a bit cumbersome but it happened with kinto-signer.

Refs https://sentry.prod.mozaws.net/operations/kinto-stage/issues/341032/

### DIFF
--- a/kinto/core/storage/postgresql/migrations/migration_011_012.sql
+++ b/kinto/core/storage/postgresql/migrations/migration_011_012.sql
@@ -1,7 +1,9 @@
 -- Select all existing records and delete their tombstone if any.
 DELETE FROM deleted d
-WHERE (d.id, d.parent_id, d.collection_id) IN
-  (SELECT r.id, r.parent_id, r.collection_id FROM records r);
+USING records r
+WHERE d.id = r.id
+AND   d.parent_id = r.parent_id
+AND   d.collection_id = r.collection_id;
 
 -- Bump storage schema version.
 INSERT INTO metadata (name, value) VALUES ('storage_schema_version', '12');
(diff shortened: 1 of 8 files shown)


Natim:
We should probably add a migration function for Postgresql backend to make sure we remove all existing tombstones that have a valid record.

leplatrem:
I could have sworn it was the case !

GG r+

Natim:
@leplatrem Could you validate my migration file too?

kinto/core/storage/postgresql/schema.sql
```diff
@@ -241,4 +241,4 @@ INSERT INTO metadata (name, value) VALUES ('created_at', NOW()::TEXT);
 
 -- Set storage schema version.
 -- Should match ``kinto.core.storage.postgresql.PostgreSQL.schema_version``
-INSERT INTO metadata (name, value) VALUES ('storage_schema_version', '11');
+INSERT INTO metadata (name, value) VALUES ('storage_schema_version', '12');
```
leplatrem: You need to update the schema attribute on the storage backend :)

kinto/core/storage/postgresql/migrations/migration_011_012.sql
```diff
@@ -0,0 +1,7 @@
+-- Select all existing records and delete their tombstone if any.
+DELETE FROM deleted d
+WHERE (d.id, d.parent_id, d.collection_id) IN
+  (SELECT r.id, r.parent_id, r.collection_id FROM records r);
```
leplatrem: This should work, it can also be written like that:
```sql
DELETE 
FROM deleted d 
USING records r
WHERE d.id = r.id AND d.parent_id = r.parent_id AND d.collection_id = r.collection_id
```

Natim: Yes it is how I wrote it at first, I can revert :)

leplatrem:
r+

PR #718, opened 2016-07-15
Allow filtering and sorting by any attribute on buckets, collections and groups list endpoints

leplatrem:
@Natim r?

Bug found when implementing https://github.com/Kinto/kinto-http.js/pull/116


Natim:
r+

PR #720, opened 2016-07-18
Prepare 3.2.3

leplatrem:
@Natim r?


Natim:
r+ But let's release 3.3.0 so that we can release it on kinto-settings stage tomorrow with kinto-dist 0.7.0

PR #733, opened 2016-07-22
Clean-up load tests (fixes #258)

### LINKED ISSUE #258: Usage profiles for load tests
Currently the simultation load test runs random actions on the API, with arbitrary percentages.

Could be nice to have like a *presets* argument that would adjust the actions proportions:

* `kinto.js` : polling for changes and batch create/updates/delete
* Firefox security settings: lots of polling for changes with very rare create/update/delete
* exhaustive: make sure every action is run at least once ([as suggested by @Natim](https://github.com/mozilla-services/cliquet/pull/533#discussion_r43618102))

Related : #201

leplatrem:
Fixes #258 

@Natim r?


Natim:
f+

PR #735, opened 2016-07-25
Replace protocol by API (fixes #664)

### LINKED ISSUE #664: Shall we speak about protocol or interface?
The documentation is a little bit confusing:

> __Kinto-Core Protocol__
    A system of rules that explains the way to interact with the HTTP API endpoints (utilities, synchronization, headers etc.), and how data is organized (JSON responses etc.).

In my opinion, there is no such protocol except the one used to transmit data from and to kinto: http.

Instead we may speak about a _meta REST API_ to manipulate the wanted REST API. For instance, if I want to create an REST API about pictures, I will use the _meta REST API_ to create a bucket, then to create the collection `pictures` then to add a schema to a record of `pictures`.

The protocol in my example is still __HTTP__ with the 5 classic verbs like **GET**, **POST**, **PUT**, **PATCH** and **DELETE**. But the way I control kinto is by using that protocol to perform actions through the correct route like `POST /buckets`.

What do you think?

leplatrem:
Fixes #664 

Review? @enguerran

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,7 @@ This document describes changes between each past release.
 - Switch from cliquet-pusher to kinto-pusher in Dockerfile and tutorial.
 - List posssible response status on every endpoint documentation (#736)
 - Remove duplicated and confusing docs about generic resources
+- Replace the term ``protocol`` by ``API`` in documentation (fixes #664)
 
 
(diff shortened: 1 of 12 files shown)


enguerran:
Sorry, I was out of the Web for vacations ;)
I'm late but at least it LGTM.

PR #736, opened 2016-07-25
List every possible response status on every endpoint documentation

### LINKED ISSUE #279: list the possible responses status codes
When reading http://kinto.readthedocs.org/en/latest/api/buckets.html#bucket-delete I was confused about the signification of getting back a 403 on an unexistant bucket.

Maybe we could add for each API a list of possible response codes and a short explanation for each one ?

leplatrem:
Fixes #279 

r? @tarekziade


tarekziade:
This is really awesome!! I am wondering if we can use some reSt magic (like a glossary?) to avoid the duplication of the status codes text, so we just type something like :term:`400` ?

PR #739, opened 2016-07-26
Fix regression from dry-run feature

leplatrem:
I broke `master` when I manually merged #685 

These fixes should repair the damage :)

### DIFF
--- a/kinto/tests/core/test_storage_migrations.py
+++ b/kinto/tests/core/test_storage_migrations.py
@@ -6,7 +6,7 @@ from pyramid import testing
 
 from kinto.core.cache import postgresql as postgresql_cache
 from kinto.core.permission import postgresql as postgresql_permission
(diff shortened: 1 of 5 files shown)


kinto/tests/core/test_storage_migrations.py
```diff
@@ -208,6 +208,9 @@ def test_migration_12_clean_tombstones(self):
 class PostgresqlPermissionMigrationTest(unittest.TestCase):
     def __init__(self, *args, **kw):
         super(PostgresqlPermissionMigrationTest, self).__init__(*args, **kw)
+        from kinto.core.utils import sqlalchemy
+        if sqlalchemy is None:
```
Natim: Can we use a skipIf decorator instead?

leplatrem: There is a skip if already, but the test is instanciated anyway

Issue #591, opened 2016-05-13
Add the plugins versions


tarekziade:
It would be nice to have the versions of each plugin displayed on / in capabilities.

When deployed with kinto-dist I guess we can find them back, but that's extra work :)

almet:
That's a good idea! For this, we would have to:

- Add a version parameter in the cliquet [`add_api_capability` method](https://github.com/mozilla-services/cliquet/blob/master/cliquet/__init__.py#L126)
- It would show up directly [in the hello view](https://github.com/mozilla-services/cliquet/blob/master/cliquet/views/hello.py#L48)
- Write some tests to test this works as expected

Natim:
We need to add a global option to let OPS hide this version information because in production the version information can be used to detect vulnerable deployments if we were to have CVE in the future.

almet:
That's the case with the current deployment anyway (the kinto version is present).

We should open another issue for this. I'm not sure hiding the version will gain us some security? I would love to have mozsec input on this!

tarekziade:
@Natim That's orthogonal. If we want to avoid any leakage of information, we should firewall the access to those /__VIEWS   

That's why I've initially created the __ prefix standard

Natim:
Yes then maybe do we need to add a /__versions__ view rather than adding them on the root endpoint (that doesn't contains any __ prefix)

leplatrem:
>  If we want to avoid any leakage of information, we should firewall the access to those /__VIEWS 

Apparently this is not going to be the case: https://bugzilla.mozilla.org/show_bug.cgi?id=1268823#c2

tarekziade:
Even though we're not hiding them I am skeptical about the fact that hiding the version numbers is a good practice - apache and nginx don't do it afaik.

Let's ask @ulfr

Also: let's not add yet another info page, this is getting overkill

Natim:
It is usually a good practice to ask nginx to hide its version number: http://krischan.eu/index.php/2014/01/11/howto-hide-your-web-server-version-number-apache-lighttpd-nginx/

jvehent:
We don't hide release versions today. I don't think we need to in the future.
```bash
$ curl https://shavar.services.mozilla.com/__version__
{
    "commit": "a92303b58b9c9cea0b74896a9b6e011e964e51b9",
    "source": "https://github.com/mozilla-services/shavar.git",
    "version": "0.6.9"
}
```

Natim:
[...]

PR #747, opened 2016-07-29
Add the /__version__ endpoint with the kinto or kinto-dist released version.

Natim:
Refs #591 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem


docs/api/index.rst
```diff
@@ -11,6 +11,12 @@ API
 Changelog
 ---------
 
+1.9 (2016-07-29)
```
leplatrem: set to `unreleased` until it's released (unless we release 1.9 now)

CHANGELOG.rst
```diff
@@ -16,6 +16,7 @@ This document describes changes between each past release.
 - Added ability to plug custom StatsD backend implementations via a new ``kinto.statsd_backend``
   setting. Useful for Datadog™ integration for example (fixes #626).
 - Added a ``delete-collection`` action to the ``kinto`` command. (#727)
+- Added a ``/__version__`` endpoint with the version that have been deployed. (#747)
```
leplatrem: Mention it in a Protocol changes like other versions

docs/api/1.x/utilities.rst
```diff
@@ -80,10 +80,30 @@ This endpoint is suitable for a load balancer membership test.
...
 
 The returned value is a JSON mapping containing open source contribution
 information as advocated by https://www.contributejson.org
+
+
+GET /__version__
+==================
+
+Return a JSON mapping containing information about what distribution
+have been deployed by OPS.
```
leplatrem: typo: has

docs/api/1.x/utilities.rst
```diff
@@ -80,10 +80,30 @@ This endpoint is suitable for a load balancer membership test.
...
+
+::
+
+    {
+     "name":"kinto",
```
leplatrem: nitpick: identation

docs/api/1.x/utilities.rst
```diff
@@ -80,10 +80,30 @@ This endpoint is suitable for a load balancer membership test.
...
+     "version":"3.3.2",
+     "commit":"ab8db089ee63dc8e14f4bcfc427a86f311dd7e52",
+     "source":"https://github.com/Kinto/kinto"
```
leplatrem: not: in shavar it was the `git remote` url I think

version.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+cd "$(dirname "$0")"
+NAME=$(basename $PWD)
+SOURCE=$(git remote -v | grep origin | grep push | cut -f 2 | sed -e 's|git@|https://|g' | sed -e 's|github.com:|github.com/|g' | sed 's|.git (push)||g')
```
leplatrem: Why not just `SOURCE = $(shell git config remote.origin.url)`

version.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
leplatrem: Why not in `Makefile` ?

Natim: Oh we could do that yes. :+1:

leplatrem:
I'm a bit suprised by the complexity of the view code, compared to https://github.com/mozilla-services/shavar/commit/cebd41f123d83d2198eef67d6a7da923873cd136

Also, consider:
* Update the release docs if there is anything new to do
* Reify the view result since it runs many commands that will never have different results within the same run

CHANGELOG.rst
```diff
@@ -6,6 +6,10 @@ This document describes changes between each past release.
 3.4.0 (unreleased)
 ==================
 
+**Protocol**
+
+- Added a ``/__version__`` endpoint with the version that have been deployed. (#747)
```
leplatrem: nit: «has»

kinto/core/views/version.py
```diff
@@ -0,0 +1,70 @@
...
+import pkg_resources
+
+from pyramid.security import NO_PERMISSION_REQUIRED
+from kinto.core import Service
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+ORIGIN = os.path.dirname(os.path.dirname(os.path.dirname(HERE)))
+
+VERSION_PATH = 'version.json'  # Relative to where we start it.
+VERSION_JSON = None
```
leplatrem: nitpick: move that to `except` ? (not sure)

Natim:
The complexity comes from the fact that we have a fallback in case the version.json file is not present.

kinto/core/views/version.py
```diff
@@ -0,0 +1,70 @@
...
+    if VERSION_JSON is None:
+        settings = request.registry.settings
+        project_name = settings['project_name']
+
+        try:
+            # Try to get the project_name distribution
+            distribution = pkg_resources.get_distribution(project_name)
+        except pkg_resources.DistributionNotFound:
+            # If it doesn't work let's fallback to kinto
+            project_name = 'kinto'
```
leplatrem: project_name is always `kinto`. We use that for settings prefixes if I remember well

Natim:
The major problem is that you cannot add the commit of the tag inside the file if you didn't commit yet. So it is impossible to actually release with the version.json file.

kinto/core/views/version.py
Natim: No it's not, for syncto it is syncto for instance and readinglist, readinglist.

kinto/core/views/version.py
```diff
@@ -0,0 +1,70 @@
...
+            distribution = pkg_resources.get_distribution(project_name)
+
+        package_version = distribution.version
+        package_metadata = distribution._get_metadata(distribution.PKG_INFO)
+
+        home_page = [m for m in package_metadata if m.startswith('Home-page:')]
+
+        source = None
+        if home_page:
+            source = home_page[0].split(':', 1)[1].strip()
```
leplatrem: This is weird, it would deserve some comments and some safety checks maybe

version.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -e
+
+cd "$(dirname "$0")"
+NAME=$(basename $PWD)
+SOURCE=$(git config remote.origin.url | sed -e 's|git@|https://|g' | sed -e 's|github.com:|github.com/|g' | sed 's|\.git||g')
```
leplatrem: Why don't you give the git remote url ?

leplatrem: Oh I think I get it: ssh vs. https. Then no need to strip the trailing `.git` https://shavar.services.mozilla.com/__version__

PR #749, opened 2016-08-01
Add a bucket stats endpoint.

### LINKED ISSUE #360: Is there an api to get bucket/collection stats ?
Stats like no of documents, how much memory it holds etc will be useful.

Natim:
Fixes #360 

- [ ] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

f? @tarekziade

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,11 @@ This document describes changes between each past release.
 3.4.0 (unreleased)
 ==================
 
+**Protocol**
+
+- Add the ``/buckets/bid/stats`` endpoint to provide stats information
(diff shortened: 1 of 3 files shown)


tarekziade:
-1 on having a specific stats endpoint. We should store this info in the metadata with the other info. They are metadata like last_modified.

msathis:
@tarekziade I agree. GET request of `/bucket` and `/bucket/collection` should return the stats as well.

Natim:
I am closing for now since we don't have the use case yet we will come back to it later.
@msathis feel free to propose an implementation if you need it, we will review it and support you if you need it.

Issue #173, opened 2015-08-31
Implement quotas in Kinto


Natim:
When building Syncto I have seen that Firefox Sync handle quota and return the remaining space on device with a header.

   X-Weave-Quota-Remaining: 125

https://github.com/mozilla-services/server-syncstorage/blob/d961d1012e2c48921180934c369b136b83916a7d/syncstorage/views/decorators.py#L133-L137

I wonder if I can map this to ``Quota-Remaining`` for Kinto and open a Kinto.js specific error for quota.

leplatrem:
This looks like a feature request : add quota management in Kinto !
It could be a very relevant feature indeed, especially with public Kinto servers :)

Once we have an actual use-case, we could provide some specs here... Should quotas be defined at the collection level, per user, in config, in bucket/collections objects ? 

Meanwhile, for SyncTo, maybe you could just pass the header along. And in see if Kinto.js has an elegant way to introduce failures on custom response attributes like headers...

leplatrem:
Could you explain why you closed please ?

ametaireau:
I think that it is a useful feature to provide for Kinto, if we want to have servers open to the world at some point :-) Reopening for now, we can close it if/once we all agree this feature isn't useful.

Natim:
I closed because your answer answered my question.
We can rename the issue to implement Quota in Kinto but as we don't have the use case yet we can also create a new one then.

tarekziade:
We have a use case for Webextensions.

We need to implement the quotas defined in https://developer.chrome.com/extensions/storage#property-sync  - where Kinto is used as a Key-Value storage (so one collection/addon and one bucket/user I guess)

On the server side we should add:

-  **QUOTA_BYTES**: The maximum total amount (in bytes) of data that can be stored in sync storage, as measured by the JSON stringification of every value plus every key's length.

- **QUOTA_BYTES_PER_ITEM**: The maximum size (in bytes) of each individual item in sync storage, as measured by the JSON stringification of its value plus its key length.

- **MAX_ITEMS**:  	The maximum number of items that can be stored in sync storage.

leplatrem:
The main question is: is it enough to have it by bucket or per collection? Or should it be a complex combination of both?

Some insights on how it could work:

* Add a new built-in plugin (`kinto.plugins.quotas`)
* Introduce a new capability with exposed quotas values 
* Register a `AfterResourceChanged` subscriber that listens to every action on the record resource
[...]

PR #751, opened 2016-08-02
Make sure ACTIONS.DELETE returns old and new records.

Natim:
Refs #173 

- [x] Add tests.
- [x] Add a changelog entry.

This is a mandatory change for the quotas feature.

r? @leplatrem


Natim:
r=@glasserc in #752

PR #752, opened 2016-08-02
Quotas plugin

Natim:
Refs #173 
Based on #751 

- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem 

- [x] Add a listener that keep track of storage changes in the bucket.
- [x] Add a view that display some stats about the bucket (collection_count, record_count and storage_size)
- [x] Add a listener that uses the configuration and the resulting storage_size to raise a Quota error if needed.


tarekziade:
Like I said in https://github.com/Kinto/kinto/issues/173#issuecomment-236604658 I am still -1 on having a specific view for holding this information, which is of the same nature than last_modified.

Natim:
That's not the goal of this PR, we can change that before merging if need be.

Right now, the solution proposed by @leplatrem in https://github.com/Kinto/kinto/issues/173#issuecomment-222102623 which seems to me to be cleaner.

Updating the bucket on each change in it might have a big impact on the kinto protocol and create a lot of HTTP 412. We had a similar discussion before with kinto-attachment and kinto-history and it seems to be safer not to update user data with plugins.

tarekziade:
> Right now, I am trying to implement the solution proposed by @leplatrem in #173 (comment) which seems to me to be cleaner.

Please don't implement any code if we disagree on the design. This is a potential waste of time.

> Updating the bucket on each change in it might have a big impact on the kinto protocol and create a lot of HTTP 412

I am not sure why you want to update the bucket on each change. 

Please let's pause for a second here until we know exactly what we want. thx

Natim:
> I am not sure why you want to update the bucket on each change.

I will update this PR with the behavior your are mentioning to show you :)

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,112 @@
...
+    # Update the bucket quotas values for each impacted record.
+    for (uri, obj_id, old, new) in targets:
+        if is_bucket(uri):
+            old = strip_stats_keys(old)
+            new = strip_stats_keys(new)
+
+        if action == 'create':
+            bucket_info['storage_size'] += record_size(new)
+            if resource_name == 'collection':
+                bucket_info['collection_count'] += 1
```
Natim: This code is really likely to be subject to race conditions. (If two people update records in the same bucket for instance.) @tarekziade was suggesting doing something like Redis CAS or we could maybe do it directly with a PostgreSQL request so that it become a consistent change.

kinto/core/resource/__init__.py
```diff
@@ -342,10 +342,11 @@ def collection_delete(self):
         self._raise_412_if_modified()
 
         filters = self._extract_filters()
+        records, _ = self.model.get_records(filters=filters)
         deleted = self.model.delete_records(filters=filters)
```
glasserc: Is it possible for there to be records returned that are then not removed by the delete (i.e. a race condition with another client)? Or is this all happening in a transaction?

Natim: It is happening in a transaction thanks to the postgresql backend :)

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,212 @@
...
+
+def on_resource_changed(event):
+    """
+    Everytime an object is created/changed/deleted, we update the
+    bucket counters. The entries are served as read-only in the
+    :mod:`kinto.plugins.quotas.views` module.
+
+    If a new object override the quotas, we reject the request.
+    """
+    payload = copy.deepcopy(event.payload)
```
glasserc: Why do you copy the payload?

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,212 @@
...
+
+    storage = event.request.registry.storage
+
+    if action == 'delete' and resource_name == 'bucket':
+        try:
+            storage.delete(parent_id=bucket_uri,
+                           collection_id='quota',
+                           object_id='bucket_info')
+        except RecordNotFoundError:  # Pragma: no cover
+            pass
```
glasserc: Is the "no cover" pragma here because it's difficult to test, or because this is a "can't happen" scenario?

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,212 @@
...
+    # Update the bucket quotas values for each impacted record.
+    for (uri, obj_id, old, new) in targets:
+        old_size = record_size(old)
+        new_size = record_size(new)
+
+        if max_bytes_per_item is not None:
+            if new_size > max_bytes_per_item:
+                raise http_error(HTTPInsufficientStorage(),
+                                 errno=ERRORS.FORBIDDEN.value,
+                                 message=HTTPInsufficientStorage.explanation)
```
glasserc: I think this can trigger even on a delete, which seems wrong to me.

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,212 @@
...
+                             message=HTTPInsufficientStorage.explanation)
+
+    if bucket_max_items is not None:
+        if bucket_info['record_count'] > bucket_max_items:
+            raise http_error(HTTPInsufficientStorage(),
+                             errno=ERRORS.FORBIDDEN.value,
+                             message=HTTPInsufficientStorage.explanation)
+
+    if collection_max_bytes is not None:
+        if collection_info['storage_size'] > collection_max_bytes:
```
glasserc: Same concern about these checks being triggered on deletes -- I think we should always permit deletes, even if you're over quota.

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,212 @@
...
+                             errno=ERRORS.FORBIDDEN.value,
+                             message=HTTPInsufficientStorage.explanation)
+
+    if collection_max_items is not None:
+        if collection_info['record_count'] > collection_max_items:
+            raise http_error(HTTPInsufficientStorage(),
+                             errno=ERRORS.FORBIDDEN.value,
+                             message=HTTPInsufficientStorage.explanation)
+
+    storage.update(parent_id=bucket_uri,
```
glasserc: Could you tell me a little bit more about what `parent_id` means? I looked at the StorageBase class and I'm still kind of confused. In normal use, I would have assumed that it's a bucket ID, but here you're using a URL, and below you're using a collection URL. Where/how is this info getting stored? Is it in a special inaccessible Postgres table? Or is it in the same table user records are stored? If so, how do we protect the user from updating these records?

kinto/plugins/quotas/listener.py
Natim: You are right, it is a left over from the history plugin I took the idea from. No need for that here.

kinto/plugins/quotas/listener.py
Natim: I see it more as can't happen scenario but I can try to add a test for it.

kinto/plugins/quotas/listener.py
Natim: Yes, let me add a test for it then. :+1:

kinto/plugins/quotas/listener.py
Natim: You are correct that in case of the bucket we could use the bucket_id but for the collection, because we can have the same collection_id in two distinct buckets we have to use both the bucket_id and the collection id, but then you could also have the same id for a group and a collection the bucket, reason why in kinto (and not kinto,core) we made this convention of using the uri as the record unique id.

Natim: > how do we protect the user from updating these records?

We prevent it because we don't have any API for the user to access it.
If we were to add this API we would need to make sure the user cannot modify it.

kinto/plugins/quotas/listener.py
Natim: Actually it can happen if the collection is deleted and the plugin wasn't activated before the last change.

kinto/plugins/quotas/test_plugin.py
```diff
@@ -0,0 +1,543 @@
...
+        data = self.storage.get("quota", self.collection_uri,
+                                "collection_info")
+        self.assertStatsEqual(data, {
+            "record_count": 1,
+            "storage_size": storage_size
+        })
+
+
+class QuotaMaxBytesExceededCollectionSettingsListenerTest(
+        FormattedErrorMixin, QuotaCollectionMixin, QuotaWebTest):
```
glasserc: This is a clever way to define tests.

kinto/plugins/quotas/test_plugin.py
```diff
@@ -0,0 +1,543 @@
...
+
+    def get_app_settings(self, extra=None):
+        settings = super(QuotaMaxBytesPerItemExceededBucketListenerTest,
+                         self).get_app_settings(extra)
+        settings['quotas.bucket_test_max_bytes_per_item'] = '52'
+        return settings
+
+
+class QuotaCollectionMixin(object):
+    def test_507_is_raised_if_quota_exceeded_on_record_creation(self):
```
glasserc: I guess there's no need to check for 507s on updates?

Natim: Yes we should add a test for it.

kinto/plugins/quotas/test_plugin.py
```diff
@@ -0,0 +1,338 @@
...
+        self.create_record()
+        body = {'data': {'foo': 42}}
+        resp = self.app.post_json('%s/records' % self.collection_uri,
+                                  body, headers=self.headers, status=507)
+
+        self.assertFormattedError(
+            resp, 507, ERRORS.FORBIDDEN, "Insufficient Storage",
+            "There was not enough space to save the resource")
+
+        # Check that the storage was not updated.
```
glasserc: Should we verify that the record didn't get created too?

kinto/plugins/quotas/listener.py
glasserc: I think a race condition shouldn't be possible, since the ResourceChanged event happens within a transaction.

kinto/plugins/quotas/listener.py
Natim: I added tests and it seems that because the quota is removed before there it doesn't trigger the quota on delete already.

kinto/plugins/quotas/test_plugin.py
Natim: Tests added.

kinto/plugins/quotas/test_plugin.py
Natim: Yes I am adding a test there.

Natim:
I realized that transaction would only work with the Postgresql backend.

Natim:
Looking at it again, it seems that we need to add additional work:
- [ ] Add documentation
- [ ] Add a test on batch with multiple collection with records delete for instance
- [ ] Add tests for the default bucket

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Clients can check for the ``quotas`` capability in the
+:ref:`root URL endpoint <api-utilities>`.
+
+.. note::
+
+    In terms of performance, enabling this plugin generates two or
+    three additional queries on backends per request.
+
+* The bucket quota consider the bucket attributes, groups attributes,
```
glasserc: `considers`. I would also reword this, something like "A bucket's quota is a limit on the size of bucket attributes, group attributes, collection attributes, and record attributes."

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,91 @@
...
+  collections attributes, records attributes.
+* Deleted items size is considered to be null so if you add something
```
glasserc: `Deleted items'`, with the apostrophe to indicate possession (the size of the deleted items). I'd probably reword this as "Deleted items are considered to have zero size" or "size zero".

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,91 @@
...
+  and remove it, it will looks like it was never created for the
```
glasserc: look like

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,91 @@
...
+  quota even if its tombstone is still there.
+* The quota plugins only works with the transactional storage backends
```
glasserc: plugin

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,91 @@
...
+You can configure three types of quotas:
+
+* **QUOTA_BYTES**: The maximum total amount (in bytes) of data that
+  can be stored in sync storage, as measured by the Canonical JSON
+  stringification of every value plus every key's length.
+* **QUOTA_BYTES_PER_ITEM**: The maximum size (in bytes) of each
+  individual item in sync storage, as measured by the Canonical JSON
+  stringification of its value plus its key length.
+* **MAX_ITEMS**: The maximum number of items that can be stored in
+  sync storage.
```
glasserc: That can be stored in a collection or bucket?

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,91 @@
...
+  ``kinto.quotas.collection_{bucket_name}_max_items`` i.e
+  ``kinto.quotas.collection_blocklists_max_items``
+* Specifically for a given bucket collection using
+  ``kinto.quotas.collection_{bucket_name}_{collection_name}_max_bytes``,
+  ``kinto.quotas.collection_{bucket_name}_{collection_name}_max_bytes_per_item`` and
+  ``kinto.quotas.collection_{bucket_name}_{collection_name}_max_items`` i.e
+  ``kinto.quotas.collection_blocklists_certificates_max_items``
+
+
+How does it works?
```
glasserc: "work", because "it does work"

Natim:
Final r+ @leplatrem

CHANGELOG.rst
```diff
@@ -26,7 +26,8 @@ This document describes changes between each past release.
 - Added ability to plug custom StatsD backend implementations via a new ``kinto.statsd_backend``
   setting. Useful for Datadog™ integration for example (fixes #626).
 - Added a ``delete-collection`` action to the ``kinto`` command. (#727)
-- Added verbosity options to the ``kinto`` command (#745)
+- Added verbosity options to the ``kinto`` command. (#745)
+- Added a quotas plugin. (#752)
```
leplatrem: nitpick: added a built-in plugin that allows to define quotas per bucket or collection...

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+  (i.e PostgreSQL)
+
+
+Configuration
+=============
+
+You can configure three types of quotas:
+
+* **QUOTA_BYTES**: The maximum total amount (in bytes) of data that
+  can be stored in sync storage, as measured by the Canonical JSON
```
leplatrem: sync storage? canonical json?

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+  stringification of every value plus every key's length.
+* **QUOTA_BYTES_PER_ITEM**: The maximum size (in bytes) of each
+  individual item in sync storage, as measured by the Canonical JSON
+  stringification of its value plus its key length.
+* **MAX_ITEMS**: The maximum number of items that can be stored in
+  a collection or bucket.
+
+You can configure it at multiple levels.
```
leplatrem: in INI settings file

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+
+For buckets:
+
+* Globally for each buckets using ``kinto.quotas.bucket_max_bytes``,
```
leplatrem: typo: each bucket

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+  ``kinto.quotas.bucket_max_bytes_per_item`` and
+  ``kinto.quotas.bucket_max_items``
+* Specifically for some buckets using
+  ``kinto.quotas.bucket_{bucket_name}_max_bytes``,
```
leplatrem: nitpick: I think we use *bucket id* elsewhere ?

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+  ``kinto.quotas.bucket_{bucket_name}_max_bytes_per_item`` and
+  ``kinto.quotas.bucket_{bucket_name}_max_items`` i.e
```
leplatrem: nitpick: I think it's `e.g.` instead of `i.e.`

kinto/plugins/quotas/__init__.py
```diff
@@ -0,0 +1,14 @@
+from kinto.core.events import ResourceChanged
+
+from .listener import on_resource_changed
+
+
+def includeme(config):
+    config.add_api_capability('quotas',
+                              description='Quotas Management on Buckets.',
```
leplatrem: nitpick: *and collections*

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+        settings.get('quotas.collection_{}_{}'.format(bucket_id, name),
+                     # Global to all buckets collections
+                     settings.get('quotas.collection_{}'.format(name), None)))
+
+
+def on_resource_changed(event):
+    """
+    Everytime an object is created/changed/deleted, we update the
+    bucket counters. The entries are served as read-only in the
+    :mod:`kinto.plugins.quotas.views` module.
```
leplatrem: This doesn't seem to be relevant anymore

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+
+    If a new object override the quotas, we reject the request.
```
leplatrem: typo: override**s**: But, isn't *exceed* or *go over* instead of override?

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+                                                    'max_bytes_per_item')
+    collection_max_bytes = get_collection_settings(settings, bucket_id,
+                                                   collection_id, 'max_bytes')
+    collection_max_items = get_collection_settings(settings, bucket_id,
+                                                   collection_id, 'max_items')
+    collection_max_bytes_per_item = get_collection_settings(
+        settings, bucket_id, collection_id, 'max_bytes_per_item')
+
+    # XXX: Maybe we want to differenciate between
+    #      bucket/collection/records/group payload.
```
leplatrem: I don't understand this comment

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+        uri = '/'.join(parts)
+
+        old = impacted.get('old', {})
+        new = impacted.get('new', {})
+
+        targets.append((uri, obj_id, old, new))
+
+    try:
+        bucket_info = copy.deepcopy(
+            storage.get("quota", bucket_uri, 'bucket_info'))
```
leplatrem: Would be better to add explicitly parameter names like above

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+        bucket_info = {
+            "collection_count": 0,
+            "record_count": 0,
+            "storage_size": 0,
+        }
+
+    collection_info = {
+        "record_count": 0,
+        "storage_size": 0,
+    }
```
leplatrem: Why not follow the same approach as above for `bucket_info`, put this in except ?

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+        try:
+            collection_info = copy.deepcopy(
+                storage.get("quota", collection_uri, 'collection_info'))
+        except RecordNotFoundError:
+            pass
+
+    # Update the bucket quotas values for each impacted record.
+    for (uri, obj_id, old, new) in targets:
+        if resource_name == "collection":
+            collection_uri = uri
```
leplatrem: Why this ? (please add comment, `collection_uri` is defined far away)

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+        old_size = record_size(old)
+        new_size = record_size(new)
```
leplatrem: Maybe we should not compute these if no `*bytes*` setting is defined

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+
+    try:
+        bucket_info = copy.deepcopy(
+            storage.get("quota", bucket_uri, 'bucket_info'))
+    except RecordNotFoundError:
+        bucket_info = {
+            "collection_count": 0,
+            "record_count": 0,
+            "storage_size": 0,
+        }
```
leplatrem: What happens if the plugin is enabled once the bucket already has some data?

leplatrem: We may list that limitation in the docs

kinto/plugins/quotas/listener.py
Natim: Because in case of a bucket you don't have any collection_info but you are using the variable below.

kinto/plugins/quotas/listener.py
Natim: You are right that this might not be sufficient and we should probably get the collection_info here too.

This is to fix the batch on multiple collections for instance.

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+                collection_info['storage_size'] -= old_size
+                collection_info['storage_size'] += new_size
+        elif action == 'delete':
+            bucket_info['storage_size'] -= old_size
+            if resource_name == 'collection':
+                bucket_info['collection_count'] -= 1
+                # When we delete the collection all the records in it
+                # are deleted without notification.
+                collection_records, _ = storage.get_all('record',
+                                                        collection_uri)
```
leplatrem: * Add explicit argument names (`collection_id`, `parent_id`)
* How can you get the records if they were deleted then ?

kinto/plugins/quotas/listener.py
Natim: This is optimization yes.

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+You can configure three types of quotas:
+
+* **QUOTA_BYTES**: The maximum total amount (in bytes) of data that
+  can be stored in sync storage, as measured by the Canonical JSON
+  stringification of every value plus every key's length.
+* **QUOTA_BYTES_PER_ITEM**: The maximum size (in bytes) of each
+  individual item in sync storage, as measured by the Canonical JSON
+  stringification of its value plus its key length.
+* **MAX_ITEMS**: The maximum number of items that can be stored in
+  a collection or bucket.
```
leplatrem: Those upercase constants look weird (taken from storage.sync isn't it ?) Why don't they all start with QUOTA ? Why using them after all ?

kinto/plugins/quotas/listener.py
Natim: Everything that was in the bucket before the quota is "free"

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+    for (uri, obj_id, old, new) in targets:
+        if resource_name == "collection":
+            collection_uri = uri
+        old_size = record_size(old)
+        new_size = record_size(new)
+
+        if max_bytes_per_item is not None and action != "delete":
+            if new_size > max_bytes_per_item:
+                message = ("MAX_BYTES_PER_ITEM size exceeded: %d Bytes. "
+                           "MAX_BYTES_PER_ITEM is %d Bytes" % (
```
leplatrem: I don't think we should keep the use of those upercase constants here. Plus, it's QUOTA_BYTES_PER_ITEM

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+                             message=message)
+
+    if bucket_max_items is not None:
+        if bucket_info['record_count'] > bucket_max_items:
+            message = ("Bucket max items exceeded: %d records. "
+                       "MAX_ITEMS is %d records" % (
+                           bucket_info['record_count'], bucket_max_items))
+            raise http_error(HTTPInsufficientStorage(),
+                             errno=ERRORS.FORBIDDEN.value,
+                             message=message)
```
leplatrem: nitpick: use a small helper `raise_insufficient`

kinto/plugins/quotas/listener.py
```diff
@@ -0,0 +1,235 @@
...
+                           collection_info['record_count'],
+                           collection_max_items))
+            raise http_error(HTTPInsufficientStorage(),
+                             errno=ERRORS.FORBIDDEN.value,
+                             message=message)
+
+    storage.update(parent_id=bucket_uri,
+                   collection_id='quota',
+                   object_id='bucket_info',
+                   record=bucket_info)
```
leplatrem: use constants for `quota` and `bucket_info` values?

kinto/plugins/quotas/test_plugin.py
```diff
@@ -0,0 +1,889 @@
...
+        self.storage.delete(parent_id='/buckets/test',
+                            collection_id='quota',
+                            object_id='bucket_info')
+        transaction.commit()
+        self.app.delete(self.bucket_uri, headers=self.headers)
+
+    #
+    # Collection
+    #
+    def test_stats_are_not_accessible_if_collection_does_not_exists(self):
```
leplatrem: typo: does not exist

kinto/plugins/quotas/listener.py
Natim: > How can you get the records if they were deleted then ?

This is because they were not deleted in this yet.

docs/api/1.x/quotas.rst
Natim: Maybe @tarekziade have got a point about this.

kinto/plugins/quotas/test_plugin.py
```diff
@@ -0,0 +1,889 @@
...
+        self.assertStatsEqual(data, {
+            "collection_count": 1,
+            "record_count": 3,
+            "storage_size": 232
+        })
+
+    def test_bulk_delete(self):
+        self.create_bucket()
+        self.create_collection()
+        self.create_record()
```
leplatrem: Why do you create a collection and a record here ?

Natim: I did that in order to validate that DELETE works properly.

kinto/plugins/quotas/listener.py
Natim: Actually I find it easier to read like that because it removes a level of indirection and make it explicit that it will raise. I don't think creating a level of indirection with only another function call adds much value here.

kinto/plugins/quotas/listener.py
Natim: I find it very explicit, how would you like it to be?

leplatrem: I guess that those upercase strings refer to the ones in the docs ? `QUOTA_BYTES_PER_ITEM` or `QUOTA_BYTES` ? But you used `MAX_BYTES_PER_ITEM` and `MAX_BYTES`. Why not just something like `Maximum bytes size per item exceeded (%d > %d Bytes)` ? (*not crucial, but while we're at it..*)

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+    three additional queries on backends per request.
+
+* A bucket's quota is a limit on the size of bucket attributes, group
+  attributes, collection attributes, and record attributes.
+* Deleted items are considered to have a size zero so if you add something
+  and remove it, it will look like it was never created for the
+  quota even if its tombstone is still there.
+* The quota plugin only works with the transactional storage backends
+  (e.g. PostgreSQL)
+
```
leplatrem: Note: add mention of handling of existing data when plugin is enabled

docs/api/1.x/quotas.rst
```diff
@@ -0,0 +1,93 @@
...
+You can configure three types of quotas:
+
+* **QUOTA_BYTES**: The maximum total amount (in bytes) of data that
+  can be stored in a bucket or collection, as measured by the JSON
+  stringification of every value plus every key's length.
+* **QUOTA_BYTES_PER_ITEM**: The maximum size (in bytes) of each
+  individual item in the bucket or collection, as measured by the JSON
+  stringification of its value plus its key length.
+* **MAX_ITEMS**: The maximum number of items that can be stored in
+  a collection or bucket.
```
leplatrem: Those uppercase strings come from https://developer.chrome.com/apps/storage. Not sure if they bring anything compared to simple bold string (`Maximum total bytes`, `Maximum bytes per object`, `Maximum number of objects`...).

I realize that we use *objects* in Kinto and not *items*

kinto/plugins/quotas/listener.py
glasserc: Try adding a test where the quota is reduced after a bunch of objects have been stored, and then doing a delete.

PR #753, opened 2016-08-11
Remove the Redis backend from kinto core.

### LINKED ISSUE #712: Challenge the relevance of Redis backends in core
My opinion is that having Redis in the main Kinto repo, and not in a separate repo:

* is mainly an intellectual exercise to demostrate that the backend interface is agnostic
* introduces a overhead of maintenance (redis methods to port, redis server to run the tests, redis on travis)
* has happened because Remy and Alex are huge fans of Redis :)

Whereas concretely, the Redis has a number of limitations that prevents its usage in production:

* there is no «per-request» transaction
* there is no schema migration
* the filtering and sorting is done in memory in python, after deserializing each records from json strings
* [some methods are implemented naively](https://github.com/Kinto/kinto/blob/3.2.1/kinto/core/storage/memory.py#L31-L40)

Using the redis backend for cache and postgres for storage and permissions can be relevant, but the absence of transaction (two-phase commit) prevents redis to be used along postgres for storage or permission backends.

Of course we could spend ~~some~~ a lot of time improving all this, but we don't have any use-case and it looks like the few production deployments we've heard of (mozilla, kinto-heroku etc.) are using postgresql. So we could simply move the redis backend to a separate repo :)

**What do you think ?**

Natim:
Fixes #712 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem 

Tests will need a release of kinto-redis to pass.


kinto/tests/core/support.py
```diff
@@ -99,9 +99,9 @@ def make_app(self, settings=None, config=None):
     def get_app_settings(self, additional_settings=None):
         settings = DEFAULT_SETTINGS.copy()
 
-        settings['storage_backend'] = 'kinto.core.storage.redis'
-        settings['cache_backend'] = 'kinto.core.cache.redis'
-        settings['permission_backend'] = 'kinto.core.permission.redis'
+        settings['storage_backend'] = 'kinto_redis.storage'
+        settings['cache_backend'] = 'kinto_redis.cache'
+        settings['permission_backend'] = 'kinto_redis.permission'
```
glasserc: Does this mean that `kinto_redis` is required to run tests?

Natim: Currently it is, but we can change that.

glasserc:
r+! :100:

CHANGELOG.rst
```diff
@@ -11,6 +11,7 @@ This document describes changes between each past release.
 - ``kinto --version`` was renamed ``kinto version``
 - ``ResourceChanged`` and ``AfterResourceChanged`` events now return
   ``old`` and ``new`` records for the ``delete`` action. (#751)
+- Redis backends are not parts of the core anymore. (#712)
```
leplatrem: * Note: this is a breaking change (would lead to kinto 4.0).
* I suggest that we give some details. For example `In settings, replace kinto.core.cache.redis by kinto_redis.cache`
* I don't think we use the term `backend` for the listener, so some details about that would be required too I think
* typo : *are not part of*

docs/core/permission.rst
```diff
@@ -359,7 +359,7 @@ PostgreSQL
 Redis
 -----
 
-.. autoclass:: kinto.core.permission.redis.Permission
+.. autoclass:: kinto_redis.permission.Permission
```
leplatrem: Does this work ?

Natim: It works if kinto_redis is installed in the docs build venv.

docs/tutorials/notifications-custom.rst
```diff
@@ -268,6 +268,10 @@ Kinto should load the listeners without errors:
     Starting subprocess with file monitor
     2016-01-21 16:21:59,941 INFO  [kinto.core.initialization][MainThread] Setting up 'redis' listener
 
+.. note::
+
+   You will need to install `kinto-redis` to use this listener.
```
leplatrem: * rSt requires double back ticks
* Wouldn't it be better to have this note on top ?

docs/configuration/settings.rst
```diff
@@ -128,7 +128,7 @@ Storage
 
 .. code-block:: ini
 
-    kinto.storage_backend = kinto.core.storage.redis
+    kinto.storage_backend = kinto_redis.storage
```
leplatrem: *me thinking out loud*: Since those settings values are given as examples, I think we could give a fully qualified postgres config instead?

leplatrem:
Apart from missing details about the breaking changes, it looks very good to me!

PR #756, opened 2016-08-12
Add a test to show a problem in the history plugin.

Natim:
While working on the quotas plugin, I used a similar way of going through the `impacted_record` list and I discovered a bug that applies to the history plugin too.


leplatrem:
r+

PR #760, opened 2016-08-16
Preserve unknown fields by default

### LINKED ISSUE #719: Preserve unknown fields by default in core resources
Now that we have kinto.core, does it make sense to keep `preserve_unknown` to False by default ?

It obliges us to set it to True on every kinto resources (records, buckets, collections, groups).

Would you agree to invert the default value and get rid of specializations in kinto views ?

Related:
* #158 
* #239 
* #469 
* #718

leplatrem:
Fixes #719

r? @Natim

### DIFF
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -212,12 +212,17 @@ class UserResource(object):
         return known_fields
 
     def is_known_field(self, field):
-        """Return ``True`` if `field` is defined in the resource mapping.
+        """Return ``True`` if `field` is defined in the resource schema.
+        If the resource schema allows unknown fields, this will always return
+        ``True``.
 
         :param str field: Field name
         :rtype: bool
 
         """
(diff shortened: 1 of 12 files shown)


kinto/core/resource/__init__.py
```diff
@@ -592,16 +597,20 @@ def process_record(self, new, old=None):
...
         :rtype: dict
         """
-        new_last_modified = new.get(self.model.modified_field)
-        not_specified = old is None or self.model.modified_field not in old
+        modified_field = self.model.modified_field
+        new_last_modified = new.get(modified_field)
 
-        if new_last_modified is None or not_specified:
+        # Drop the new last_modified if it is not an integer.
+        is_integer = isinstance(new_last_modified, int)
```
Natim: Is it enough? Don't you need to try to convert it to an integer?

Natim:
r+

PR #763, opened 2016-08-17
Drop (unused) notion of unique fields in kinto.core

leplatrem:
Originally, for the reading list service, we had the notion of *unique fields*, because we wanted to ensure that stored articles were «unique».

The whole code in charge of this can be removed, mainly because:
* it was never used beyond that original use-case
* it is useless when data is encrypted
* it requires to scan the collection on each create/update before checking
* it allows to remove quite a lot of code

- [x] Update documentation.
- [x] Update tests.
- [x] Add a changelog entry.

r? @Natim


Natim:
I am always a bit reluctant to remove feature that are implemented and working.
The simple fact that we needed it for readinglist seems important to me.

However this seems to be a edge case that would be implemented using a plugin if we need it again.

Natim:
r+

PR #769, opened 2016-08-18
Remove kinto.id_generator from docs (fixes #757)

### LINKED ISSUE #757: kinto.id_generator option doesn't work
#### Issue
As per the [docs](http://kinto.readthedocs.io/en/latest/configuration/settings.html?highlight=id_generator#feature-settings), we should be able to define a `kinto.id_generator` options that will work for records, but this is not the case in the current (3.4.0.dev0) implementation.

#### Easy workaround ####
As an easy fix, you can specify the `record` type as in `kinto.record_id_generator`.

leplatrem:
Fixes #757 

I was trying to fix the expected behaviour of the `kinto.id_generator` setting, which should override every id default generators.

The problem I face is that we currently have no way to distinguish between default settings and specified settings at the place we consume them. Since I can't tell if the setting was defined in .ini or not, I can't do much :(

I thus decided to remove `kinto.id_generator` from docs :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ This document describes changes between each past release.
 **Internal changes**
 
 - Fix documentation of errors codes (fixes #766)
+- ``kinto.id_generator`` was removed from documentation since it does not
+  behave as expected (fixes #757, thanks @doplumi)
 
 
 4.0.0 (2016-08-17)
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -9,6 +9,8 @@ This document describes changes between each past release.
 **Internal changes**
 
 - Fix documentation of errors codes (fixes #766)
+- ``kinto.id_generator`` was removed from documentation since it does not
```
Natim: Don't we want to rather fix the behavior?
I like the idea that you can configure all id_generators with one config.

leplatrem: I spent some time but gave up. Don't hesitate to try but I think it's harder than expected :)

PR #770, opened 2016-08-18
Allow objects ids to start with _ or - (fixes #668)

### LINKED ISSUE #668: can record IDs start with an underscore?
Does Kinto reserve some of the record ID namespace for `__magic__`? My use case involves using Kinto as a generic key-value store, and since I don't control the keyspace, I have to translate between Kinto keys and my keys. I was figuring to do a string replace and turn everything not alphanumeric into an `_`-encoded character, like `_20_` for space, `_5f_` for underscore, etc. But depending on what exact set of record IDs is valid, I might have to prefix them first with something alphanumeric like "key-".

leplatrem:
Fixes #668 

@glasserc r? if this is not relevant anymore, please close and drop the PR

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ This document describes changes between each past release.
 4.1.0 (unreleased)
 ------------------
 
+**Protocol**
+
+- Allow objects ids to start with ``_`` or ``-`` (fixes #668)
+
+Protocol is now at version **1.10**. See `API changelog`_.
+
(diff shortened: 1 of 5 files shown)


Natim:
I wish we could prevent this IDs from starting with something else than a letter or digit.

glasserc:
I don't think we need this; I've already worked around this in ext.storage.sync. (See `keyToId` in https://reviewboard.mozilla.org/r/55838/diff/22.)

PR #771, opened 2016-08-18
Split tests from package

### LINKED ISSUE #605: [cliquet] Split tests from kinto package
original : https://github.com/mozilla-services/cliquet/issues/267

Should we have tests within the kinto package or outside ? that was discussed but I don't think a decision was made.

Another related question is not wether tests is in kinto/ or outside, but rather : do we ship tests when we release kinto ? some people strip them when they build the release.

leplatrem:
Fixes #605 

I have wanted this for a *looooooong* time, and here you go, I found the motivation to do it.

A good practice is to have the tests outside the package, and the reusable testing tools are in `kinto.core.testing`.

I made a lot of cleaning for `BaseWebTest` classes and stuff

r? @Natim


.travis.yml
```diff
@@ -38,7 +38,7 @@ matrix:
     - env: TOX_ENV=flake8
       install: pip install flake8
```
Natim: You can probably remove this as well if you want to use the tox venv. I remember doing this to make it run faster than having to install all the dependencies.

Makefile
```diff
@@ -87,7 +87,7 @@ tests-once: install-dev version-file install-postgres install-monitoring
 	$(VENV)/bin/py.test --cov-report term-missing --cov-fail-under 100 --cov kinto
 
 flake8: install-dev
-	$(VENV)/bin/flake8 kinto
+	$(VENV)/bin/flake8 kinto tests
```
Natim: :+1:

.travis.yml
leplatrem: I was thinking maybe it's useful for not sending reports to coverals ?

kinto/core/testing.py
```diff
@@ -1,34 +1,29 @@
...
-import mock
 import os
-import threading
-import functools
+from collections import defaultdict
 
 try:
     import unittest2 as unittest
-except ImportError:
+except ImportError:  # pragma: no cover
```
Natim: I think since we are not supporting python2.6 anymore we can probably just import unittest here.

Natim: unittest2 is a backport of the new features added to the unittest testing framework in Python 2.7 

https://pypi.python.org/pypi/unittest2

Natim:
r+wc

PR #780, opened 2016-08-22
Permissions endpoint can now be filtered, sorted and paginated

leplatrem:
Ref #600 #653 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.
- [x] Find a way to allow GET as anonymous `#needhelp`
- [ ] Approach may be a bit too hacky (c.f `last_modified` missing from perm entries)

feedback? @Natim

### DIFF
--- a/kinto/core/errors.py
+++ b/kinto/core/errors.py
@@ -171,6 +171,7 @@ def raise_invalid(request, location='body', name=None, description=None,
     :raises: :class:`~pyramid:pyramid.httpexceptions.HTTPBadRequest`
     """
     request.errors.add(location, name, description, **kwargs)
+    request.errors.request = request  # Needed by json_error_handler to reapply_cors
     response = json_error_handler(request.errors)
     raise response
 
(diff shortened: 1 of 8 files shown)


Natim:
> Find a way to allow GET as anonymous

I think this is when you have a permission dynamic set on it and that you match on the ``system.Everyone`` principal.

Natim:
I am wondering if we should go that way and not just add filtering and sorting to the previous service.

Natim:
After investigation it seems that you are using a `UserResource` rather than a `ShareableResource` Data from a UserResource are linked to a user so you cannot link data to a non authenticated user.

If we use a ShareableResource it should work.

Natim:
Ok I manage to make it works.

Natim:
What would it actually takes to add the `last_modified` value? Do we need it?

kinto/core/errors.py
```diff
@@ -171,6 +171,7 @@ def raise_invalid(request, location='body', name=None, description=None,
     :raises: :class:`~pyramid:pyramid.httpexceptions.HTTPBadRequest`
     """
     request.errors.add(location, name, description, **kwargs)
+    request.errors.request = request  # Needed by json_error_handler to reapply_cors
```
leplatrem: https://github.com/mozilla-services/cornice/issues/378

PR #785, opened 2016-08-22
Fix bug with subobject filtering on strict resources

leplatrem:
r? @Natim


Natim:
I think with this patch it will not be possible anymore to have a dot in the field name, is it?

leplatrem:
> I think with this patch it will not be possible anymore to have a dot in the field name, is it?

This patch only applies to strict schema resources (only history basically). Kinto resources are schemaless (preserve_unknown=True), thus any field is considered as «known».

Natim:
oh ok

PR #787, opened 2016-08-23
Improve max line length according to PEP-8

Natim:
r? @leplatrem 

https://www.python.org/dev/peps/pep-0008/#maximum-line-length


leplatrem:
r+

PR #799, opened 2016-08-29
Support for search on string fields

### LINKED ISSUE #791: Support for filtering records based on a text search
As discussed originally [here](http://kinto.readthedocs.io/en/stable/api/1.x/filtering.html#comment-2858589565), I believe there needs to be some basic support for filtering based on text search on a string field. -- something like `?in_myStringField=foo` matching any records where `myStringField` is `LIKE '%foo%'`

I understand that one, can add elastic-search plugin to benefit from full-text search plus many other things it supports, but that sounds like a long way to go if all you need is to build a simple CRUD application with standard text search.

mvalipour:
Fixes #791 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @

### DIFF
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -922,7 +922,7 @@ class UserResource(object):
                 )
                 continue
 
-            m = re.match(r'^(min|max|not|lt|gt|in|exclude)_([\w\.]+)$', param)
+            m = re.match(r'^(min|max|not|lt|gt|in|exclude|like)_([\w\.]+)$', param)
             if m:
                 keyword, field = m.groups()
                 operator = getattr(COMPARISON, keyword.upper())
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -922,7 +922,8 @@ class UserResource(object):
                 )
                 continue
 
-            m = re.match(r'^(min|max|not|lt|gt|in|exclude|like)_([\w\.]+)$', param)
+            allKeywords = '|'.join([i.name.lower() for i in COMPARISON])
+            m = re.match(r'^('+allKeywords+')_([\w\.]+)$', param)
             if m:
                 keyword, field = m.groups()
                 operator = getattr(COMPARISON, keyword.upper())
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -613,6 +613,7 @@ class Storage(StorageBase):
             COMPARISON.NOT: '<>',
             COMPARISON.IN: 'IN',
             COMPARISON.EXCLUDE: 'NOT IN',
+            COMPARISON.LIKE: 'ILIKE',
         }
 
         conditions = []
@@ -653,6 +654,9 @@ class Storage(StorageBase):
                 # WHERE field IN ();  -- Fails with syntax error.
                 if len(value) == 0:
                     value = (None,)
+
+            if filtr.operator == COMPARISON.LIKE:
+                value = '%{0}%'.format(value)
 
             # Safely escape value
             value_holder = '%s_value_%s' % (prefix, i)
--- a/tests/core/resource/test_filter.py
+++ b/tests/core/resource/test_filter.py
@@ -8,13 +8,18 @@ class FilteringTest(BaseTest):
     def setUp(self):
         super(FilteringTest, self).setUp()
         self.patch_known_field.start()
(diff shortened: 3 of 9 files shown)


kinto/core/resource/__init__.py
```diff
@@ -922,7 +922,8 @@ def _extract_filters(self, queryparams=None):
                 )
                 continue
 
-            m = re.match(r'^(min|max|not|lt|gt|in|exclude)_([\w\.]+)$', param)
+            allKeywords = '|'.join([i.name.lower() for i in COMPARISON])
```
mvalipour: I felt it's naive to have this regex listing exactly what is in `Comparison` enum.
Only downside is that it also includes `eq_` as a prefix, which surely should be fine?

tests/core/resource/test_filter.py
```diff
@@ -8,13 +8,18 @@ class FilteringTest(BaseTest):
     def setUp(self):
         super(FilteringTest, self).setUp()
         self.patch_known_field.start()
-        for i in range(6):
```
mvalipour: very cryptic to have the state of the world in a for loop for our unit tests. do you agree that it benefits us to have a nicely readable and solid list of records instead?

mvalipour:
Once, we are all happy with the solution, I can go ahead and add documentation, etc.

Natim:
So if I understand the code correctly it is going to used like that: ``/records?like_name=Mo``

kinto/core/storage/postgresql/__init__.py
```diff
@@ -654,6 +655,9 @@ def _format_conditions(self, filters, id_field, modified_field,
                 if len(value) == 0:
                     value = (None,)
 
+            if filtr.operator == COMPARISON.LIKE:
+                value = '%{0}%'.format(value)
```
Natim: A test is missing for this case.

tests/core/resource/test_filter.py
Natim: Yes I guess it helps.

kinto/core/storage/postgresql/__init__.py
Natim: Apparently tests from test_filter never enters _format_condition.

Natim: We need to add tests for the storage as well so that we can test that storage backends correctly supports the new operator. See: https://github.com/Kinto/kinto/blob/master/tests/core/test_storage.py#L386

mvalipour: Yes, that's were I was stuck. There doesn't seem to be any unit-tests running in postgres mode. So is covering it in `test_storage` enough?

mvalipour:
@Natim I thought `contains` might be a confusing term. I'm happy to debate this, but maybe `like_` or even `search_` will speak better about what it does?

kinto/core/storage/postgresql/__init__.py
Natim: > So is covering it in test_storage enough?

Yes I think so.

Natim:
I don't have a strong opinion on this. I guess we can keep `like_` for now.

kinto/core/storage/postgresql/__init__.py
Natim: Yes I will be enough and tested for all backends. Including the `kinto-redis` community one.

mvalipour:
@Natim Am I right assuming this change is not classified as *a change in HTTP API* ?

tests/core/test_storage.py
```diff
@@ -1209,6 +1209,15 @@ def test_number_of_fetched_records_can_be_limited_in_settings(self):
         results, count = limited.get_all(**self.storage_kw)
         self.assertEqual(len(results), 2)
 
+    def test_records_filtered_when_searched_by_string_field(self):
```
Natim: Please add the test in the global StorageTest rather than the specific PostgreSQL one. All implemented backend should support this feature.

Natim:
> Am I right assuming this change is not classified as a change in HTTP API ?

Yes it is a big change in the HTTP API because we are adding a parameter :)

tests/core/test_storage.py
mvalipour: fair enough. done.

Natim:
r+

Natim:
I updated the API version with: https://github.com/Kinto/kinto/commit/d08523c0ea0b9ee054a042b30285062369e45d42

PR #806, opened 2016-09-06
Changelog @leplatrem review.

Natim:
r? @leplatrem 
Refs: https://github.com/Kinto/kinto/commit/d08523c0ea0b9ee054a042b30285062369e45d42#commitcomment-18907205

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -14,7 +14,7 @@ Changelog
 1.10 (unreleased)
 '''''''''''''''''
 
-- Add text case insensitive filtering on plural endpoints (e.g ``?like_person=Tim``)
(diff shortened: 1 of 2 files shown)


docs/api/index.rst
```diff
@@ -14,7 +14,7 @@ Changelog
 1.10 (unreleased)
 '''''''''''''''''
 
-- Add full-text case insensitive filtering on plural endpoints (e.g ``?like_person=Tim``)
+- Add text case insensitive filtering on plural endpoints (e.g ``?like_person=Tim``)
```
glasserc: I personally think full-text isn't wrong here, but if we're going to change it, I think we should include a stronger word than "filtering". How about "Add pattern-matching query to filtering on plural endpoints" or "Add substring query to filtering on plural endpoints"?

Issue #805, opened 2016-09-05
DELETE fails when history plugin is enabled


juhieta:
Hi, I updated to release 4.1.1 and enabled the history plugin. After that I started getting errors with DELETE operations.

For example, trying to delete all collections:

```
DELETE /v1/buckets/test/collections:

File "/usr/local/lib/python3.5/dist-packages/pyramid/tweens.py", line 22, in excview_tween
    response = handler(request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid_tm/__init__.py", line 101, in tm_tween
    reraise(*exc_info)
  File "/usr/local/lib/python3.5/dist-packages/pyramid_tm/compat.py", line 15, in reraise
    raise value
  File "/usr/local/lib/python3.5/dist-packages/pyramid_tm/__init__.py", line 90, in tm_tween
    manager.commit()
  File "/usr/local/lib/python3.5/dist-packages/transaction/_manager.py", line 123, in commit
    return self.get().commit()
  File "/usr/local/lib/python3.5/dist-packages/transaction/_transaction.py", line 265, in commit
    self._callBeforeCommitHooks()
  File "/usr/local/lib/python3.5/dist-packages/transaction/_transaction.py", line 336, in _callBeforeCommitHooks
    hook(*args, **kws)
  File "/code/kinto/core/events.py", line 76, in _notify_resource_events_before
    request.registry.notify(event)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/registry.py", line 91, in notify
    [ _ for _ in self.subscribers(events, None) ]
  File "/usr/local/lib/python3.5/dist-packages/zope/interface/registry.py", line 442, in subscribers
    return self.adapters.subscribers(objects, provided)
  File "/usr/local/lib/python3.5/dist-packages/zope/interface/adapter.py", line 596, in subscribers
    subscription(*objects)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/config/adapters.py", line 130, in subscriber_wrapper
    return derived_subscriber(*arg)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/config/adapters.py", line 103, in derived_subscriber
    return subscriber(arg[0])
  File "/code/kinto/plugins/history/listener.py", line 62, in on_resource_changed
    bucket_perms = perms_by_object_id[bucket_uri]
KeyError: '/buckets/test' lang=None uid=...
```

Seems that the code is not getting the permissions as this is the failing line:

[...]

PR #808, opened 2016-09-07
History plugin and DELETE on plural endpoints

leplatrem:
I could not reproduce #805 locally with httpie. I added some tests, but can't make them fail either.


leplatrem:
I'll merge the PR as those tests don't harm.

Issue #604, opened 2016-05-20
[cliquet] Add stability tests for storage using Vaurien


tarekziade:
original : https://github.com/mozilla-services/cliquet/issues/216
that coud be useful to test the heartbeat

leplatrem:
Closing for lack of interest

PR #809, opened 2016-09-07
Fix timeout on heartbeat with PostgreSQL (fixes #804)

### LINKED ISSUE #804: deadlock on __heartbeat__
If I set up the same postgresql database/user for the storage, cache and permission connectors, I get a thread deadlock on the second call to __heartbeat__, leading to a blocking call, that makes the HB fails.

Looks like a DB connector race condition issue

leplatrem:
I could see in PostgreSQL logs that heartbeat operations were starting a transaction without closing them (either commit/rollback), leading into errors like:
```
process 14701 still waiting for ShareLock on transaction 3024907 after 1000.114 ms
```

* [x] Quickfix
* [x] Find way to test properly (e2e with vaurien #604 ?)
* [x] Update changelog

### DIFF
--- a/kinto/core/cache/__init__.py
+++ b/kinto/core/cache/__init__.py
@@ -1,7 +1,5 @@
 import random
 
-import transaction
-
 from kinto.core.logs import logger
 
 
@@ -94,7 +92,5 @@ def heartbeat(backend):
         except:
             logger.exception("Heartbeat Failure")
             return False
-        finally:
-            transaction.abort()
 
     return ping
--- a/kinto/core/cache/__init__.py
+++ b/kinto/core/cache/__init__.py
@@ -1,5 +1,7 @@
 import random
 
+import transaction
+
 from kinto.core.logs import logger
 
 
@@ -92,5 +94,7 @@ def heartbeat(backend):
         except:
             logger.exception("Heartbeat Failure")
             return False
+        finally:
+            transaction.abort()
 
     return ping
--- a/kinto/core/views/heartbeat.py
+++ b/kinto/core/views/heartbeat.py
@@ -1,4 +1,6 @@
 from concurrent.futures import ThreadPoolExecutor, wait
+
+import transaction
 from pyramid.security import NO_PERMISSION_REQUIRED
 
 from kinto import logger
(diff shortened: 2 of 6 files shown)


kinto/core/cache/__init__.py
```diff
@@ -92,5 +94,7 @@ def ping(request):
         except:
             logger.exception("Heartbeat Failure")
             return False
+        finally:
+            transaction.abort()
```
Natim: Don't you want to commit rather than abort?

leplatrem: Why not... it will leave a useless trace in the DB, but we delete it on each call anyway.
Your call!

leplatrem:
@Natim @tarekziade r?

kinto/core/views/heartbeat.py
```diff
@@ -17,6 +19,8 @@ def get_heartbeat(request):
     def heartbeat_check(name, func):
         status[name] = False
         status[name] = func(request)
+        # Rollback any open transaction (will release table locks etc.)
+        transaction.abort()
```
Natim: Don't we want to do that at a upper level? I mean make sure the transaction have been canceled or committed each time a response have been sent?

kinto/core/views/heartbeat.py
```diff
@@ -19,7 +19,10 @@ def get_heartbeat(request):
     def heartbeat_check(name, func):
         status[name] = False
         status[name] = func(request)
-        # Rollback any open transaction (will release table locks etc.)
+        # Since the heartbeat checks run concurrently, their transactions
+        # overlap and might end in shared lock errors. By aborting here
+        # we clean-up the state on each heartbeat call instead of once at the
+        # end of the request. See bug Kinto/kinto#804
         transaction.abort()
```
Natim: Any idea why we abort rather than committing?

Natim:
r+

kinto/core/views/heartbeat.py
leplatrem: The commit would have the same result, except that something would be left in the database between two calls.

Would you prefer ?

Natim: Yes i guess it makes sense to make sure the database is writable.

PR #816, opened 2016-09-13
Next-Page with filtering on a not mandatory field raises a KeyError.

Natim:
Reproductible easily with the Kinto-Admin. Let me write a test for it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Protocol is now at version **1.10**. See `API changelog`_.
 - Fix heartbeat transaction locks with PostgreSQL backends (fixes #804)
 - Fix crash with PostgreSQL storage backend when filtering with integer on
   a missing field (fixes #813)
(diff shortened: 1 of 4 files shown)


Natim:
@leplatrem I would like to talk a bit with you to make sure we are adopting the correct strategy here.

n1k0:
LGTM :)

leplatrem:
The fix looks good in apparence indeed. I'm ok to merge.

We could go further and try to setup a test where this would have an impact. For example, make sure that no record is skipped not duplicated when going through pagination when the last record of a page has not the field used in the filter or sort...

Natim:
your intuition was correct...

leplatrem:
What's the status of this now that #1116 was merged?

Natim:
Let me add the tests in a new PR so that we can see if we still have the issue.

Natim:
Superseeded by #1253

PR #821, opened 2016-09-15
Allow parent_id pattern matching in Storage.get_all.

Natim:
r? @leplatrem 

- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 **New features**
 
-- Allow ``parent_id`` pattern matching in ``Storage.get_all``. (#821)
+- The storage backend now allows ``parent_id`` pattern matching in ``kinto.core.storage.get_all``. (#821)
 
 
 4.2.0 (2016-09-15)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,9 @@ This document describes changes between each past release.
 4.3.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
+**New features**
+
+- Allow ``parent_id`` pattern matching in ``Storage.get_all``. (#821)
 
 
 4.2.0 (2016-09-15)
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -225,7 +225,7 @@ class Storage(MemoryBasedStorage):
 
     def _get_collections_by_parent_id(self, parent_id, store):
         if parent_id:
-            parent_id_match = re.compile(parent_id.replace('*', '.*'))
+            parent_id_match = re.compile("^%s$" % parent_id.replace('*', '.*'))
             return {pid: collections
                     for pid, collections in store.items()
                     if parent_id_match.match(pid)}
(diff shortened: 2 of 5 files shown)


Natim:
Apparently there are a lot of edge case to support for that.

CHANGELOG.rst
```diff
@@ -6,7 +6,9 @@ This document describes changes between each past release.
 4.3.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
+**New features**
+
+- Allow ``parent_id`` pattern matching in ``Storage.get_all``. (#821)
```
leplatrem: nit: The storage backend now allows [...] in ``kinto.core.storage.get_all()``.

tests/core/test_storage.py
```diff
@@ -342,6 +342,17 @@ def test_delete_raise_when_unknown(self):
...
+        self.create_record(parent_id='abc', collection_id='c')
+        record = self.create_record(parent_id='abc', collection_id='c')
+        self.storage.delete(object_id=record['id'], parent_id='abc', collection_id='c')
+        self.create_record(parent_id='efg', collection_id='c')
+
+        records, total_records = self.storage.get_all(parent_id='ab*', collection_id='c',
+                                                      include_deleted=True)
+        self.assertEquals(len(records), 2)
+        self.assertEquals(total_records, 1)
+
```
leplatrem: I would suggest to add a test without parent matching to make sure that we kept the strict comparison behaviour when two parent_ids are similar (eg. `parent_id: xabcx` should not be returned with `parent_id=abc`)

By the way, maybe we should have it for `delete_all()` also...

kinto/core/storage/memory.py
```diff
@@ -230,11 +230,36 @@ def get_all(self, collection_id, parent_id, filters=None, sorting=None,
...
+        for pid, collections in by_parent_id.items():
+            if collection_id is not None:
+                collections = {collection_id: collections[collection_id]}
+            for collection, colrecords in collections.items():
+                for r in colrecords.values():
+                    records.append(r)
+
+        records, count = self.extract_record_set(records,
+                                                 filters, None,
+                                                 id_field, deleted_field)
```
leplatrem: nit: since this method has many parameters, I suggest that we provide the kwargs names

kinto/core/storage/memory.py
```diff
@@ -230,11 +230,36 @@ def get_all(self, collection_id, parent_id, filters=None, sorting=None,
...
+            by_parent_id = {pid: collections
+                            for pid, collections in self._cemetery.items()
+                            if parent_id_match.match(pid)}
+
+            for pid, collections in by_parent_id.items():
+                if collection_id is not None:
+                    collections = {collection_id: collections[collection_id]}
+                for collection, colrecords in collections.items():
+                    for r in colrecords.values():
+                        deleted.append(r)
```
leplatrem: nit: those two blocks of code are very similar. Maybe we could reuse a small helper function shared with `delete_all()` ?

tests/core/test_storage.py
Natim: You are right, this tests fails with Postgresql :(

kinto/core/storage/memory.py
Natim: It doesn't really improve readability because parameters have got the same name as the variable that contains the value.

`(records=records, filters=filters, sorting=None, id_field=id_field, deleted_field=deleted_field)`

PR #827, opened 2016-09-21
Fix creation consistency (ref #792)

### LINKED ISSUE #792: Consistency on creation ? 403 versus 412
If we try to create an object that was already created by someone else, the server returns `403 Unauthorized`.

If we add the header `If-None-Match: *`, it also returns `403` even I have the permission to create new objects (and not `412 Precondition failed`), which can be confusing.
In the documentation we don't make those corner cases very clear.

Is that consistent by the way ? Knowing that we want to prevent Alice to know what objects ids Bob has created if she has no permission to read them.

leplatrem:
Fixes #792 

- [x] Add failing test
- [x] Fix tests.
- [x] Add a changelog entry.

### DIFF
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -221,6 +221,9 @@ class RouteFactory(object):
         method = request.method.lower()
         required_permission = self.method_permissions.get(method)
 
+        # For create permission, the object id is the plural endpoint
+        collection_path = service.collection_path.format(**request.matchdict)
+
         # In the case of a "PUT", check if the targetted record already
         # exists, return "write" if it does, "create" otherwise.
         if request.method.lower() == "put":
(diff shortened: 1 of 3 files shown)


leplatrem:
@magopian r?

kinto/core/authorization.py
```diff
@@ -221,6 +221,9 @@ def _find_required_permission(self, request, service):
         method = request.method.lower()
         required_permission = self.method_permissions.get(method)
 
+        # For create permission, the object id is the plural endpoint
```
magopian: Do you like ending comments with a `.`?

I'm not sure why, but this comment sounds a bit confusing to me. Maybe something like
`The create permission object id is the one for the plural endpoint.`?

kinto/core/authorization.py
```diff
@@ -233,10 +236,15 @@ def _find_required_permission(self, request, service):
...
                 # The record does not exist, the permission to create on
                 # the related collection is required.
-                permission_object_id = service.collection_path.format(
-                    **request.matchdict)
+                permission_object_id = collection_path
                 required_permission = "create"
             else:
-                required_permission = "write"
+                # If we are doing a safe creation, we consider that user demands
+                # a create permission (See Kinto/kinto#792).
```
magopian: `For safe creations, the user needs a create permission (See Kinto/kinto#792).`

tests/test_views_objects_permissions.py
```diff
@@ -183,3 +183,66 @@ def test_record_permissions_are_modified_by_patch(self):
...
+                              MINIMALIST_GROUP,
+                              headers=self.alice_headers)
+
+    def test_cannot_read_others_objects_if_only_allowed_to_create(self):
+        self.app.get('/buckets/create/groups', headers=self.bob_headers, status=403)
+        self.app.get('/buckets/create/groups/child', headers=self.bob_headers, status=403)
+
+    def test_safe_creation_with_put_returns_412_if_allowed_to_create(self):
+        headers = self.bob_headers.copy()
+        headers.update({'If-None-Match': '*'})
```
magopian: do you want to save those headers on the class so you can simply use `self.bob_if_none_match_headers`?

Issue #350, opened 2015-12-18
No inheritance for permissions forced in settings


leplatrem:
For example, if I give write permissions in settings

```ini
kinto.bucket_write_principals = basicauth:c6c27f0c7297ba7d4abd2a70c8a2cb88
```

This user has permission to write, but not to read, whereas `write` implies `read` when permissions are specified throught the JSON API.

magopian:
Since i updated from 11.x to 12.x, it seems I can't even write with the account that created the collection (to add a `read` permission explicitely for example, as a temporary workaround)

jesse-thompson:
Hi, I was looking to contribute. This would be my first time contributing to a project, and after looking around I'm not sure if this issue has been fixed or not. When going through the "getting started" steps, I'm still only seeing bob as having write permission.
Looking through the code, I'm leaning towards the applicable file being kinto -> authorization.py and the applicable code block being the PERMISSIONS_INHERITANCE_TREE at line 26.
I only saw bucket_write_principals in kinto -> views -> permissions.py, but only within a comment. I'm not really sure where to go with that.
Am I way off or should I dive in here?

jesse-thompson:
So I was looking [here](https://github.com/Kinto/kinto/blob/456d1cfe4c4964802ea5816af18196734f1a7365/kinto/authorization.py#L26-L56) and I noticed `read` wasn't included in all instanced where `write` is. I'm not saying this is the cause, but I was curious about why that is.

Issue #110, opened 2015-07-05
ETag is only returned on GET


magopian:
Shouldn't it also be returned on other verbs, like POST/PATCH/PUT...?

leplatrem:
I also think so.

Since the protocol implementation is done in Cliquet, let's track this there https://github.com/mozilla-services/cliquet/issues/352

I leave the issue opened here, we'll close it when fixed upstream :)

ametaireau:
First-hand, it seems weird to have these `Etags` returned on something else than `GET`s, because it can be understood that it applies to the resource that was queried. For intance, if you do a `POST` on `/records` and that an `Etag` is returned. Is it supposed to be matched against `/records` or `/records/{id}`.

At least, we should be sure to document this and make sure it doesn't prevent the clients from working like they should.

leplatrem:
Done, available in 1.4.0

PR #828, opened 2016-09-21
Parent metadata with child creation (fixes #803, #101)

### LINKED ISSUE #101: Finer grained permissions
Today, when we give the ``collection:write`` permission it implies the right to share this collection to anyone.

When you are logging in some application, you might want to refrain it to share your data.

A good way to do that is to explicitly ask for permission using scopes.

My proposal is to not give any permission with the ``kinto`` scope but to only allow permission to what defines in the scope:

 - todo:tasks:write
 - todo:tasks:permission
 - todo:tasks:schema

### LINKED ISSUE #803: record:create permissions needs to access the schema.
I had a quick chat with @enguerran who is planning to use Kinto with a similar use case than the formbuilder:

 - People can create collection with an associate schema in a bucket:
     - bucket_permissions: collection_create: Authenticated
     - collection_permissions: record:create: Authenticated

The problem if you do that is that people that needs to create record in the collection cannot actually access the schema.

However if you add the read permission on the collection, they can read the schema as well as all records in the collection (their and other's people onces.)

The use case there is to have the admin to be able to read all records and users to be able to see only their own records for a given collection.

We also want to be able to list the collection the user can administrate. It makes it impossible if we give the read permission on the collection.

A quick fix would be to allow to read the schema or the collection data if we give the `record:create` permission.

leplatrem:
Fixes #803, Fixes #101

- [x] Add failing tests
- [x] Fix tests.
- [x] Update documentation
- [x] Bump protocol
- [x] Add a changelog entry.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -14,8 +14,10 @@ Changelog
 1.11 (unreleased)
 '''''''''''''''''
 
-- Parent metadata is now readable if children creation is allowed
-- Now returns a ``412`` instead of a ``403`` if the ``If-None-Match: *`` header is
+- Parent attributes are now readable if children creation is allowed
+- Return an empty list on the plural endpoint instead of |status-403| if the ``create``
+  permission is allowed
+- Now returns a |status-412| instead of a |status-403| if the ``If-None-Match: *`` header is
   provided and the ``create`` permission is allowed
 
 1.10 (2016-09-15)
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -11,11 +11,17 @@ API
 Changelog
 ---------
 
+1.11 (unreleased)
+'''''''''''''''''
+
+- Parent metadata is now readable if children creation is allowed
+- Now returns a ``412`` instead of a ``403`` if the ``If-None-Match: *`` header is
+  provided and the ``create`` permission is allowed
+
 1.10 (2016-09-15)
 '''''''''''''''''
 
 - Add substring query to filtering on plural endpoints (e.g ``?like_person=Tim``)
-
 
 1.9 (2016-08-17)
 ''''''''''''''''
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -84,18 +84,18 @@ class AuthorizationPolicy(object):
         # If not allowed on this collection, but some records are shared with
         # the current user, then authorize.
         # The ShareableResource class will take care of the filtering.
-        is_list_operation = (context.on_collection and
-                             not permission.endswith('create'))
+        is_list_operation = (context.on_collection and not permission.endswith('create'))
         if not allowed and is_list_operation:
             shared = context.fetch_shared_records(permission,
                                                   principals,
                                                   self.get_bound_permissions)
             # If allowed to create this kind of object on parent, then allow to obtain the list.
             if len(bound_perms) > 0:
-                parent_uri = bound_perms[0][0]  # Bounds perms are ordered.
+                # Here we consider that parent URI is one path level above.
+                parent_uri = '/'.join(object_id.split('/')[:-1])
                 parent_create_perm = [(parent_uri, create_permission)]
             else:
-                parent_create_perm = [('', 'create')]
+                parent_create_perm = [('', 'create')]  # Root object.
             allowed_to_create = context.check_permission(principals, parent_create_perm)
 
             allowed = shared or allowed_to_create
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -69,17 +69,19 @@ class AuthorizationPolicy(object):
         if permission == DYNAMIC:
             permission = context.required_permission
 
+        create_permission = '%s:create' % context.resource_name
         if permission == 'create':
-            permission = '%s:%s' % (context.resource_name, permission)
+            permission = create_permission
+
+        object_id = context.permission_object_id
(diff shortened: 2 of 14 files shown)


leplatrem:
I applied the same logics to bucket metadata (`collection:create` and `group:create` gives access to (parent) bucket metadata).

Maybe we shouldn't and keep it for record+collection metadata only?

What do you think @enguerran / @magopian ?

magopian:
I don't see any drawbacks in having the same behavior. Do you think it could be a security issue or something?

docs/api/index.rst
```diff
@@ -11,12 +11,18 @@ API
 Changelog
 ---------
 
+1.11 (unreleased)
+'''''''''''''''''
+
+- Parent metadata is now readable if children creation is allowed
+- Now returns a ``412`` instead of a ``403`` if the ``If-None-Match: *`` header is
+  provided and the ``create`` permission is allowed
```
magopian: Do you want to add a note about the fact that `GET`ing the list of records will return an empty list (and not a 403) if the user doesn't have the read permission on the collection, and has not created any records yet? (as discussed in the issue #803)

tests/test_views_objects_permissions.py
```diff
@@ -236,3 +235,41 @@ def test_safe_creation_with_post_returns_403_if_only_allowed_to_read(self):
...
+                     headers=get_user_headers('jean:paul'),
+                     status=403)
+        self.app.get('/buckets/beer/collections/barley',
+                     headers=get_user_headers('mahmud:hatim'),
+                     status=403)
+
+    def test_list_can_be_obtained_if_allowed_to_create(self):
+        resp = self.app.get('/buckets/beer/collections', headers=self.bob_headers)
+        self.assertEqual(len(resp.json['data']), 1)
+        self.assertEqual(resp.json['data'][0]['id'], 'barley')
```
magopian: I thought one could only see the list of his own created entries, or is that only for records? Is it expected that Bob can see this collection, while he only has the "create" permission on collections, and the 'barley' collection has been created by Alice?

docs/api/index.rst
leplatrem: I believe `test_list_can_be_obtained_if_allowed_to_create` and `test_list_is_denied_if_not_allowed_to_create` cover this (?)

magopian: Sure, the tests and the code are there, but how was wondering if you wanted to add a note to the changelog about that?

tests/test_views_objects_permissions.py
leplatrem: the `barley` collection was created by Bob (and Alice is allowed to create records). But you're right, the current version of the patch is not correct. I added a failing test that shows that users should only see their own records in this particular case.

leplatrem:
> Do you think it could be a security issue or something? 

I wouldn't say a security issue, but it makes the impact of the change bigger.

enguerran:
I cannot take a decision on this as I am not yet aware enough of kinto permissions system. I do discussed with @Natim. But I cannot understand how my needs are different from formbuilder's.

Extract from https://github.com/Kinto/formbuilder/blob/master/formbuilder/actions/server.js:

- bucket: `{ "collection:create": ["System.Authenticated"], "write": [] }`
- collection (a form): `{ "record:create": ["System.Authenticated"], "read": [userId] }`

The extra need is
> We also want to be able to list the collection the user can administrate. It makes it impossible if we give the read permission on the collection.

In formbuilder, access are controlled by "encoded" URL. If you have the admin token, you'll have access to the collection's administration. If you have the user token, you'll have access to the form and can only create a record. As far as I could understand formbuilder's code.

If I create a collection of "url with admin token" in a specific user's default bucket, I guess an admin could have access to several forms' administration.

Again, I am just a beginner user of kinto and maybe I am wrong.

leplatrem:
This cannot be reviewed without going through each commit. I'm very confident in the code and the test suite though.

I changed and simplified a lot of code. Basically:
* I renamed many variables and method names in the authorization code. It is now more consistent with what it does.
* I added a level of keys in the inheritance tree of permissions. Before it used to have some concat and join strings for the keys.
* There is now a `read:attributes` permission in the inheritance tree of permissions. It is not exposed yet to the HTTP API but we could eventually. It gives the finer grain we need (#110) to allow reading attributes and not the whole list of records.
* I changed the way the authorization code relies on settings. This paves the way for #350 
* I removed the specific case we had for allowing listing buckets. It now behaves as other: if you can create buckets, you can receive an empty list on `/buckets` if none are accessible to you.

Don't hesitate to ping me for questions!

@magopian r?

kinto/core/authorization.py
```diff
@@ -91,7 +90,15 @@ def permits(self, context, principals, permission):
             shared = context.fetch_shared_records(permission,
                                                   principals,
                                                   self.get_bound_permissions)
-            allowed = shared is not None
+            # If allowed to create this kind of object on parent, then allow to obtain the list.
+            if len(bound_perms) > 0:
+                parent_uri = bound_perms[0][0]  # Bounds perms are ordered.
```
leplatrem: This takes the first object URI of the bound perms. 
But the first is not always the parent! (Ex. record → collection → bucket)
So I guess this is wrong for records, because it will check that the user has the permission `record:create` on `/buckets/test` which will never be the case.

leplatrem:
Final r+ my friends?

leplatrem:
\o/

PR #832, opened 2016-09-24
Statds improvements

leplatrem:
r? @Natim  / @magopian  :)

### DIFF
--- a/kinto/core/statsd.py
+++ b/kinto/core/statsd.py
@@ -14,10 +14,6 @@ from kinto.core import utils
 
 class Client(object):
     def __init__(self, host, port, prefix):
-        if statsd_module is None:
-            error_msg = "Please install Kinto with monitoring dependencies (e.g. statsd package)"
-            raise ConfigurationError(error_msg)
-
         self._client = statsd_module.StatsClient(host, port, prefix=prefix)
 
     def watch_execution_time(self, obj, prefix='', classname=None):
@@ -48,6 +44,13 @@ def statsd_count(request, count_key):
 
 
 def load_from_config(config):
+    # If this is called, it means that a ``statsd_url`` was specified in settings.
+    # (see ``kinto.core.initialization``)
+    # Raise a proper error if the ``statsd`` module is not installed.
+    if statsd_module is None:
+        error_msg = "Please install Kinto with monitoring dependencies (e.g. statsd package)"
+        raise ConfigurationError(error_msg)
+
     settings = config.get_settings()
     uri = settings['statsd_url']
     uri = urlparse.urlparse(uri)
--- a/kinto/core/statsd.py
+++ b/kinto/core/statsd.py
@@ -6,6 +6,7 @@ try:
 except ImportError:  # pragma: no cover
     statsd_module = None
 
+from pyramid.exceptions import ConfigurationError
 from six.moves.urllib import parse as urlparse
 
 from kinto.core import utils
@@ -13,6 +14,10 @@ from kinto.core import utils
 
 class Client(object):
     def __init__(self, host, port, prefix):
+        if statsd_module is None:
+            error_msg = "Please install Kinto with monitoring dependencies (e.g. statsd package)"
+            raise ConfigurationError(error_msg)
+
         self._client = statsd_module.StatsClient(host, port, prefix=prefix)
 
     def watch_execution_time(self, obj, prefix='', classname=None):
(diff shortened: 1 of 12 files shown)


kinto/plugins/history/__init__.py
```diff
@@ -12,7 +12,14 @@ def includeme(config):
     # Activate end-points.
     config.scan('kinto.plugins.history.views')
 
+    # If StatsD is enabled, monitor execution time of listener.
+    if config.registry.statsd:
+        key = 'listeners.history'
```
leplatrem: Maybe it should be `plugins.history` here instead of `listeners.X`, since we use those for listeners configured via `.ini` files?

kinto/core/statsd.py
```diff
@@ -6,13 +6,18 @@
...
 
+from pyramid.exceptions import ConfigurationError
 from six.moves.urllib import parse as urlparse
 
 from kinto.core import utils
 
 
 class Client(object):
     def __init__(self, host, port, prefix):
+        if statsd_module is None:
```
magopian: Should this only be raised if statsd is configured, as explained in [the docs above](https://github.com/Kinto/kinto/pull/832/files#diff-de0a9b67ffe2874a076b5fdb15bad484R24)?

kinto/plugins/history/__init__.py
```diff
@@ -12,7 +12,14 @@ def includeme(config):
     # Activate end-points.
     config.scan('kinto.plugins.history.views')
 
+    # If StatsD is enabled, monitor execution time of listener.
+    if config.registry.statsd:
+        key = 'plugins.history'
+        listener = config.registry.statsd.timer(key)(on_resource_changed)
+    else:
+        listener = on_resource_changed
```
magopian: How do you like the "simple assignment before if" instead? Eg

```python
listener = on_resource_changed
if config.registry.statsd:
    key = 'plugins.history'
    listener = config.registry.statsd.timer(key)(on_resource_changed)
```

I know some people prefer this, some people despise it ;)

tests/core/test_statsd.py
```diff
@@ -18,7 +19,20 @@ def _private_method(self):
...
+        self.previous = statsd.statsd_module
+        statsd.statsd_module = None
+
+    def tearDown(self):
+        statsd.statsd_module = self.previous
+
+    def test_client_instantiation_raises_properly(self):
+        with self.assertRaises(ConfigurationError):
+            statsd.Client('localhost', 1234, 'prefix')
+
```
magopian: Maybe a test that makes sure that the client instantiation doesn't raise if the statsd is not configured and not installed, and neither if statsd is configured and installed?

magopian:
I mean "a minor change is needed to either the docs or to the code and tests" ;)

kinto/core/statsd.py
leplatrem: Yes, but this client is instantiated only if `statsd_url` is specified in the settings (cf. `initialization.py`)

magopian: oh, right, it makes sense now ;)

magopian:
r+

tests/core/test_initialization.py
```diff
@@ -273,83 +273,76 @@ def setUp(self):
...
-        mocked.assert_called_with('host', 8080, 'kinto.core')
+        self.mocked.assert_called_with(self.config)
+        # See `tests/core/test_statsd.py` for instantiation tests.
 
-    @mock.patch('kinto.core.statsd.Client')
-    def test_statsd_is_expose_in_the_registry_if_url_is_set(self, mocked):
+    def test_statsd_is_expose_in_the_registry_if_url_is_set(self):
         initialization.setup_statsd(self.config)
-        self.assertEqual(self.config.registry.statsd, mocked.return_value)
+        self.assertEqual(self.config.registry.statsd, self.mocked.return_value)
```
magopian: i'm not saying you should use that (it depends on wether you want conciseness or code coherence), but just so you know, with pytest you can use simple `assert`s

```assert self.config.registry.statsd == self.mocked.return_value```

File: tests/core/test_initialization.py [hunk 6]
```diff
@@ -273,83 +273,76 @@ def setUp(self):
...
         c.watch_execution_time.assert_any_call(None, prefix='authentication')
 
-    @mock.patch('kinto.core.statsd.Client')
-    def test_statsd_counts_nothing_on_anonymous_requests(self, mocked):
+    def test_statsd_counts_nothing_on_anonymous_requests(self):
         kinto.core.initialize(self.config, '0.0.1', 'project_name')
         app = webtest.TestApp(self.config.make_wsgi_app())
         app.get('/')
-        self.assertFalse(mocked.count.called)
+        self.assertFalse(self.mocked.count.called)
```
magopian: same here, you could do something like `assert not self.mock.count.called`

Issue #788, opened 2016-08-23
Missing doc for manual postgresql install


leplatrem:
```
<ochameau> natim: quick followup about docs, there is nothing about sqlalchemy, zope.sqlalchemy, psycopg2 python deps being required when using psql. without doing all the related pip install, you end up with various exception while running kinto migrate
<leplatrem> ochameau: oh, yeah we made some stuff automatically
<natim> Nothing about pip install kinto[postgresql] ?
<leplatrem> and it may not be explicit when done manually
<leplatrem> please open an issue, it makes sense!
<leplatrem> when using `kinto init` everything is installed automatically
<ochameau> I first did kinto init without any config and then relaunched it with a modified config now using psql instead of memory...
<ochameau> I first gave a try to kinto in memmory
<ochameau> nothing about kinto[postgresql] over here: http://kinto.readthedocs.io/en/stable/configuration/production.html#postgresql-install
<natim> We should definitely improve that, would you have a minute to do a PR on the doc?
<ochameau> this page, doesn't mention psql 9.4+, which fails with cryptic jsonb exceptions - https://github.com/Kinto/kinto/wiki/How-to-run-a-PostgreSQL-server%3F
```

leplatrem:
Note: also add how to set your database to `UTC`

zeddmaxx:
@leplatrem I would like to work on this, could you please elaborate on what exactly needs to be done?

leplatrem:
Ok, so here is some insights about this cryptic ticket :)

* Our target are the users who have played with the local in-memory backend, configured their plugins and stuff, and now want to go to production.
* I don't think they'll be happy to restart configuration from scratch with `kinto init` (*but I may be wrong*)
* Some Python packages will be missing, because we install those only during `kinto init`
* I don't think we should go into the details of installing PostgreSQL for a particular distro/version of Linux/Ubuntu, we should link to external official docs, which will be maintained and up-to-date for sure. Our wiki page may also be improved but we don't want it in the docs, a link at most.
* We want to give every instructions that are specific to Kinto, the following are enough I believe:
   * PostgreSQL 9.4+ is required (link to external docs)
   * Install python dependencies with `pip install -Ue "kinto[postgresql]"`
   * create a specific database like `mydatabase` with UTF8 encoding
   * set your database to `UTC` timezone (`ALTER DATABASE mydatabase SET timezone TO 'UTC';`)
   * change your backend settings to `kinto.core.storage.postgresql` (see settings docs)
[...]

PR #834, opened 2016-09-24
Postgres storage: remove id/last_modified from data field (fixes #762)

### LINKED ISSUE #762: Postgresql storage: id (and last_modified) are stored in the data column
We could «save» some storage space by avoiding this. It is rather trivial to do: strip record of those fields before storing it as `data`.

Current insert query is:
```sql
	        WITH delete_potential_tombstone AS (
	            DELETE FROM deleted
	             WHERE id = '861756c7-9405-4819-a2cb-336606d3540e'
	               AND parent_id = '1234'
	               AND collection_id = 'test'
	        )
	        INSERT INTO records (id, parent_id, collection_id, data, last_modified)
	        VALUES ('861756c7-9405-4819-a2cb-336606d3540e', '1234',
	                'test', ('{"code":{"sub":"c"},"id":"861756c7-9405-4819-a2cb-336606d3540e"}')::JSONB,
	                from_epoch(NULL))
	        RETURNING id, as_epoch(last_modified) AS last_modified;

```

gabisurita:
Fixes #762 

- [x] Add a changelog entry.

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -229,10 +229,10 @@ class Storage(StorageBase):
                 pass
         else:
             record[id_field] = id_generator()
-        
+
         # Remove redundancy in data field
         query_record = record.copy()
-        del query_record[id_field] 
+        del query_record[id_field]
         if modified_field in record:
             del query_record[modified_field]
 
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -232,9 +232,8 @@ class Storage(StorageBase):
 
         # Remove redundancy in data field
         query_record = record.copy()
-        del query_record[id_field]
-        if modified_field in record:
-            del query_record[modified_field]
+        query_record.pop(id_field, None)
+        query_record.pop(modified_field, None)
 
         query = """
         WITH delete_potential_tombstone AS (
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -290,6 +290,12 @@ class Storage(StorageBase):
                id_field=DEFAULT_ID_FIELD,
                modified_field=DEFAULT_MODIFIED_FIELD,
                auth=None):
+
+        # Remove redundancy in data field
+        query_record = record.copy()
+        query_record.pop(id_field, None)
+        query_record.pop(modified_field, None)
+
         query_create = """
         WITH delete_potential_tombstone AS (
             DELETE FROM deleted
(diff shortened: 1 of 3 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -230,6 +230,12 @@ def create(self, collection_id, parent_id, record, id_generator=None,
         else:
             record[id_field] = id_generator()
 
+        # Remove redundancy in data field
+        query_record = record.copy()
+        del query_record[id_field]
+        if modified_field in record:
+            del query_record[modified_field]
```

You could even make it shorter like this:
```
query_record.pop(id_field, None)
query_record.pop(modified_field, None)
```

leplatrem:
Thanks for this contribution!

We should also do it for `update()`.

Since this applies to the postgresql backend internals, it would exceptionnally acceptable to merge it without tests. But maybe it would be relevant to have a test in `storage/testing.py` that asserts that the specified `record` object is not altered (that's why you added the line with `.copy()` I suppose). What do you think ?

gabisurita:
Removing the id field already leads to an error on the insertion, but the removing the timestamp doesn't. I think including a test for it would be a good idea. I'll work on it.

leplatrem:
The current code looks great and seems to do the job! Adding a test would make it perfect :)
(*do not hesitate to ask for help if you're stuck*)

gabisurita:
Sorry for the delay, I was having some troubles in get the storage tests running locally. Some tests were failing due to the database timezone. This probably should be included on the backend documentation, but I think that's already covered by issue #788. 

Leaving the last_modified field off the record already produce errors on several tests, so I think another test for this is not needed.
`
Results (159.86s):
    1386 passed
       6 failed
         - /home/gabi/outreachy/kinto/kinto/core/storage/testing.py:567: AssertionError: 1475004498137L not less than 1475004498124L
         - /home/gabi/outreachy/kinto/kinto/core/storage/testing.py:548: AssertionError: 1475004498292L != 1448881675541
         - /home/gabi/outreachy/kinto/kinto/core/storage/testing.py:498: AssertionError: 100 != 200
         - /home/gabi/outreachy/kinto/kinto/core/storage/testing.py:650: KeyError: 'last_modified'
         - /home/gabi/outreachy/kinto/kinto/core/storage/testing.py:629: AssertionError: 1475004507708L not less than 1475004507701L
         - /home/gabi/outreachy/kinto/kinto/core/storage/testing.py:610: KeyError: 'last_modified'
`

leplatrem:
Well, it's always better to be able to run the tests locally... 

I pulled the branch, and the whole test suite is green indeed. And you're right there is a test called `test_create_copies_the_record_before_modifying_it` in the storage tests.

So I think we're good to merge!

Thanks for the good work!

PR #835, opened 2016-09-26
Location of version.json is now configurable (fixes #830)

### LINKED ISSUE #830: Make the location of version.json configurable
Currently @phrawzty had to put the `version.json` at the root location of the container.

If the location would be a setting we could specify via env variable in the dockerfile for example

leplatrem:
Fixes #830

- [x] Add tests.
- [x] Add a changelog entry.

r? @magopian

### DIFF
--- a/tests/core/test_views_version.py
+++ b/tests/core/test_views_version.py
@@ -9,34 +9,32 @@ from .support import BaseWebTest
 
 
 class VersionViewTest(BaseWebTest, unittest.TestCase):
-    def tearDown(self):
-        try:
-            os.remove('version.json')  # Clean-up file at default location.
-        except OSError:
-            pass
+    def setUp(self):
         try:
             delattr(version_view, '__json__')  # Clean-up memoization.
         except AttributeError:
             pass
 
     def test_return_the_version_file_in_current_folder_if_present(self):
-        with open('version.json', 'w') as f:
-            json.dump({'version': '0.8.1'}, f)
-
-        response = self.app.get('/__version__')
-        assert 'version' in response.json
+        content = {'version': '0.8.1'}
+        fake_file = mock.mock_open(read_data=json.dumps(content))
+        with mock.patch('os.path.exists'):
+            with mock.patch('kinto.core.views.version.open', fake_file, create=True):
+                response = self.app.get('/__version__')
+                assert response.json == content
 
     def test_return_a_404_if_version_file_if_not_present(self):
         self.app.get('/__version__', status=404)
 
     def test_return_the_version_file_specified_in_setting_if_present(self):
         custom_path = tempfile.mktemp()
+        content = {'foo': 'lala'}
         with open(custom_path, 'w') as f:
-            json.dump({'foo': 'lala'}, f)
+            json.dump(content, f)
         self.addCleanup(lambda: os.remove(custom_path))
 
         with mock.patch.dict(self.app.app.registry.settings,
                              [('version_json_path', custom_path)]):
             response = self.app.get('/__version__')
 
-        assert 'foo' in response.json
+        assert response.json == content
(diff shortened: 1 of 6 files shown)


tests/core/test_views_version.py
```diff
@@ -1,30 +1,42 @@
...
+        except AttributeError:
+            pass
+
+    def test_return_the_version_file_in_current_folder_if_present(self):
+        with open('version.json', 'w') as f:
+            json.dump({'version': '0.8.1'}, f)
 
-        assert response.json == version.VERSION_JSON
+        response = self.app.get('/__version__')
+        assert 'version' in response.json
```
magopian: Maybe test the full response?
```assert response.json == "{'version': '0.8.1'}"```
This way, if we have some other code in the future that overrides the version, this test will fail ;)

File: tests/core/test_views_version.py [hunk 2]
```diff
@@ -1,30 +1,42 @@
...
+            os.remove('version.json')  # Clean-up file at default location.
+        except OSError:
+            pass
+        try:
+            delattr(version_view, '__json__')  # Clean-up memoization.
+        except AttributeError:
+            pass
+
+    def test_return_the_version_file_in_current_folder_if_present(self):
+        with open('version.json', 'w') as f:
```
magopian: I would highly recommend avoiding writing files (and even more if the files are in "non test locations") during a test. This means that the test will overwrite the local version.json file (and the teardown will remove it anyway).

How about using mock instead?

tests/core/test_views_version.py
```diff
@@ -1,30 +1,42 @@
...
-    def test_return_the_version_file_if_present(self):
-        version.VERSION_JSON = {
-            "commit": "ce480821debc080d9cf5cb67e9a6ba71b2c3f57e",
-            "version": "0.8.1",
-            "name": "kinto-dist",
-            "source": "https://github.com/mozilla-services/kinto-dist"
-        }
-        response = self.app.get('/__version__')
+        try:
+            os.remove('version.json')  # Clean-up file at default location.
```
magopian: I would rather make use of `self.addCleanup` when needed (as you did with the second test), because running this will remove any local version.json file.

tests/core/test_views_version.py
```diff
@@ -1,30 +1,42 @@
...
+        except OSError:
+            pass
+        try:
+            delattr(version_view, '__json__')  # Clean-up memoization.
```
magopian: It would make sense to have this in the `setUp` too (in case some other parts of the code are exercised before, and thus the memoization already happened).

tests/core/test_views_version.py
```diff
@@ -1,30 +1,42 @@
...
+        custom_path = tempfile.mktemp()
+        with open(custom_path, 'w') as f:
+            json.dump({'foo': 'lala'}, f)
+        self.addCleanup(lambda: os.remove(custom_path))
+
+        with mock.patch.dict(self.app.app.registry.settings,
+                             [('version_json_path', custom_path)]):
+            response = self.app.get('/__version__')
+
+        assert 'foo' in response.json
```
magopian: Maybe test the full response?
```assert response.json == "{'foo': 'lala'}"```

File: tests/core/test_views_version.py [hunk 6]
```diff
@@ -9,34 +9,32 @@
 
 
 class VersionViewTest(BaseWebTest, unittest.TestCase):
-    def tearDown(self):
-        try:
-            os.remove('version.json')  # Clean-up file at default location.
-        except OSError:
-            pass
+    def setUp(self):
```
magopian: But we also need it in the `tearDown` don't we? So the memoization doesn't stay around for following tests? Or are you going to call `self.addCleanup` or something?

magopian:
i believe you adressed all the comments (hard to say with the new "review" feature of github, code changes don't hide the comments anymore :(

r+

PR #837, opened 2016-09-27
Hide permissions if read permission only (fixes #123)

### LINKED ISSUE #123: We should consider hidding resources permissions when we only have the read permission
This is related to https://github.com/mozilla-services/kinto/pull/88

When we have the ``read`` permission to list existing buckets, we can ask for a bucket and in that case we can list its permissions. I think it is only something that somebody that can modify this permission should be able to see.

leplatrem:
Fixes #123

- [x] Add documentation.
- [x] Add end view tests.
- [x] Decide whether it should be empty or removed from response
- [x] Bump protocol version
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ This document describes changes between each past release.
 - Fix error response consistency with safe creations if the ``create`` permission
   is granted (fixes #792). The server now returns a ``412`` instead of a ``403`` if
   the ``If-None-Match: *`` header is provided and the ``create`` permission is granted.
-- The ``permissions`` is now empty in the response if the user has not the permission
+- The ``permissions`` attribute is now empty in the response if the user has not the permission
   to write on the object (fixes #123)
 
 **New features**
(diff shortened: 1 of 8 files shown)


leplatrem:
@n1k0+@magopian r?

CHANGELOG.rst
```diff
@@ -11,6 +11,8 @@ This document describes changes between each past release.
 - Fix error response consistency with safe creations if the ``create`` permission
   is granted (fixes #792). The server now returns a ``412`` instead of a ``403`` if
   the ``If-None-Match: *`` header is provided and the ``create`` permission is granted.
+- The ``permissions`` is now empty in the response if the user has not the permission
```
n1k0: The `permissions` *object*

tests/core/resource/test_object_permissions.py
```diff
@@ -77,8 +84,8 @@ def setUp(self):
         self.record_uri = '/articles/%s' % record_id
         self.permission.add_principal_to_ace(self.record_uri,
                                              'read',
-                                             'fxa:user')
-        self.resource.model.current_principal = 'basicauth:userid'
+                                             'account:readonly')
+        # self.resource.model.current_principal = 'basicauth:userid'
```
n1k0: nit: leftover?

docs/api/index.rst
```diff
@@ -11,6 +11,13 @@ API
 Changelog
 ---------
 
+1.11 (unreleased)
+'''''''''''''''''
+
+- The ``permissions`` are now empty in the response if the user has not the
```
magopian: somehow this doesn't sound very good... maybe reword to "if the user does not have the permission to write"?

magopian:
all good it seems, :shipit: !

PR #839, opened 2016-09-28
Include tutorial for postgres backend configuration (fixes #788) 

### LINKED ISSUE #788: Missing doc for manual postgresql install

gabisurita:
Fixes #788 (maybe?)

- [ ] Add documentation.

This may not include all the information needed, but I think it's first step.

### DIFF
--- a/docs/configuration/production.rst
+++ b/docs/configuration/production.rst
@@ -19,17 +19,30 @@ Install and setup PostgreSQL
 
 (*requires PostgreSQL 9.4 or higher*).
 
-*Kinto* dependencies do not include *PostgreSQL* tooling and drivers by default.
+*Kinto* dependencies do not include *PostgreSQL* tooling and drivers by default, which should be installed and configured before proceeding to the next steps. More information is available at the `PostgreSQL Documentation <http://www.postgresql.org/docs>`_.
 
 
 PostgreSQL client
 -----------------
 
+Before installing the Python libraries for Postgres we need to first install the header files for the PostgreSQL database backend.
+
 On Debian / Ubuntu based systems::
 
     $ sudo apt-get install libpq-dev
 
+On RedHat / Fedora / Mint based systems::
+
+    $ yum install postgresql-devel
+
 On Mac OS X, `install a server or use port <http://superuser.com/questions/296873/install-libpq-dev-on-mac-os>`_.
+
+Install PostgreSQL Python Dependencies
+--------------------------------------
+
+Kinto PostgreSQL backend rely on several Python toolkits, which can be installed with a single *pip* command ::
+
+    $ sudo pip install -Ue kinto[postgresql]
 
 
 Run a PostgreSQL server
@@ -40,8 +53,27 @@ The instructions to run a local PostgreSQL database are out of scope here.
 A detailed guide is :github:`available on the Kinto Wiki <Kinto/kinto/wiki/How-to-run-a-PostgreSQL-server%3F>`.
 
 
+Database Setup
+--------------
+
+*Kinto* default database name is set to ``postgres`` on ``localhost:5432``. To change it refer to `Creating a configuration file`_.
+
+To create a specific database for *kinto*:
+
+.. code-block:: sql
+
+    CREATE DATABASE dbname WITH ENCODING UTF8;
+
+By default *Kinto* uses *UTC timezone*, so make sure that the assigned database for kinto is set to UTC with:
+
+.. code-block:: sql
+
+    ALTER DATABASE dbname SET TIMEZONE TO UTC;
+
 Privileges basics
 -----------------
+
+*Kinto* default user is set to ``postgres`` with password ``postgres``. To change that refer to `Creating a configuration file`_.
 
 In order to initialize the database tables and objects, the specified user must
 have some privileges. For example, to create a user from scratch:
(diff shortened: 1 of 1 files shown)


docs/configuration/production.rst
```diff
@@ -76,24 +108,34 @@ change the default privileges to allow reading the future tables:
...
+
+    kinto.storage_backend = kinto.core.storage.postgresql
+    kinto.storage_url = postgres://dbuser:dbpassword@localhost/dbname
+    kinto.cache_backend = kinto.core.cache.postgresql
+    kinto.cache_url = postgres://dbuser:dbpassword@localhost/dbname
+    kinto.permission_backend = kinto.core.permission.postgresql
+    kinto.permission_url = postgres://dbuser:dbpassword@localhost/dbname
+
+
+Initialization
```
leplatrem: `Creation of tables and indices` maybe ?

docs/configuration/production.rst
```diff
@@ -76,24 +108,34 @@ change the default privileges to allow reading the future tables:
...
+--------------
 
 The last step consists in creating the necessary tables and indices, run the ``migrate`` command:
 
-.. code-block :: bash
+.. code-block:: bash
 
     $ kinto --ini production.ini migrate
```
leplatrem: We could add a `..important::` paragraph here to mention that this command should be ran every time Kinto is upgraded.

docs/configuration/production.rst
```diff
@@ -40,9 +53,28 @@ The instructions to run a local PostgreSQL database are out of scope here.
 A detailed guide is :github:`available on the Kinto Wiki <Kinto/kinto/wiki/How-to-run-a-PostgreSQL-server%3F>`.
 
 
+Database Setup
+--------------
+
+*Kinto* default database name is set to ``postgres`` on ``localhost:5432``. To change it refer to `Creating a configuration file`_.
+
+To create a specific database for *kinto*:
```
leplatrem: nitpick: *for kinto* is superfluous here I think

docs/configuration/production.rst
```diff
@@ -19,18 +19,31 @@ Install and setup PostgreSQL
...
+On RedHat / Fedora / Mint based systems::
+
+    $ yum install postgresql-devel
+
 On Mac OS X, `install a server or use port <http://superuser.com/questions/296873/install-libpq-dev-on-mac-os>`_.
 
+Install PostgreSQL Python Dependencies
+--------------------------------------
+
+Kinto PostgreSQL backend rely on several Python toolkits, which can be installed with a single *pip* command ::
```
leplatrem: * typo: *backends*
* I would say *Python packages*

docs/configuration/production.rst
```diff
@@ -19,18 +19,31 @@ Install and setup PostgreSQL
...
+
+    $ sudo pip install -Ue kinto[postgresql]
```
leplatrem: Do not mention `sudo` here. Even though it may work, it installs the Python libraries globally on the system, which is not recomended.

Also, I think `-U` is enough.

docs/configuration/production.rst
```diff
@@ -40,9 +53,28 @@ The instructions to run a local PostgreSQL database are out of scope here.
 A detailed guide is :github:`available on the Kinto Wiki <Kinto/kinto/wiki/How-to-run-a-PostgreSQL-server%3F>`.
 
 
+Database Setup
```
leplatrem: nitpick: in other titles, only the first word is capitalized

docs/configuration/production.rst
```diff
@@ -76,24 +108,34 @@ change the default privileges to allow reading the future tables:
...
 
 
-Initialization
---------------
+Creating a configuration file
+-----------------------------
 
-Once a PostgreSQL is up and running somewhere, select the PostgreSQL option when
-running the ``init`` command:
+Once a PostgreSQL is up and running somewhere, you should edit your configuration file or create one with the ``init`` command and select the PostgreSQL option. The file is created by default as ``config\kinto.ini``, to specify another name use the ``--ini`` parameter.
```
leplatrem: * *or create a new one*
* ``config/kinto.ini`` (Unix path style)
* another filename

gabisurita:
I've made a few changes that I think cover all your comments.

One more thing, do you think we should mention some of the Python dependencies? Or at least cite the major dependencies, like sqlalchemy and psycopg2?

leplatrem:
👍 

Yes, you could mention the packages why not: *Kinto PostgreSQL backends rely on specific Python packages (like SQLAlchemy and psycopg2)*

leplatrem:
Nice, thanks a lot! Merging 🎉

PR #841, opened 2016-09-30
Filter out records that don't have a field when filtering on MAX with memory storage

### LINKED ISSUE #815: Disparity between memory and postgresql when filtering with missing values
Once #814 will be merged, there is another minor bug to tackle:

when filtering with `?max_field=10`, the memory backend returns records which don't have `field` defined, and postgresql does not.

magopian:
Fixes #815

This does a bit more than just dealing with the MAX use case: we should now be closer to what we do with the postgresql backend: if we're comparing to a number, we can't include any records that don't have the given filtered field. If we're comparing to a string, then records that don't have this field will just be treated as if they had an empty string for this field. See what we do for postgresql here: https://github.com/Kinto/kinto/blob/54572a2d9aa4a56fdf49cf463324dd2dee8260a3/kinto/core/storage/postgresql/__init__.py#L658

### DIFF
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -329,13 +329,16 @@ def apply_filters(records, filters):
             if f.operator in (COMPARISON.IN, COMPARISON.EXCLUDE):
                 right, left = left, right
             else:
-                # Python3 cannot compare None to other value.
                 if left is None:
-                    if f.operator in (COMPARISON.GT, COMPARISON.MIN):
+                    right_is_number = (
+                        isinstance(right, (int, float)) and
+                        right not in (True, False))
+                    if right_is_number:
+                        # Python3 cannot compare None to a number.
                         matches = False
                         continue
-                    elif f.operator in (COMPARISON.LT, COMPARISON.MAX):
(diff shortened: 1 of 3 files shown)


magopian:
@leplatrem r?

kinto/core/storage/memory.py
```diff
@@ -329,13 +329,14 @@ def apply_filters(records, filters):
             if f.operator in (COMPARISON.IN, COMPARISON.EXCLUDE):
                 right, left = left, right
             else:
-                # Python3 cannot compare None to other value.
                 if left is None:
-                    if f.operator in (COMPARISON.GT, COMPARISON.MIN):
+                    if isinstance(right, (int, float)) and \
+                       right not in (True, False):
```
leplatrem: nitpicks: use intermediary variables to avoid folding if conditions (ex. `is_number = (....)`)

magopian:
@leplatrem better like that? https://github.com/Kinto/kinto/pull/841/files#diff-73e8dd09932ff9ff7ea1678c7a44fc39R333

leplatrem:
yes!

PR #846, opened 2016-10-04
Fix group association when members contains system.Authenticated (fixes #776)

### LINKED ISSUE #776: Adding system.Authenticated to a group does not make any user belong to it
* Create a group with `"members": ["system.Authenticated"]`
* Add `write` perm to object for `/buckets/test/group/friends`
* Any authenticated user should be allowed to write

leplatrem:
Fixes #776

- [ ] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @Natim

### DIFF
--- a/kinto/core/permission/memory.py
+++ b/kinto/core/permission/memory.py
@@ -56,8 +56,10 @@ class Permission(PermissionBase):
 
     @synchronized
     def get_user_principals(self, user_id):
+        # Fetch the groups the user is in.
         user_key = 'user:%s' % user_id
         members = self._store.get(user_key, set())
+        # Fetch the groups system.Authenticated is in.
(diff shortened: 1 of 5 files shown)


kinto/core/permission/memory.py
```diff
@@ -58,7 +58,8 @@ def remove_principal(self, principal):
     def get_user_principals(self, user_id):
         user_key = 'user:%s' % user_id
         members = self._store.get(user_key, set())
-        return members
+        group_authenticated = self._store.get('user:system.Authenticated', set())
```
Natim: This needs a comment, I don't understand what we are trying to achieve here.

Natim: Oh yes I do actually, we are trying to get the groups linked to system.Authenticated.

PR #850, opened 2016-10-05
Raise an HTTPBadRequest when trying to add a group in a group or system.Everyone to a group.

### LINKED ISSUE #794: Return 400 if a group contains system.Everyone or a group URL
Unless I'm mistaken:

* We don't support groups for anonymous requests
* We don't support recursivity in groups definitions

So we should reject with `400` if such groups definitons are created

Natim:
Fixes #794 

- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem

### DIFF
--- a/kinto/views/groups.py
+++ b/kinto/views/groups.py
@@ -5,7 +5,7 @@ from kinto.core.events import ResourceChanged, ACTIONS
 from pyramid.events import subscriber
 
 
-def validate_members(node, member):
+def validate_member(node, member):
     if member.startswith('/buckets/') or member == 'system.Everyone':
         raise colander.Invalid(node, "%r is not a valid user ID." % member)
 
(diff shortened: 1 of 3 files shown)


kinto/views/groups.py
```diff
@@ -21,6 +23,15 @@ def get_parent_id(self, request):
         parent_id = utils.instance_uri(request, 'bucket', id=bucket_id)
         return parent_id
 
+    def process_record(self, *args, **kwargs):
+        record = super(Group, self).process_record(*args, **kwargs)
+        for member in record['members']:
+            if member.startswith('/buckets/') or member == 'system.Everyone':
+                raise http_error(HTTPBadRequest(),
+                                 errno=ERRORS.INVALID_PARAMETERS,
+                                 message="'%s' is not a valid user ID." % member)
```
leplatrem: I think we should define a colander validator on members instead.

Natim: You mean using the GroupSchema to do that?

Natim: Updated :D

leplatrem:
r+ !

PR #854, opened 2016-10-06
Fix filtering with numeric id (fixes #851)

### LINKED ISSUE #851: Filtering history with ?collection_id=0 returns an empty list
* Create a collection whose name is `0`
* Filter the history with `?collection_id=0`
* Instead of obtaining the history entry, the obtained list is empty

gabisurita:
Fixes #851 

- [x] Add tests.
- [x] Add a changelog entry.
- [x] Move the id handling from resource to storage.
- [x] Move the collection_id logic from resource to the history plugin.

r? @leplatrem


Natim:
Any reason why you wanted to close the PR?

Natim:
Do you mind adding a test to make sure it works with your fix?

gabisurita:
I've closed the PR because this solves the specific problem for #851, but not the general problem that I mentioned there. I just realized that when building tests to this problem. I thought that I did this pull request too early to keep it.

Natim:
Actually we can iterate on PR so if you add commit in your branch you will be able to update the PR.

gabisurita:
I know, I just wasn't sure if I would actually be able to fix this soon. Sometimes people don't like to have open pull requests on their repositories that aren't getting close to a solution. 

Sorry for the mess I've made here.  :/

gabisurita:
This actually solves the problem when the filtering by 'id' and 'collection_id'. I've also added some view_tests for this problem.

kinto/core/resource/__init__.py
```diff
@@ -950,6 +950,10 @@ def _extract_filters(self, queryparams=None):
                 if has_invalid_value:
                     raise_invalid(self.request, **error_details)
 
+            if field in ('id', 'collection_id'):
```
Natim: What is this `collection_id` field here?

Natim:
> Sorry for the mess I've made here. :/

Btw don't be sorry I don't feel like you made any mess :)

kinto/core/resource/__init__.py
leplatrem: In the original issue, we found this bug by filtering the history with `?collection_id`.
But this actually applies to `group_id` or `record_id` too.

We could do a quick fix like this one with ` field.endswith('id')`. This will solve the history filtering problem.

But the issue is wider and applies to any field. And that's what makes it hard to fix: `?toto=123 -> {"toto": "123"} or {"toto": 123}`. I think it's more at the storage level to take care of this.

My 2 cents :)

Natim:
After thinking a bit about this issue with @leplatrem over IRC we propose the following:

- Handle the `id` field in the storage, a resource id is always a string.
- Handle `collection_id`, `group_id` and `record_id` in the history plugin.

leplatrem:
I'll try to give you some insights: 
* You can add a filtering test with numeric id here: https://github.com/Kinto/kinto/blob/f8d52232aa717f0829e50f4b0728abdfe8b28e3e/kinto/core/storage/testing.py#L312-L313
* You can *cast* id to string here: https://github.com/Kinto/kinto/blob/f8d52232aa717f0829e50f4b0728abdfe8b28e3e/kinto/core/storage/postgresql/__init__.py#L642-L643
* You can override the method `_extract_filters()` method in the history resource here: https://github.com/Kinto/kinto/blob/f8d52232aa717f0829e50f4b0728abdfe8b28e3e/kinto/plugins/history/views.py#L30 and *cast* to string the `collection_id`, `group_id`, `record_id` there if one is specified 

The fix is spread over the code base, but this way every component is aware what to do with its own stuff. 

Modifying code is usually hard to get it right at the first time :)  If you feel like you're not feeling like doing it, don't hesitate to step back, that would be completely understandable!

Good luck!

gabisurita:
Sorry, again. Im

gabisurita:
^ Sorry, I've closed the pull request because GitHub doesn't allow resets on a "pull requested" branch. Sorry about that. I'm still getting used to work with pull requests.

Natim:
After a reset you can push force.

gabisurita:
Apparently only if you reset within the the committs on the pull request. I've reseted to a commit before I did any modifications, added more commits and when I pushed it (with --force) the PR was automatically closed. I'm still trying to figure out what happened.

Anyways, I think know the solution is working both on the regular id and the history plugin filters. :)

kinto/plugins/history/views.py
```diff
@@ -31,3 +32,15 @@ class History(resource.ShareableResource):
...
+
+    def _extract_filters(self, queryparams=None):
+        filters = super(History, self)._extract_filters(queryparams)
+
+        for filt in filters:
+            if filt.field in ['record_id', 'collection_id', 'bucket_id']:
+                if isinstance(filt.value, int):
+                    filt_id = Filter(filt.field, str(filt.value), filt.operator)
+                    filters.remove(filt)
+                    filters.insert(0, filt_id)
```
leplatrem: Although this is absolutely correct, I would recommend to avoid the *mutation* of the original list, and build a new list instead. It is generally considered as a good practice to avoid *side-effects* on the input variables it receives.

```py
filters = super(...)
filters_str_id = []
for filt in filters:
    if filt.field in (...):
        filt = Filter(..., str(filt.value), ...)
    filters_str_id.append(filt)       
return filters_str_id
```

kinto/plugins/history/views.py
```diff
@@ -31,3 +32,15 @@ class History(resource.ShareableResource):
     def get_parent_id(self, request):
         self.bucket_id = request.matchdict['bucket_id']
         return instance_uri(request, 'bucket', id=self.bucket_id)
+
+    def _extract_filters(self, queryparams=None):
+        filters = super(History, self)._extract_filters(queryparams)
+
+        for filt in filters:
+            if filt.field in ['record_id', 'collection_id', 'bucket_id']:
```
leplatrem: nitpick: you can use a *tuple* instead of a list: `('record_id', ...)`

CHANGELOG.rst
```diff
@@ -20,6 +20,8 @@ This document describes changes between each past release.
 - Fix group association when members contains ``system.Authenticated`` (fixes #776)
 - Raise an error when members contains ``system.Everyone`` or a group ID (#850)
 - Fix StatsD view counter with 404 responses (#853)
+- Fixes filtering when the parameter is zero. (fixes #851)
```
leplatrem: Better «*when filtering on id with numeric values*» or something like that, no ?

Natim:
Great work thank :)

PR #859, opened 2016-10-10
Implement JSON Merge Patch  (RFC 7396 #623)

### LINKED ISSUE #623: Implement RFC 7396 (json merge)
original: https://github.com/mozilla-services/cliquet/issues/597

The proposal is:  If None is set in a PATCH request for a record attribute, that removes it

gabisurita:
Fixes #623 

- [x] Add tests.
- [x] Add a changelog entry.

Let me know if I got it wrong or there's something missing! 

Obs: Should this increase API_VERSION or be included on the API entry docs?

### DIFF
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -474,6 +474,9 @@ class UserResource(object):
 
         updated = self.apply_changes(existing, changes=changes)
 
+        # remove field if null attribute is passed (RFC 7396)
+        updated = {key: value for key, value in updated.items() if value}
+
         record_id = updated.setdefault(self.model.id_field,
                                        self.record_id)
         self._raise_400_if_id_mismatch(record_id, self.record_id)
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -17,7 +17,7 @@ from kinto.core.events import ACTIONS
 from kinto.core.storage import exceptions as storage_exceptions, Filter, Sort
 from kinto.core.utils import (
     COMPARISON, classname, native_value, decode64, encode64, json,
-    encode_header, decode_header, dict_subset
+    encode_header, decode_header, dict_subset, recursive_update_dict
 )
 
 from .model import Model, ShareableModel
@@ -474,9 +474,6 @@ class UserResource(object):
 
         updated = self.apply_changes(existing, changes=changes)
 
-        # remove field if null attribute is passed (RFC 7396)
-        updated = {key: value for key, value in updated.items() if value}
-
         record_id = updated.setdefault(self.model.id_field,
                                        self.record_id)
         self._raise_400_if_id_mismatch(record_id, self.record_id)
@@ -635,7 +632,13 @@ class UserResource(object):
                 raise_invalid(self.request, **error_details)
 
         updated = record.copy()
-        updated.update(**changes)
+
+        # recursive patch and remove field if null attribute is passed (RFC 7396)
+        content_type = str(self.request.headers.get('Content-Type'))
+        if content_type == 'application/merge-patch+json':
+            recursive_update_dict(updated, changes, ignores=[None])
+        else:
+            updated.update(**changes)
 
         try:
             return self.mapping.deserialize(updated)
--- a/tests/core/resource/test_record.py
+++ b/tests/core/resource/test_record.py
@@ -258,12 +258,7 @@ class PatchTest(BaseTest):
         result = self.resource.patch()['data']
         self.assertIn('a', result)
         self.assertNotIn('b', result['a'])
-        self.resource.request.json = {'data': {'aa': {'bb':{'cc':None}}}}
-        result = self.resource.patch()['data']
-        self.assertIn('aa', result)
-        self.assertIn('bb', result['aa'])
-        self.assertNotIn('cc', result['aa']['bb'])
-        self.resource.request.json = {'data': {'aa': {'bb':{'cc':None}}}}
+        self.resource.request.json = {'data': {'aa': {'bb': {'cc': None}}}}
         result = self.resource.patch()['data']
         self.assertIn('aa', result)
         self.assertIn('bb', result['aa'])
--- a/tests/core/resource/test_record.py
+++ b/tests/core/resource/test_record.py
@@ -227,6 +227,15 @@ class PatchTest(BaseTest):
         self.assertEquals(self.stored['id'], self.result['id'])
         self.assertEquals(self.result['position'], 10)
 
+    def test_patch_removes_attribute_if_none(self):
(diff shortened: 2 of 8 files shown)


gabisurita:
Behaviour:

```
$ echo '{"data" : {"touched": "yes" }}' | http PATCH http://localhost:8888/v1/buckets/default/collections/c0/records/r0 --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 169
Date: Mon, 10 Oct 2016 21:15:54 GMT
Etag: "1476134154134"
Last-Modified: Mon, 10 Oct 2016 21:15:54 GMT
Server: waitress

{
    "data": {
        "id": "r0", 
        "last_modified": 1476134154134, 
        "touched": "yes"
    }, 
    "permissions": {
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}

$ echo '{"data" : {"touched": null }}' | http PATCH http://localhost:8888/v1/buckets/default/collections/c0/records/r0 --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 153
Date: Mon, 10 Oct 2016 21:16:09 GMT
Etag: "1476134169802"
Last-Modified: Mon, 10 Oct 2016 21:16:09 GMT
Server: waitress

{
    "data": {
        "id": "r0", 
        "last_modified": 1476134169802
    }, 
    "permissions": {
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}
```

kinto/core/resource/__init__.py
```diff
@@ -474,6 +474,9 @@ def patch(self):
 
         updated = self.apply_changes(existing, changes=changes)
 
+        # remove field if null attribute is passed (RFC 7396)
+        updated = {key: value for key, value in updated.items() if value}
```
Natim: I would use `if value is not None` because we don't want to remove it for `False` and `0`
You might to add a test to make sure we don't remove the field with `False`, `0`, `""` (empty string) etc.

Natim:
That's so great thanks !
I think the API doc must be updated as well :)

leplatrem:
> Let me know if I got it wrong or there's something missing!

Nothing wrong! But yes some stuff is missing [from the spec](https://tools.ietf.org/html/rfc7396)

* Only use this behaviour when the specific `Content-Type` is used
* Merge recursively the objects (see a pseudo code example is given in the RFC)

I believe you should move this to `core.utils` and run some of the tests specified in the specs (especially the last one).

I know it would be better if we would have this new behaviour by default, but we can't break existing clients, so we better leave behind the request header `Content-Type: application/merge-patch+json` as the spec says.

> Obs: Should this increase API_VERSION or be included on the API entry docs?

Yes it should be incremented!

tests/core/resource/test_record.py
```diff
@@ -227,6 +227,15 @@ def test_patch_record_returns_updated_fields(self):
         self.assertEquals(self.stored['id'], self.result['id'])
         self.assertEquals(self.result['position'], 10)
 
+    def test_patch_removes_attribute_if_none(self):
+        self.resource.request.json = {'data': {'field': 'aaa'}}
+        self.resource.patch()['data']
+        self.resource.request.json = {'data': {'field': None}}
+        result = self.resource.patch()['data']
+        self.assertEquals('field' in result.keys(), False)
```
leplatrem: You can use `self.assertNotIn('field', result)`

kinto/core/resource/__init__.py
leplatrem: Theses should be moved to `self.apply_changes()` I think

gabisurita:
> I know it would be better if we would have this new behaviour by default, but we can't break existing clients, so we better leave behind the request header Content-Type: application/merge-patch+json as the spec says.

That would include accept 'application/merge-patch+json' as a valid content type on the viewset, right?

Running the RFC example over a record: 

```
$ echo '{"data":{ "title": "Goodbye!", "author" : { "givenName" : "John", "familyName" : "Doe" }, "tags":[ "example", "sample" ], "content": "This will be unchanged" }}' | http PATCH http://localhost:8888/v1/buckets/default/collections/c0/records/r1 --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 317
Date: Tue, 11 Oct 2016 15:52:53 GMT
Etag: "1476201173217"
Last-Modified: Tue, 11 Oct 2016 15:52:53 GMT
Server: waitress

{
    "data": {
        "author": {
            "familyName": "Doe", 
            "givenName": "John"
        }, 
        "content": "This will be unchanged", 
        "id": "r1", 
        "last_modified": 1476201173217, 
        "phoneNumber": "+01-123-456-7890", 
        "tags": [
            "example", 
            "sample"
        ], 
        "title": "Goodbye!"
    }, 
    "permissions": {
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}

[castro@localhost kinto]$ echo '{"data":{ "title": "Hello!", "phoneNumber": "+01-123-456-7890", "author": { "familyName": null }, "tags": [ "example" ] }}' | http PATCH http://localhost:8888/v1/buckets/default/collections/c0/records/r1 Content-Type:application/merge-patch+json --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 287
Date: Tue, 11 Oct 2016 15:52:54 GMT
Etag: "1476201174716"
Last-Modified: Tue, 11 Oct 2016 15:52:54 GMT
Server: waitress

{
    "data": {
        "author": {
            "givenName": "John"
        }, 
        "content": "This will be unchanged", 
        "id": "r1", 
        "last_modified": 1476201174716, 
        "phoneNumber": "+01-123-456-7890", 
        "tags": [
            "example"
        ], 
        "title": "Hello!"
    }, 
    "permissions": {
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}
```

leplatrem:
This is just excellent! Awesome job, definitely not easy!

We'll probably put that header by default in our official python/js clients.

gabisurita:
@leplatrem 

I just realized there is a problem with permissions related to the different content type header. For some reason, permissions is not  validated on [process_record](https://github.com/Kinto/kinto/blob/master/kinto/core/resource/__init__.py#L1142).

A quick fix would be get permissions not from ``request.validated``, but directly from ``request.json``, but I sure if this breaks something else.

Any clue why this is happening? Should we open an issue?

gabisurita:
Edit: never mind. Everything is ok, I've just used a badly formatted request.

Issue #710, opened 2016-07-07
Rename core notions


leplatrem:
Now that `cliquet` is in `kinto.core`, some naming is confusing.

In the core, resources are called collections, and objects are called records.

But, in Kinto, collections and records are instantiation of resources!

I suggest some renaming in kinto.core: 
- `collection` to `resource`
- `record` to `object`
- `collection_id` to `resource_name`  ~~(or `object_type` ?)~~
- `record_id` to `object_id`

That represents quite some work, but reducing confusion is always worth it.

leplatrem:
* `impacted_records` → `impacted_objects`

leplatrem:
*  `Total-Records` -> `Total-Objects`

almet:
This would allow a lot more readability, that's a good idea.

mansimarkaur:
I'd love to work on this.

mansimarkaur:
`object` is a global variable in Python. Should we be concerned about that when we move from `record` -> `object`?

Natim:
>Should we be concerned about that when we move from `record` -> `object`?

Yes

mansimarkaur:
So, what else should we name a record as if not object?

Natim:
Cannot we be concerned and still use it?

Natim:
To elaborate I think that most of the time we will use `object_id` and if we need to use `object` we can choose to use `obj` instead to not override the `object` Python global variable?

mansimarkaur:
Yes, obj seems to be a good option. Thanks!

leplatrem:
* `Collection endpoint` → `Plural endpoint`
* `Record endpoint` → `Object endpoint`

leplatrem:
`with_deleted=False` → `with_tombstones=False`

leplatrem:
`process_record()` → `process_object()`

glasserc:
`collection_timestamp` -> ?? I don't like `resource_timestamp`. Maybe `plural_timestamp`, or `children_timestamp`?

peterbe:
What about https://kinto.readthedocs.io/en/stable/api/1.x/records.html
We renamed the `Total-Records` header to `Total-Objects` (but kept both to be gentle). Should the `/records` (and its documentation) be redirects for `/objects` now?

leplatrem:
Nope, because in this case `records` is not a «core notion». Now core has the notion of objects, and buckets, collections, groups, records and accounts leverage it.

[...]

PR #861, opened 2016-10-13
Added details attribute to 404 errors.

### LINKED ISSUE #818: Add details attribute to 404 errors
Something like `{"details": {"id": "abc", "resource_name": collection}}`

It should be feasible to just raise custom error responses now that we have #748

mansimarkaur:
Fixes #818 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @natim @glasserc @leplatrem 
Please tell me if I'm wrong somewhere.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 **New features**
 
-- Added details attribute to 404 errors.
+- Added details attribute to 404 errors. (#818)
 - Added a new built-in plugin ``kinto.plugins.admin`` to serve the kinto admin.
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ This document describes changes between each past release.
 
 **New features**
 
+- Added details attribute to 404 errors.
 - Added a new built-in plugin ``kinto.plugins.admin`` to serve the kinto admin.
 
 
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -680,12 +680,12 @@ class UserResource(object):
         try:
             return self.model.get_record(record_id)
         except storage_exceptions.RecordNotFoundError:
-            detail_dict = {
+            details = {
                 "id": record_id,
-                "resource_name": "record"
+                "resource_name": self.request.current_resource_name
             }
             response = http_error(HTTPNotFound(), errno=ERRORS.INVALID_RESOURCE_ID,
-                                  details=detail_dict)
+                                  details=details)
             raise response
 
     def _add_timestamp_header(self, response, timestamp=None):
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -680,8 +680,12 @@ class UserResource(object):
         try:
             return self.model.get_record(record_id)
         except storage_exceptions.RecordNotFoundError:
-            response = http_error(HTTPNotFound(),
-                                  errno=ERRORS.INVALID_RESOURCE_ID)
+            detail_dict = {
+                "id": record_id,
+                "resource_name": "record"
+            }
+            response = http_error(HTTPNotFound(), errno=ERRORS.INVALID_RESOURCE_ID,
+                                  details=detail_dict)
             raise response
 
     def _add_timestamp_header(self, response, timestamp=None):
--- a/kinto/core/testing.py
+++ b/kinto/core/testing.py
@@ -70,7 +70,7 @@ class FormattedErrorMixin(object):
     """
 
     def assertFormattedError(self, response, code, errno, error,
-                             message=None, info=None):
+                             message=None, info=None, details=None):
         if isinstance(errno, Enum):
             errno = errno.value
 
@@ -79,7 +79,7 @@ class FormattedErrorMixin(object):
         self.assertEqual(response.json['code'], code)
         self.assertEqual(response.json['errno'], errno)
         self.assertEqual(response.json['error'], error)
-
+        
         if message is not None:
             self.assertIn(message, response.json['message'])
         else:  # pragma: no cover
--- a/kinto/core/testing.py
+++ b/kinto/core/testing.py
@@ -89,7 +89,6 @@ class FormattedErrorMixin(object):
         else:  # pragma: no cover
             self.assertNotIn('info', response.json)
             
-
 def get_user_headers(user):
     """Helper to obtain a Basic Auth authorization headers from the specified
     `user` (e.g. ``"user:pass"``)
--- a/kinto/core/testing.py
+++ b/kinto/core/testing.py
@@ -88,7 +88,8 @@ class FormattedErrorMixin(object):
             self.assertIn(info, response.json['info'])
         else:  # pragma: no cover
             self.assertNotIn('info', response.json)
-            
+
+
 def get_user_headers(user):
     """Helper to obtain a Basic Auth authorization headers from the specified
     `user` (e.g. ``"user:pass"``)
--- a/kinto/core/testing.py
+++ b/kinto/core/testing.py
@@ -79,7 +79,6 @@ class FormattedErrorMixin(object):
         self.assertEqual(response.json['code'], code)
         self.assertEqual(response.json['errno'], errno)
         self.assertEqual(response.json['error'], error)
-        
         if message is not None:
             self.assertIn(message, response.json['message'])
         else:  # pragma: no cover
--- a/kinto/core/testing.py
+++ b/kinto/core/testing.py
@@ -89,6 +89,11 @@ class FormattedErrorMixin(object):
         else:  # pragma: no cover
             self.assertNotIn('info', response.json)
 
+        if details is not None:
+            self.assertEqual(response.json['details'], details)
+        else:
+            self.assertNotEqual(404, code)
+
 
 def get_user_headers(user):
     """Helper to obtain a Basic Auth authorization headers from the specified
--- a/kinto/core/testing.py
+++ b/kinto/core/testing.py
@@ -70,7 +70,7 @@ class FormattedErrorMixin(object):
     """
 
     def assertFormattedError(self, response, code, errno, error,
-                             message=None, info=None, details=None):
+                             message=None, info=None):
         if isinstance(errno, Enum):
             errno = errno.value
 
@@ -88,12 +88,7 @@ class FormattedErrorMixin(object):
             self.assertIn(info, response.json['info'])
         else:  # pragma: no cover
             self.assertNotIn('info', response.json)
-
-        if details is not None:
-            self.assertEqual(response.json['details'], details)
-        else:
-            self.assertNotEqual(404, code)
-
+            
 
 def get_user_headers(user):
     """Helper to obtain a Basic Auth authorization headers from the specified
--- a/kinto/views/__init__.py
+++ b/kinto/views/__init__.py
@@ -30,10 +30,10 @@ def object_exists_or_404(request, collection_id, object_id, parent_id=''):
                            object_id=object_id)
     except exceptions.RecordNotFoundError:
         # XXX: We gave up putting details about parent id here (See #53).
-        detail_dict = {
-                "id": collection_id,
-                "resource_name": "collection"
+        details = {
+                "id": object_id,
+                "resource_name": collection_id
             }
         response = http_error(HTTPNotFound(), errno=ERRORS.MISSING_RESOURCE,
-                              details=detail_dict)
+                              details=details)
         raise response
--- a/kinto/views/__init__.py
+++ b/kinto/views/__init__.py
@@ -2,7 +2,7 @@ import random
 import string
 
 from kinto.core.storage import generators, exceptions
-from pyramid.httpexceptions import (HTTPNotFound)
+from pyramid.httpexceptions import HTTPNotFound
 from kinto.core.errors import http_error, ERRORS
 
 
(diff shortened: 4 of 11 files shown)


tests/core/test_views_errors.py
```diff
@@ -40,19 +40,16 @@ def test_backoff_header_is_present_on_error_responses(self):
             self.assertEquals(response.headers['Backoff'], '10')
 
     def test_404_is_valid_formatted_error(self):
-        response = self.app.get('/unknown', status=404)
-        self.assertFormattedError(response, 404, ERRORS.MISSING_RESOURCE, "Not Found",
-                                  "The resource you are looking for could not be found.")
```
glasserc: Shouldn't this still be a normal 404? Why are we removing this test?

kinto/core/testing.py
```diff
@@ -70,7 +70,7 @@ class FormattedErrorMixin(object):
     """
 
     def assertFormattedError(self, response, code, errno, error,
-                             message=None, info=None):
+                             message=None, info=None, details=None):
```
glasserc: It seems like we don't use this `details` variable anywhere here. Am I missing something?

kinto/views/__init__.py
```diff
@@ -29,4 +30,10 @@ def object_exists_or_404(request, collection_id, object_id, parent_id=''):
                            object_id=object_id)
     except exceptions.RecordNotFoundError:
         # XXX: We gave up putting details about parent id here (See #53).
-        raise httpexceptions.HTTPNotFound()
+        detail_dict = {
+                "id": collection_id,
+                "resource_name": "collection"
+            }
```
glasserc: I see where we define the custom 404 for collections, and for records, but I don't see any place where we define one for buckets. Where is that?

tests/core/test_views_errors.py
```diff
@@ -40,19 +40,16 @@ def test_backoff_header_is_present_on_error_responses(self):
...
-                                message="Customized.")
+                                message="Customized.",
+                                details={"id": "abc", "resource_name": "collection"})
         with mock.patch('tests.core.testapp.views.Mushroom._extract_filters',
                         side_effect=custom_404):
             response = self.app.get(self.sample_url, headers=self.headers, status=404)
         self.assertFormattedError(
-            response, 404, ERRORS.MISSING_RESOURCE, "Not Found", "Customized.")
+            response=response, code=404, errno=ERRORS.MISSING_RESOURCE, error="Not Found",
+            message="Customized.", details={"id": "abc", "resource_name": "collection"})
```
glasserc: I don't think this test is relevant to what you want to change. This looks like a test for the underlying `kinto.core` resource code and specifically like it's trying to verify that if a resource has an `_extract_filters` method that raises an exception, that exception gets passed through correctly.

I think you need tests to verify that getting a non-existent bucket, collection, or record has the correct details attached. You might look for those tests in `tests/test_views_*.py`.

tests/core/test_views_errors.py
mansimarkaur: I thought we want to add the details attribute to all 404 errors as in the normal 404 should also contain a details attribute in the response.

kinto/views/__init__.py
leplatrem: Here `"id"` should have value of `object_id` and `resource_name` value of `collection_id` (ref #710)

tests/core/test_views_errors.py
leplatrem: I don't think the test should be removed from here though ;)

As Ethan suggested, you can add more tests in `tests/test_views_*.py` instead

CHANGELOG.rst
```diff
@@ -8,6 +8,7 @@ This document describes changes between each past release.
 
 **New features**
 
+- Added details attribute to 404 errors.
```
leplatrem: nitpick: add reference to original issue.

Also, this change belongs to the *protocol* section of changes, and requires a line in the API changelog (docs/api/index.rst)

kinto/core/resource/__init__.py
```diff
@@ -674,8 +674,12 @@ def _get_record_or_404(self, record_id):
         try:
             return self.model.get_record(record_id)
         except storage_exceptions.RecordNotFoundError:
-            response = http_error(HTTPNotFound(),
-                                  errno=ERRORS.INVALID_RESOURCE_ID)
+            detail_dict = {
```
leplatrem: nitpick: you can name this variable just `details`

kinto/core/resource/__init__.py
```diff
@@ -674,8 +674,12 @@ def _get_record_or_404(self, record_id):
...
+                "id": record_id,
+                "resource_name": "record"
```
leplatrem: Here `resource_name` should have value of `self.request.current_resource_name`

kinto/views/__init__.py
mansimarkaur: @glasserc  Even i couldn't find anything for buckets.
@leplatrem id is to object_id I understood from #710 but what exactly does a resource_name define?

tests/core/test_views_errors.py
mansimarkaur: Shouldn't the valid format of a 404 error contain the details attribute as well, now? If yes, then  I think this test should be modified to ensure that the error contains details. What do you think?

kinto/views/__init__.py
leplatrem: The resource name is one of `bucket`, `collection`, `group`, `record`. Here by using `collection_id` it would work as expected. The name of this parameter is very bad I know, because it seems related to collections only, but it actually represent the resource id/name. In #710 we would fix that naming problem.

tests/core/test_views_errors.py
mansimarkaur: I found tests for unknown records in `tests/core/resource/test_record.py`.
I couldn't find anything `relevant in tests_views_*.py`
Am I missing something?

tests/core/test_views_errors.py
mansimarkaur: I found tests for unknown records in tests/core/resource/test_record.py.
I couldn't find anything relevant in tests_views_*.py
Am I missing something?

tests/test_views_records.py
```diff
@@ -38,15 +38,19 @@ def test_unknown_bucket_raises_403(self):
 
     def test_unknown_collection_raises_404(self):
         other_collection = self.collection_url.replace('barley', 'pills')
-        self.app.get(other_collection, headers=self.headers, status=404)
+        resp = self.app.get(other_collection, headers=self.headers, status=404)
+        self.assertIn('id', resp.json['details'])
+        self.assertIn('resource_name', resp.json['details'])
```
glasserc: It looks like you're only checking that the attributes are present in the error message, not that they are correct. I think we can add stronger tests that also check that `resp.json['details']['id']` is equal to `"barley"`, and similarly that the `resource_name` is `"collection"`.

I'm not sure every single test needs to be updated. Personally, I think it's OK if we leave these tests alone -- they check that we got a 404, and that's good enough. Instead I would add new tests that check that when we make a request that we expect to get a 404, that that 404 comes back with these details.

Since we expect all 404s to have a certain structure (with `id` and `resource_name` having certain attributes), it might be worthwhile to add a method somewhere called something like `check404` that enforces this structure. Then, the tests that use this method will look more declarative.

kinto/views/__init__.py
```diff
@@ -2,7 +2,8 @@
 import string
 
 from kinto.core.storage import generators, exceptions
-from pyramid import httpexceptions
+from pyramid.httpexceptions import (HTTPNotFound)
```
leplatrem: nitpick: superfluous parenthesis

kinto/core/testing.py
```diff
@@ -89,8 +88,7 @@ def assertFormattedError(self, response, code, errno, error,
             self.assertIn(info, response.json['info'])
         else:  # pragma: no cover
             self.assertNotIn('info', response.json)
-
-
+
```
leplatrem: nitpick: leave blank lines ;)

kinto/views/__init__.py
```diff
@@ -29,4 +30,10 @@ def object_exists_or_404(request, collection_id, object_id, parent_id=''):
                            object_id=object_id)
     except exceptions.RecordNotFoundError:
         # XXX: We gave up putting details about parent id here (See #53).
-        raise httpexceptions.HTTPNotFound()
+        details = {
+                "id": object_id,
+                "resource_name": collection_id
+            }
```
leplatrem: nitpick: something looks wrong with the indentation here

tests/test_views_collections.py
```diff
@@ -52,7 +52,9 @@ def test_collections_are_isolated_by_bucket(self):
         self.app.put_json('/buckets/sodas',
                           MINIMALIST_BUCKET,
                           headers=self.headers)
-        self.app.get(other_bucket, headers=self.headers, status=404)
+        resp = self.app.get(other_bucket, headers=self.headers, status=404)
+        self.assertIn('id', resp.json['details'])
+        self.assertIn('resource_name', resp.json['details'])
```
leplatrem: Usually we only keep the assertions that are relevant to the specification (ie. the test title).

For example, the test on the values of id and details is not properly relevant for the spec «*collections are isolated by bucket*».

I suggest that you only keep the assertions regarding `details` in dedicated tests similar to `test_unknown_collection_raises_404` from `test_views_records.py`, but for each of bucket, group, collection and record in the other `test_views_` files.

tests/test_views_records.py
```diff
@@ -103,7 +105,9 @@ def test_records_are_isolated_by_bucket_and_by_collection(self):
                           MINIMALIST_COLLECTION,
                           headers=self.headers)
         other = self.record_url.replace('barley', 'ba').replace('beers', 'be')
-        self.app.get(other, headers=self.headers, status=404)
+        response = self.app.get(other, headers=self.headers, status=404)
+        self.assertEqual(response.json['details']['id'], self.record['id'])
+        self.assertEqual(response.json['details']['resource_name'], 'record')
```
leplatrem: Here we don't need the assertion on the `details` fields. It's unrelated to the *specs* of this test:  «test_unknown_collection_does_not_query_timestamp»

tests/test_views_buckets.py
```diff
@@ -227,7 +227,10 @@ def test_buckets_can_be_deleted(self):
     def test_every_collections_are_deleted_too(self):
         self.app.put_json(self.bucket_url, MINIMALIST_BUCKET,
                           headers=self.headers)
-        self.app.get(self.collection_url, headers=self.headers, status=404)
+        response = self.app.get(self.collection_url, headers=self.headers, status=404)
+        self.assertEqual(response.json['details']['id'], 'barley')
+        self.assertEqual(response.json['details']['resource_name'], 'collection')
```
leplatrem: We already covered this in `test_unknown_collection_raises_404` from `test_views_collections`. The assertion is not necessary here.

tests/test_views_buckets.py
```diff
@@ -247,7 +250,9 @@ def test_every_records_are_deleted_too(self):
                           headers=self.headers)
         self.app.put_json(self.collection_url, MINIMALIST_COLLECTION,
                           headers=self.headers)
-        self.app.get(self.record_url, headers=self.headers, status=404)
+        response = self.app.get(self.record_url, headers=self.headers, status=404)
+        self.assertEqual(response.json['details']['id'], self.record_id)
+        self.assertEqual(response.json['details']['resource_name'], 'record')
```
leplatrem: These assertions should be moved to a specific test in `test_views_record.py` that says `test_unknown_record_raises_404`

kinto/views/__init__.py
```diff
@@ -29,4 +30,10 @@ def object_exists_or_404(request, collection_id, object_id, parent_id=''):
                            object_id=object_id)
     except exceptions.RecordNotFoundError:
         # XXX: We gave up putting details about parent id here (See #53).
-        raise httpexceptions.HTTPNotFound()
+        details = {
+                    "id": object_id,
+                    "resource_name": collection_id
+                }
```
leplatrem: The indentation should be like the one you did in  ` kinto/core/resource/__init__.py `

tests/test_views_buckets.py
```diff
@@ -197,8 +197,8 @@ def setUp(self):
         r = self.app.post_json(self.collection_url + '/records',
                                MINIMALIST_RECORD,
                                headers=self.headers)
-        record_id = r.json['data']['id']
-        self.record_url = self.collection_url + '/records/%s' % record_id
+        self.record_id = r.json['data']['id']
```
Natim: Why do you need to add `self` here?

kinto/views/__init__.py
```diff
@@ -29,4 +30,10 @@ def object_exists_or_404(request, collection_id, object_id, parent_id=''):
                            object_id=object_id)
     except exceptions.RecordNotFoundError:
         # XXX: We gave up putting details about parent id here (See #53).
-        raise httpexceptions.HTTPNotFound()
+        details = {
+            "id": object_id,
+            "resource_name": collection_id
+        }
+        response = http_error(HTTPNotFound(), errno=ERRORS.MISSING_RESOURCE,
+                              details=details)
```
Natim: I think details can be on the same line.

leplatrem:
This is good and ready to be merged!

The failing test is unrelated to this PR:
```
generating indices... genindex http-routingtable py-modindex

writing additional pages... index

Exception occurred:

  File "/home/travis/build/Kinto/kinto/.tox/docs/lib/python2.7/site-packages/sphinx/jinja2glue.py", line 179, in get_source

    raise TemplateNotFound(template)

TemplateNotFound: defindex.html

The full traceback has been saved in /tmp/sphinx-err-F2OoKS.log, if you want to report the issue to the developers.

Please also report this if it was a user error, so that a better error message can be provided next time.

A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

Natim:
Yes let me fix the docs build.

leplatrem:
Great! Congratz!

PR #790, opened 2016-08-25
New cornice API for validation schemas

leplatrem:
See https://github.com/mozilla-services/cornice/pull/376

* [x] Convert code to usage of `colander_validator(schema)`
* [x] Switch to `request.validated['body']`
* [x] Do not instantiate mapping at class level
* [x] Discussion about mandatory/facultative `data` on «replacing» `PUT` and make it consistent ->  #793 
* [x] Rework viewset tests
* [x] Fix tests about validation error messages (e.g. `data is missing` → `data in body: Required`)
* [x] Rename `resource.mapping` to `resource.schema` and deprecated usage of `mapping` with instantiate schema in viewset
* [x] error_handler and request attribute (#800)

PR #866, opened 2016-10-17
Implement JSON Patch (RFC 6902)

### LINKED ISSUE #458: Implement RFC 6902 (json patch)
Right now to add a single read or write permissions, you have to fetch them all, merge the new permissions with the existing ones and send them back to the server using PATCH and If-Match.

While perfectly doable, it's not super convenient when writing a client, no mention of the performances penalty.

A solution could be to add new endpoints for managing permissions more atomically?

The new endpoints could be:

- `/buckets/foo/permissions` to update everything at once
- `/buckets/foo/permissions/read` to add a single read permission (possibly many?)
- `/buckets/foo/permissions/write` same ^
- `/buckets/foo/permissions/create` same ^

With regular CRUD verbs as we know them, and the very same would go for collection permissions.

Bunch of work, and increases the surface API by a lot, so not really sure about this.

Thoughts?

gabisurita:
Fixes #458

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.
- [x] If you changed the HTTP API, update the [API_VERSION]

r? @leplatrem

### DIFF
--- a/docs/api/1.x/_details-patch-object.rst
+++ b/docs/api/1.x/_details-patch-object.rst
@@ -29,13 +29,6 @@ support to merging sub-objects and removing attibutes. For example:
 * ``{"a": {"b":"c"}}`` + ``{"a":{"d":"e"}}`` → ``{"a":{"b:c", "d":"e"}}``
 * ``{}`` + ``{"a":{"b":{"c":null}}}`` → ``{"a":{"b":{}}}``
 
-.. note::
-
-    `JSON-Patch <http://jsonpatch.com>`_ 
-    is currently not supported. Any help is welcomed though!
-    See https://github.com/Kinto/kinto/issues/458
-
-
 Light response body
 -------------------
 
@@ -60,3 +53,38 @@ See :ref:`api-permissions-payload`.
 .. ----------------
 
 .. If a read-only field is modified, a |status-400| error is returned.
+
+JSON Patch Operations
+---------------------
+
+`JSON-Patch <https://tools.ietf.org/html/rfc6902>`_ is a way to define a sequence
+of operations to be applied on a JSON object.
+
+It's possible to use ``JSON-Patch`` notation setting ``Content-Type: application/jso-patch+json``.
+
+When using this `patch` method, the request should be defined as a list of operations, for example:
+
+.. code-block:: javascript
+
+    [
+        { "op": "test", "path": "data/a", "value": "foo" },
+        { "op": "remove", "path": "/data/a" },
+        { "op": "add", "path": "/data/b", "value": [ "foo", "bar" ] },
+        { "op": "replace", "path": "/data/b", "value": 42 },
+        { "op": "move", "from": "/data/a", "path": "/data/c" },
+        { "op": "copy", "from": "/data/b", "path": "/data/d" }
+    ]
+
+For more information about each operation, please refer the 
+`JSON-Patch Specification <https://tools.ietf.org/html/rfc6902>`_.
+
+Permissions can also be altered using ``JSON-Patch``. Value is not used on permission operations
+and can be ommited.
+
+.. code-block:: javascript
+
+    [
+        { "op": "test", "path": "/permissions/read/fxa:alice" },
+        { "op": "add", "path": "/permissions/read/system.Everyone" },
+        { "op": "remove", "path": "/permissions/read/fxa:bob" }
+    ]
--- a/docs/api/1.x/_details-patch-object.rst
+++ b/docs/api/1.x/_details-patch-object.rst
@@ -21,11 +21,19 @@ The provided values are merged with the existing object. For example:
 * ``{"a":"b"}`` + ``{"a":null}`` → ``{"a":null}`` : *attributes can't be removed with patch*
 * ``{"a": {"b":"c"}}`` + ``{"a":{"d":"e"}}`` → ``{"a":{"d":"e"}}`` : *sub-objects are replaced, not merged*
 
+`JSON merge <https://tools.ietf.org/html/rfc7396>`_
+is currently supported using ``Content-Type: application/merge-patch+json``. This provides
+support to merging sub-objects and removing attibutes. For example:
+
+* ``{"a":"b"}`` + ``{"a":null}`` → ``{}`` 
+* ``{"a": {"b":"c"}}`` + ``{"a":{"d":"e"}}`` → ``{"a":{"b:c", "d":"e"}}``
+* ``{}`` + ``{"a":{"b":{"c":null}}}`` → ``{"a":{"b":{}}}``
+
 .. note::
 
-    `JSON-Patch <http://jsonpatch.com>`_ or `JSON merge <https://tools.ietf.org/html/rfc7396>`_
-    are currently not supported. Any help is welcomed though!
-    See https://github.com/Kinto/kinto/issues/623
+    `JSON-Patch <http://jsonpatch.com>`_ 
+    is currently not supported. Any help is welcomed though!
+    See https://github.com/Kinto/kinto/issues/458
 
 
 Light response body
--- a/docs/api/1.x/_details-patch-object.rst
+++ b/docs/api/1.x/_details-patch-object.rst
@@ -60,9 +60,10 @@ JSON Patch Operations
 `JSON-Patch <https://tools.ietf.org/html/rfc6902>`_ is a way to define a sequence
 of operations to be applied on a JSON object.
 
-It's possible to use ``JSON-Patch`` notation setting ``Content-Type: application/jso-patch+json``.
+It's possible to use JSON-Patch by sending the request header ``Content-Type: application/json-patch+json``.
 
-When using this `patch` method, the request should be defined as a list of operations, for example:
+When using this request header, the body should contain a list of operations,
+for example:
 
 .. code-block:: javascript
 
@@ -75,11 +76,12 @@ When using this `patch` method, the request should be defined as a list of opera
         { "op": "copy", "from": "/data/b", "path": "/data/d" }
     ]
 
-For more information about each operation, please refer the 
+For more information about each operation, please refer to
 `JSON-Patch Specification <https://tools.ietf.org/html/rfc6902>`_.
 
-Permissions can also be altered using ``JSON-Patch``. Value is not used on permission operations
-and can be ommited.
+This is very useful when altering permissions since there is no need to get
+the current value before adding some principal to the list. Value is not used
+on permission operations and can be omitted.
 
 .. code-block:: javascript
 
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -11,17 +11,12 @@ API
 Changelog
 ---------
 
-1.13 (2016-10-18)
+1.13 (2016-10-20)
 '''''''''''''''''
 
 - Add support to JSON ``patch`` format when using 
   ``Content-Type: application/json-patch+json`` (as in RFC 6902).
-- when using JSON ``patch``, changes should be described as a list of operations
-  (e.g ``[{"op": "add", "path": "/data/my_field", "value": "my_value"}]``).
-- On JSON ``patch`` operations, permissions are trated as a dictionary with no values.
-  On these operations, the ``value`` field can be omitted.
-  (e.g ``[{"op": "add", "path": "/permissions/read/system.Everyone"}]``).
-
+  For more details, see `JSON-Patch Format <1.x/records.html#json-patch-operations>`_.
 
 1.12 (2016-10-11)
 '''''''''''''''''
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -18,6 +18,11 @@ Changelog
   ``Content-Type: application/json-patch+json`` (as in RFC 6902).
   For more details, see `JSON-Patch Format <1.x/records.html#json-patch-operations>`_.
 
+1.13 (2016-10-14)
+'''''''''''''''''
+
+- ``details`` attribute present in response of 404 errors.
+
 1.12 (2016-10-11)
 '''''''''''''''''
 
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -11,6 +11,17 @@ API
 Changelog
 ---------
 
+1.13 (2016-10-18)
+'''''''''''''''''
+
+- Add support to JSON ``patch`` format when using 
+  ``Content-Type: application/json-patch+json`` (as in RFC 6902).
+- when using JSON ``patch``, changes should be described as a list of operations
+  (e.g ``[{"op": "add", "path": "/data/my_field", "value": "my_value"}]``).
+- On JSON ``patch`` operations, permissions are trated as a dictionary with no values.
+  On these operations, the ``value`` field can be omitted.
+  (e.g ``[{"op": "add", "path": "/permissions/read/system.Everyone"}]``).
+
 
 1.12 (2016-10-11)
 '''''''''''''''''
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -1157,7 +1157,13 @@ class ShareableResource(UserResource):
         existing ACE is removed (using empty list).
         """
         new = super(ShareableResource, self).process_record(new, old)
-        permissions = self.request.validated.get('permissions', {})
+        
+        content_type = str(self.request.headers.get('Content-Type'))
+        # patch is specified as a list of of operations (RFC 6902) 
+        if content_type == 'application/json-patch+json':
+            permissions = self.patch_changes.permissions
+        else:
+            permissions = self.request.validated.get('permissions', {})
 
         annotated = new.copy()
 
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -18,7 +18,7 @@ from kinto.core.storage import exceptions as storage_exceptions, Filter, Sort
 from kinto.core.utils import (
     COMPARISON, classname, native_value, decode64, encode64, json,
     encode_header, decode_header, dict_subset, recursive_update_dict,
-    extract_json_patch_changes
+    JsonPatch
 )
 
 from .model import Model, ShareableModel
@@ -466,8 +466,9 @@ class UserResource(object):
         if content_type == 'application/json-patch+json':
             #try:
                 ops = self.request.json
-                changes = extract_json_patch_changes(existing, ops)
-                print changes
+                self.patch_changes = JsonPatch(ops, existing)
+                changes = self.patch_changes.changes
+
             #except:
             #    error_details = {
             #        'name': 'data',
@@ -652,6 +653,8 @@ class UserResource(object):
         content_type = str(self.request.headers.get('Content-Type'))
         if content_type == 'application/merge-patch+json':
             recursive_update_dict(updated, changes, ignores=[None])
+        elif content_type == 'application/json-patch+json':
+            updated = self.patch_changes.updated  
         else:
             updated.update(**changes)
 
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -466,7 +466,8 @@ class UserResource(object):
 
         # patch is specified as a list of of operations (RFC 6902)
         if self._is_json_patch:
-            changes = {'data': self.request.json}
+            operations = self.request.json
+            changes = {}
         else:
             try:
                 # `data` attribute may not be present if only perms are patched.
@@ -480,7 +481,12 @@ class UserResource(object):
                 }
                 raise_invalid(self.request, **error_details)
 
-        updated = self.apply_changes(existing, changes=changes)
+        # apply changes on operations list and update changes
+        if self._is_json_patch:
+            updated = self.apply_changes(existing, changes=changes,
+                                         operations=operations)
+        else:
+            updated = self.apply_changes(existing, changes=changes)
 
         record_id = updated.setdefault(self.model.id_field,
                                        self.record_id)
@@ -607,7 +613,7 @@ class UserResource(object):
 
         return new
 
-    def apply_changes(self, record, changes):
+    def apply_changes(self, record, changes, operations=None):
         """Merge `changes` into `record` fields.
 
         .. note::
@@ -632,7 +638,7 @@ class UserResource(object):
         """
         if self._is_json_patch:
             try:
-                updated = apply_json_patch(record, changes['data'])['data']
+                updated = apply_json_patch(record, operations)['data']
                 changes.update(**updated)
             except ValueError as e:
                 error_details = {
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -709,7 +709,6 @@ class UserResource(object):
             }
             response = http_error(HTTPNotFound(), errno=ERRORS.INVALID_RESOURCE_ID,
                                   details=details)
-
             raise response
 
     def _add_timestamp_header(self, response, timestamp=None):
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -466,12 +466,11 @@ class UserResource(object):
 
         # patch is specified as a list of of operations (RFC 6902)
         if self._is_json_patch:
-            operations = self.request.json
-            changes = {}
+            requested_changes = self.request.json
         else:
             try:
                 # `data` attribute may not be present if only perms are patched.
-                changes = self.request.json.get('data', {})
+                requested_changes = self.request.json.get('data', {})
             except ValueError:
                 # If no `data` nor `permissions` is provided in patch, reject!
                 # XXX: This should happen in schema instead (c.f. ShareableViewSet)
@@ -481,12 +480,8 @@ class UserResource(object):
                 }
                 raise_invalid(self.request, **error_details)
 
-        # apply changes on operations list and update changes
-        if self._is_json_patch:
-            updated = self.apply_changes(existing, changes=changes,
-                                         operations=operations)
-        else:
-            updated = self.apply_changes(existing, changes=changes)
+        updated, applied_changes = self.apply_changes(existing,
+                                                      requested_changes=requested_changes)
 
         record_id = updated.setdefault(self.model.id_field,
                                        self.record_id)
@@ -494,7 +489,7 @@ class UserResource(object):
 
         new_record = self.process_record(updated, old=existing)
 
-        changed_fields = [k for k in changes.keys()
+        changed_fields = [k for k in applied_changes.keys()
                           if existing.get(k) != new_record.get(k)]
 
         # Save in storage if necessary.
@@ -517,7 +512,7 @@ class UserResource(object):
         elif body_behavior.lower() == 'diff':
             # Only fields that are different from those provided.
             data = {k: new_record[k] for k in changed_fields
-                    if changes.get(k) != new_record.get(k)}
+                    if applied_changes.get(k) != new_record.get(k)}
         else:
             data = new_record
 
@@ -613,7 +608,7 @@ class UserResource(object):
 
         return new
 
-    def apply_changes(self, record, changes, operations=None):
+    def apply_changes(self, record, requested_changes):
         """Merge `changes` into `record` fields.
 
         .. note::
@@ -624,11 +619,11 @@ class UserResource(object):
 
         .. code-block:: python
 
-            def apply_changes(self, record, changes):
+            def apply_changes(self, record, requested_changes):
                 # Ignore value change if inferior
                 if record['position'] > changes.get('position', -1):
                     changes.pop('position', None)
-                return super(MyResource, self).apply_changes(record, changes)
+                return super(MyResource, self).apply_changes(record, requested_changes)
 
         :raises: :exc:`~pyramid:pyramid.httpexceptions.HTTPBadRequest`
             if result does not comply with resource schema.
@@ -638,8 +633,8 @@ class UserResource(object):
         """
         if self._is_json_patch:
             try:
-                updated = apply_json_patch(record, operations)['data']
-                changes.update(**updated)
+                applied_changes = apply_json_patch(record, requested_changes)['data']
+                updated = applied_changes.copy()
             except ValueError as e:
                 error_details = {
                     'name': '',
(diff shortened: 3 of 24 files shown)


gabisurita:
Examples:

```
$ echo '[{"op":"add", "path":"/data/a", "value": "aaa"}]' | http PATCH localhost:8888/v1/buckets/default/collections/c0/records/r0 Content-Type:application/json-patch+json  --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 163
Date: Mon, 17 Oct 2016 19:39:31 GMT
Etag: "1476733171282"
Last-Modified: Mon, 17 Oct 2016 19:39:31 GMT
Server: waitress

{
    "data": {
        "a": "aaa", 
        "id": "r0", 
        "last_modified": 1476733171282
    }, 
    "permissions": {
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}

$ echo '[{"op":"move","from":"/data/a", "path":"/data/b"}]' | http PATCH localhost:8888/v1/buckets/default/collections/c0/records/r0 Content-Type:application/json-patch+json  --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 163
Date: Mon, 17 Oct 2016 19:39:37 GMT
Etag: "1476733177480"
Last-Modified: Mon, 17 Oct 2016 19:39:37 GMT
Server: waitress

{
    "data": {
        "b": "aaa", 
        "id": "r0", 
        "last_modified": 1476733177480
    }, 
    "permissions": {
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}

$ echo '[{"op":"add", "path":"/permissions/read/", "value" : "aba"}]' | http PATCH localhost:8888/v1/buckets/default/collections/c0/records/r0 Content-Type:application/json-patch+json  --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 178
Date: Mon, 17 Oct 2016 19:39:56 GMT
Etag: "1476733196928"
Last-Modified: Mon, 17 Oct 2016 19:39:56 GMT
Server: waitress

{
    "data": {
        "b": "aaa", 
        "id": "r0", 
        "last_modified": 1476733196928
    }, 
    "permissions": {
        "read": [
            "aba"
        ], 
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}

$ echo '[{"op":"add", "path":"/permissions/read/", "value" : "you"}]' | http PATCH localhost:8888/v1/buckets/default/collections/c0/records/r0 Content-Type:application/json-patch+json  --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 184
Date: Mon, 17 Oct 2016 19:40:09 GMT
Etag: "1476733209778"
Last-Modified: Mon, 17 Oct 2016 19:40:09 GMT
Server: waitress

{
    "data": {
        "b": "aaa", 
        "id": "r0", 
        "last_modified": 1476733209778
    }, 
    "permissions": {
        "read": [
            "aba", 
            "you"
        ], 
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}

$ echo '[{"op":"remove", "path":"/permissions/read/you"}]' | http PATCH localhost:8888/v1/buckets/default/collections/c0/records/r0 Content-Type:application/json-patch+json  --auth aaa:aaa
HTTP/1.1 200 OK
Content-Length: 178
Date: Mon, 17 Oct 2016 19:40:25 GMT
Etag: "1476733225272"
Last-Modified: Mon, 17 Oct 2016 19:40:25 GMT
Server: waitress

{
    "data": {
        "b": "aaa", 
        "id": "r0", 
        "last_modified": 1476733225272
    }, 
    "permissions": {
        "read": [
            "aba"
        ], 
        "write": [
            "basicauth:b4a6199b0cbf1363ef8e60c1c26e0358f49f5698c6bb4cb879437732768fcc94"
        ]
    }
}
```

kinto/core/utils.py
```diff
@@ -443,3 +443,240 @@ def instance_uri(request, resource_name, **params):
     """Return the URI for the given resource."""
     return strip_uri_prefix(request.route_path('%s-record' % resource_name,
                                                **params))
+
+
+class JsonPatch(object):
+    """
+    Process a JSON Patch over a record.
+    """
```
leplatrem: This is too big to be included like that.

I suggest that we rely on an external library instead (like https://github.com/stefankoegl/python-json-patch) and here keep a small helper only.

kinto/core/resource/__init__.py
```diff
@@ -460,17 +461,42 @@ def patch(self):
...
+            try:
+                ops = self.request.json
+                self.patch_changes = JsonPatch(ops, existing)
+
+                if '__permissions__' in existing:
+                    perms = existing['__permissions__']
+                    self.patch_changes.base_permissions.update(perms)
+
+                changes = self.patch_changes.apply_ops()
+            except StopIteration:
```
leplatrem: Why `StopIteration` ?

I think we can merge the 2 `except`s

kinto/core/resource/__init__.py
```diff
@@ -460,17 +461,42 @@ def patch(self):
...
+                error_details = {
+                    'name': 'data',
+                    'description': 'Test operation failed',
+                }
+                raise_invalid(self.request, **error_details)
+            except:
```
leplatrem: Usually *catch all* are a bad practice, it's better to specify at least one exception type

kinto/core/resource/__init__.py
```diff
@@ -460,17 +461,42 @@ def patch(self):
...
+        if content_type == 'application/json-patch+json':
+            try:
+                ops = self.request.json
+                self.patch_changes = JsonPatch(ops, existing)
+
+                if '__permissions__' in existing:
+                    perms = existing['__permissions__']
+                    self.patch_changes.base_permissions.update(perms)
+
+                changes = self.patch_changes.apply_ops()
```
leplatrem: This looks like you're applying the operations here, whereas we have a `apply_changes()` method.
I suggest that you keep the *flow*, here I would see something like:
```py
if content_type == 'application/json-patch+json':
    changes = self.request.json
else:
    try:
        changes = self.request.json.get('data', {})
        # ...
```

kinto/core/resource/__init__.py
```diff
@@ -637,6 +663,8 @@ def apply_changes(self, record, changes):
         content_type = str(self.request.headers.get('Content-Type'))
         if content_type == 'application/merge-patch+json':
             recursive_update_dict(updated, changes, ignores=[None])
+        elif content_type == 'application/json-patch+json':
+            updated = self.patch_changes.updated
```
leplatrem: This could be moved on top, before iterating on `changes` (see previous comment about *current flow*).
With a new dedicated helper:

```py
if content_type == 'application/json-patch+json':
    try:
        updated  = apply_json_patch(updated, changes)
    except ValueError as e:
        raise_invalid(...)
```

kinto/core/resource/__init__.py
```diff
@@ -460,17 +461,42 @@ def patch(self):
...
-                'name': 'data',
-                'description': 'Provide at least one of data or permissions',
-            }
-            raise_invalid(self.request, **error_details)
+        content_type = str(self.request.headers.get('Content-Type'))
+        # patch is specified as a list of operations (RFC 6902)
+        if content_type == 'application/json-patch+json':
+            try:
+                ops = self.request.json
+                self.patch_changes = JsonPatch(ops, existing)
```
leplatrem: It's usually better if we don't store such intermediary values on the resource instance. Using small function that take an input and return a result is one good practice (easy to test, composable etc.)

kinto/core/resource/__init__.py
```diff
@@ -460,17 +461,42 @@ def patch(self):
...
+
+                if '__permissions__' in existing:
+                    perms = existing['__permissions__']
+                    self.patch_changes.base_permissions.update(perms)
```
leplatrem: I know this part is a bit messy. It will be better with #790, but in the meantime **if** you can keep the *permissions* to the children class `ShareableResource` it would be better :) Didn't look in details if feasible for this case though

kinto/core/resource/__init__.py
```diff
@@ -1139,13 +1167,19 @@ def process_record(self, new, old=None):
...
+            permissions = self.patch_changes.permissions
+        else:
+            permissions = self.request.validated.get('permissions', {})
 
         annotated = new.copy()
 
         if permissions:
             is_put = (self.request.method.lower() == 'put')
-            if is_put:
+            if is_put or content_type == 'application/json-patch+json':
```
leplatrem: I don't think we should have this here. Do you remember why?

tests/core/resource/test_record.py
```diff
@@ -227,6 +227,134 @@ def test_patch_record_returns_updated_fields(self):
         self.assertEquals(self.stored['id'], self.result['id'])
         self.assertEquals(self.result['position'], 10)
 
+    def test_json_patch_add(self):
```
leplatrem: This could deserve a dedicated test suite, no? `class JsonPatchTest(...)`

tests/core/resource/test_record.py
```diff
@@ -227,6 +227,134 @@ def test_patch_record_returns_updated_fields(self):
...
+        ]
+        result = self.resource.patch()['data']
+        self.assertEqual(result['c'], 'ccc')
+        self.assertEqual(result['b'][1], 'bbb')
+        self.assertEqual(result['b'][2], 'bbbb')
+        self.assertEqual(result['b'][0]['f']['g'], 'fff')
+
+    def test_json_patch_remove(self):
+        self.resource.request.json = {'data': {'a': 'aaa', 'b': ['bb','bbb']}}
+        self.resource.patch()
```
leplatrem: With a dedicated test suite, you could move this a to `setUp()`

tests/core/resource/test_record.py
```diff
@@ -227,6 +227,134 @@ def test_patch_record_returns_updated_fields(self):
...
+        ]
+        self.assertRaises(httpexceptions.HTTPBadRequest, self.resource.patch)
+
+    def test_json_patch_format_not_accepted_without_header(self):
+        self.resource.request.json = {'data': {'a': 'aaa'}}
+        self.resource.patch()
+        self.resource.request.json = [
+            {'op': 'copy', 'from': '/data/a', 'path': '/data/c'},
+        ]
+        self.assertRaises(AttributeError, self.resource.patch)
```
leplatrem: This means the client will receive a `500 Internal Error` instead of a clean `400` error response

tests/core/resource/test_record.py
```diff
@@ -227,6 +227,134 @@ def test_patch_record_returns_updated_fields(self):
...
+            {'op': 'remove', 'path': '/data/b/1'},
+        ]
+        result = self.resource.patch()['data']
+        self.assertEqual(len(result['b']), 1)
+
+        self.resource.request.json = [
+            {'op': 'test', 'path': '/data/b/0', 'value': 'bbb'},
+            {'op': 'remove', 'path': '/data/b/0'},
+        ]
+        self.assertRaises(httpexceptions.HTTPBadRequest, self.resource.patch)
```
leplatrem: I suggest that you keep only *one* group of assertions per test. For example, you could split into two:
* test_json_patch_test_success
* test_json_patch_test_failure

Also, here we don't know if it failed because of the `remove` or the `test`

tests/core/resource/test_views.py
```diff
@@ -485,11 +485,16 @@ def test_invalid_content_type_header_on_record_returns_415(self):
...
+            "Content-Type header should be one of [",
+            "'application/json-patch+json'",
+            ", ",
+            "'application/json'",
+            ", ",
+            "'application/merge-patch+json'",
+            "]")
+        for message in messages:
+            self.assertIn(message, resp.json['message'])
+        self.assertEquals(len("".join(messages)), len(resp.json['message']))
```
leplatrem: There is something with this test, I can't suggest easily a better way, but it can be improved IMO :)

gabisurita:
So, here goes a second sketch, using jsonpatch module.

My main concern is how to treat permissions. Since there is no support to sets on JSON Object notation. I've choose to treat them as dicts. Permission operations would look like:

- ``{"op": "add", "path": "/permissions/read/alice", "value": ""}``
- ``{"op": "add", "path": "/permissions/read/alice"}`` (I've also add the option to omit ``value`` on this case).
- ``{"op": "remove", "path": "/permissions/read/alice"}``

Another approach would be to treat permissions as lists, but the remove option would need to be specified as a numeric value, which I think is not consistent with the use case presented.

I've also wasn't able to remove the read dependency on ``__permissions__``, but permissions are only updated on ``SharableResource.process_record()``.

We have a much smaller diff now! :)

kinto/core/resource/__init__.py
```diff
@@ -4,6 +4,8 @@
 import colander
 import venusian
 import six
+from jsonpatch import JsonPatchException
+from jsonpointer import JsonPointerException
```
leplatrem: I'd rather catch those to `utils.py` and raise a simple ValueError from them. From the resource point of view, these are details of implementation

kinto/core/resource/__init__.py
```diff
@@ -460,20 +463,29 @@ def patch(self):
...
+                    'name': 'data',
+                    'description': 'Provide at least one of data or permissions',
+                }
+                raise_invalid(self.request, **error_details)
 
         updated = self.apply_changes(existing, changes=changes)
 
+        # changes should be a json object for ``Response-Behavior`` processing.
+        if content_type == 'application/json-patch+json':
+            changes = updated.copy()
```
leplatrem: why not do this in the `apply_changes()` method instead ?

kinto/core/resource/__init__.py
```diff
@@ -622,6 +634,18 @@ def apply_changes(self, record, changes):
...
+        if content_type == 'application/json-patch+json':
+            try:
+                changes, _ = apply_json_patch(record, changes)
+            except (JsonPatchException, JsonPointerException) as e:
+                error_details = {
+                    'name': '',
+                    'description': 'JSON Patch operation failed'
+                }
+                raise_invalid(self.request, e, **error_details)
+
```
leplatrem: here I would do a `else`/`else if` for the rest of this method, since what's coming below does not apply to json patch. 

plus, it would reduce the complexity for the reader, where each block applies to a particular case

kinto/core/resource/__init__.py
```diff
@@ -1139,7 +1164,14 @@ def process_record(self, new, old=None):
         existing ACE is removed (using empty list).
         """
         new = super(ShareableResource, self).process_record(new, old)
-        permissions = self.request.validated.get('permissions', {})
+
+        content_type = str(self.request.headers.get('Content-Type'))
+        # patch is specified as a list of of operations (RFC 6902)
+        if content_type == 'application/json-patch+json':
```
leplatrem: since this is repeated several time it could go to a computed `@property` or be initialized as private attribute in constructor (like `if self._is_json_patch`). What do you think?

kinto/core/utils.py
```diff
@@ -443,3 +444,40 @@ def instance_uri(request, resource_name, **params):
...
+            if op['path'].startswith(('/permissions/read/',
+                                      '/permissions/write/')):
+                op['value'] = op['path'].split('/')[-1]
+
+    result = jsonpatch.apply_patch(resource, ops)
+
+    data = result['data']
+    permissions = result['permissions']
+
+    return data, permissions
```
leplatrem: I think it's acceptable to return `result` here. We give an input and a patch, we receive an output. If you need to keep them separate, then you could also take two separate arguments as input

kinto/core/utils.py
```diff
@@ -443,3 +444,40 @@ def instance_uri(request, resource_name, **params):
...
+    """
+    data = record.copy()
+    permissions = {'read': set(), 'write': set()}
+
+    # To be used with SharableResource object (patching permissions)
+    if '__permissions__' in record:
+        permissions.update(record['__permissions__'])
+
+    # Map permissions as dicts since jsonpatch module doesn't accept sets.
+    permissions = {k: {i: i for i in v} for k, v in permissions.items()}
```
leplatrem: I believe that the easiest cast for set are lists, no?  `{k: list(v) for k, v ...}`

kinto/core/utils.py
```diff
@@ -443,3 +444,40 @@ def instance_uri(request, resource_name, **params):
...
+    :param bool only_data: param to limit the scope of the patch only to 'data'.
+    :returns dict data: patched record data.
+             dict permissions: patched record permissions
+    """
+    data = record.copy()
+    permissions = {'read': set(), 'write': set()}
+
+    # To be used with SharableResource object (patching permissions)
+    if '__permissions__' in record:
+        permissions.update(record['__permissions__'])
```
leplatrem: why not something like (*roughly*): `resource = {'data': record.copy(), 'permissions': record.pop('__permissions__', {})` ?

kinto/core/utils.py
gabisurita: The problem to cast them with lists is the expected behaviour for lists on jsonpatch. There is no way to remove by value.

the easier approach I've found using the ``jsonpatch`` package was to cast them as ``dict``.

gabisurita: If it was as list, the operations would look like:

- ``{"op": "add", "path": "permissions/read/0", "value": "alice"}``
- ``{"op": "remove", "path": "permissions/read/0"}`` *(which I think is inconsistent with what was presented on the issue)*

Natim: I don't understand the /0 here, why would we need it?

kinto/core/utils.py
leplatrem: if you give `pop()` a second argument, it's becomes a default value.
```py
data = record.copy()
permissions = record.pop('__permissions__', {})
resource = {'data': data, 'permissions': permissions}
```

kinto/core/utils.py
leplatrem: Oh, very interesting! Please add all those comments regarding removing by value along the code :)

kinto/core/resource/__init__.py
```diff
@@ -460,17 +464,21 @@ def patch(self):
...
-            # If no `data` nor `permissions` is provided in patch, reject!
-            # XXX: This should happen in schema instead (c.f. ShareableViewSet)
-            error_details = {
-                'name': 'data',
-                'description': 'Provide at least one of data or permissions',
-            }
-            raise_invalid(self.request, **error_details)
+        # patch is specified as a list of of operations (RFC 6902)
+        if self._is_json_patch:
+            changes = {'data': self.request.json}
```
leplatrem: Why did you add an additional `data` level here?

docs/api/1.x/_details-patch-object.rst
```diff
@@ -60,3 +53,38 @@ See :ref:`api-permissions-payload`.
...
 
 .. If a read-only field is modified, a |status-400| error is returned.
+
+JSON Patch Operations
+---------------------
+
+`JSON-Patch <https://tools.ietf.org/html/rfc6902>`_ is a way to define a sequence
+of operations to be applied on a JSON object.
+
+It's possible to use ``JSON-Patch`` notation setting ``Content-Type: application/jso-patch+json``.
```
leplatrem: nitpick: use `*JSON-Patch*` notation instead of back ticks (for code)

docs/api/index.rst
```diff
@@ -11,6 +11,17 @@ API
...
+1.13 (2016-10-18)
+'''''''''''''''''
+
+- Add support to JSON ``patch`` format when using 
+  ``Content-Type: application/json-patch+json`` (as in RFC 6902).
+- when using JSON ``patch``, changes should be described as a list of operations
+  (e.g ``[{"op": "add", "path": "/data/my_field", "value": "my_value"}]``).
+- On JSON ``patch`` operations, permissions are trated as a dictionary with no values.
+  On these operations, the ``value`` field can be omitted.
+  (e.g ``[{"op": "add", "path": "/permissions/read/system.Everyone"}]``).
```
leplatrem: I wouldn't provide so much details here, but just a link to a record PATCH for example.

docs/api/1.x/_details-patch-object.rst
leplatrem:
```
to use *JSON Patch* with the request header ``Content-Type: application/json-patch+json`` instead of just ``Content-Type: application/json``
```

docs/api/1.x/_details-patch-object.rst
```diff
@@ -60,3 +53,38 @@ See :ref:`api-permissions-payload`.
...
+
+When using this `patch` method, the request should be defined as a list of operations, for example:
```
leplatrem: When using this request header, the body should contain a list of operations. For example:

docs/api/1.x/_details-patch-object.rst
```diff
@@ -60,3 +53,38 @@ See :ref:`api-permissions-payload`.
...
+    [
+        { "op": "test", "path": "data/a", "value": "foo" },
+        { "op": "remove", "path": "/data/a" },
+        { "op": "add", "path": "/data/b", "value": [ "foo", "bar" ] },
+        { "op": "replace", "path": "/data/b", "value": 42 },
+        { "op": "move", "from": "/data/a", "path": "/data/c" },
+        { "op": "copy", "from": "/data/b", "path": "/data/d" }
+    ]
+
+For more information about each operation, please refer the
```
leplatrem: typo: please refer to

docs/api/1.x/_details-patch-object.rst
```diff
@@ -60,3 +53,38 @@ See :ref:`api-permissions-payload`.
...
+        { "op": "replace", "path": "/data/b", "value": 42 },
+        { "op": "move", "from": "/data/a", "path": "/data/c" },
+        { "op": "copy", "from": "/data/b", "path": "/data/d" }
+    ]
+
+For more information about each operation, please refer the 
+`JSON-Patch Specification <https://tools.ietf.org/html/rfc6902>`_.
+
+Permissions can also be altered using ``JSON-Patch``. Value is not used on permission operations
+and can be ommited.
```
leplatrem: This is very useful when altering permissions since there is no need to get the current value before adding some principal to the list.

docs/api/1.x/_details-patch-object.rst
glasserc: I'd probably use neither *emphasis* or `code` -- it's a proper noun, so capitals are enough.

I'd probably use `by` somewhere here, like:
```
It's possible to use JSON-Patch by sending the request header ``Content-Type: application/json-patch+json``.
```

I don't think it's necessary to explain "instead of (something else)".

Sorry to bikeshed :)

docs/api/1.x/_details-patch-object.rst
glasserc: `omitted`

kinto/core/resource/__init__.py
glasserc: Hmmm. Do I understand this correctly? We mutate `changes` because later we use `changes` to find any keys/values in the record that might have changed, which we do so that we know whether the record needs to be saved and also so we can respond with just changed fields.

If that's correct, then we're using `changes` as a strange bidirectional argument -- both passing data in, and getting data out. It seems to me like when it's input, `changes` is `{"data": [... list of operations ...]}`, but when it's output, it's `{"data": [... list of operations ...], "key1": "new_value1", "key2": "new_value2"}`. In other words, both kinds of data are mixed up. Is that true?

If I understand correctly, then I'd much rather have some other mechanism for returning the changed fields of the record. Perhaps `apply_changes` should return `(updated, changed_fields)`. Unfortunately this would be an API-breaking change. At the very least, we should document this behavior of `apply_changes` if we're going to take advantage of it, and I would say we should clear the old `data` entry when we update it.

kinto/core/utils.py
glasserc: Yeah, that's a great insight and one that I wouldn't have realized just from looking at the code :)

tests/core/resource/test_record.py
```diff
@@ -347,6 +347,114 @@ def test_returns_changed_fields_if_behaviour_is_light(self):
...
+
+    def test_json_patch_add(self):
+        self.resource.request.json = [
+            {'op': 'add', 'path': '/data/c', 'value': 'ccc'},
+            {'op': 'add', 'path': '/data/b/1', 'value': 'bbb'},
+            {'op': 'add', 'path': '/data/b/-', 'value': 'bbbb'},
+        ]
+        result = self.resource.patch()['data']
+        self.assertEqual(result['c'], 'ccc')
+        self.assertEqual(result['b'][2], 'bbb')
```
glasserc: I think this test is confusing because you're inserting between elements 0 and 1, but you're inserting the same value as element 1. I think this test should use `result['b'][1]` and its value should be different than `bbb`. Also, you might put add an assertion about the `len` here (since you have them in other tests too).

tests/core/resource/test_record.py
```diff
@@ -347,6 +347,114 @@ def test_returns_changed_fields_if_behaviour_is_light(self):
...
+        self.assertEqual(result['b'][3], 'bbbb')
+
+    def test_json_patch_remove(self):
+        self.resource.request.json = [
+            {'op': 'remove', 'path': '/data/a'},
+            {'op': 'remove', 'path': '/data/b/0'},
+        ]
+        result = self.resource.patch()['data']
+        self.assertNotIn('a', result)
+        self.assertEqual(len(result['b']), 1)
```
glasserc: And that it should be `bbb`?

kinto/core/resource/__init__.py
gabisurita: Yes. That's correct, and it may be a strange approach, but was the way that I've found to avoid modifying ``apply_changes`` signature, without patching arguments early on update, or avoiding updating changes after the ``apply_changes`` call. Perhaps we could use a keyword argument to pass something like ``operations`` and ``changes = {}`` to ``apply_changes``. It would still be a bidirectional argument, but never on the same call.

docs/api/index.rst
gabisurita: Where should I include the link?

gabisurita:
I've added a keyword argument to ``apply_changes`` in order to solve the issues related to the ``changes`` object. Is there a problem with that?

kinto/core/resource/__init__.py
```diff
@@ -599,7 +613,7 @@ def process_record(self, new, old=None):
 
         return new
 
-    def apply_changes(self, record, changes):
+    def apply_changes(self, record, changes, operations=None):
```
leplatrem: Sorry but I don't get why we couldn't use the `changes` object to pass the operations here?

docs/api/index.rst
glasserc: I don't understand this question. I'd probably replace the second and third bullet points and just add a sentence to the first that says something like

```rst
For more details, see `JSON Patch Format <link here>`_.
```

Are you asking where the link should point to? I would have it point to the JSON Patch section above, if that's possible.

glasserc: Right, when he said "a link" I think he meant a link that points to the API docs, not a "live" URL or command to demonstrate the PATCH functionality. So pointing to the API docs sounds fine to me, since this is just an index.

kinto/core/resource/__init__.py
gabisurita: > I think we should just "bite the bullet" and make a breaking change. Have apply_changes take the "operations" (I think that's a better name) and have it return (updated, changes).

We should change the return value only when we have a JSON-Patch, right? Or should we do it also for regular calls? Just to make sure that I've got it right, the approach would be:

- Keep the signature ``def apply_changes(self, record, changes)``
- In case of a regular call, ``return updated``
- If Json Patch, ``return (updated, changes)``

Is that clear?

glasserc: I think we should do it for all calls for consistency. This also allows resources to support their own modification syntaxes, although I guess that nobody will actually do that. So I'm proposing the signature `def apply_changes(self: Resource, record: Record, requested_changes: object) -> (updated: Record, applied_changes: {FieldName: Value})`. (Or maybe `applied_changes` should just be `[FieldName]`?)

glasserc: This means that existing resources will have to be updated to return the same `requested_changes` that they got in their parameters, but I think that's OK for a little more flexibility.

leplatrem:
The tests are broken of the release of cornice 2.0. We'll make our best to merge #790 ASAP

kinto/core/resource/__init__.py
```diff
@@ -638,8 +633,8 @@ def apply_changes(self, record, changes):
         """
```
glasserc: You need to update the `:returns:` and `:rtype:`. To be honest, I would love a great description of `applied_changes`, including what it should look like when some changes are happening to a deeply nested object.

tests/core/resource/test_views.py
glasserc: I think this is OK. It's pretty clever and probably the best you can do without using something like `eval`. Of course, you don't need `", "` twice -- you might take it out entirely just to make it clear that we're not comparing against the literal text of the error message. How about something like this?

```python
self.assert_(resp.json['message'].startswith('Content-Type header should be one of ['))
self.assert_(resp.json['message'].endswith(']'))
for message in messages:
    self.assertIn(message, resp.json['message'])
```

gabisurita:
I'm having some problems after merging #790. Do you have any idea what's causing this?

```
$ echo '[{"op": "add", "path": "/data/aaa", "value": "aaa"}]' | http PATCH localhost:8888/v1/buckets/b1/collections/c1/records/r1 Content-Type:application/json-patch+json
HTTP/1.1 400 Bad Request
Content-Length: 403
Date: Mon, 24 Oct 2016 13:50:37 GMT
Server: waitress

{
    "code": 400, 
    "details": [
        {
            "description": "\"[{\"op\": \"add\", \"path\": \"/data/aaa\", \"value\": \"aaa\"}]\n\" is not a mapping type: Does not implement dict-like functionality.", 
            "location": "body", 
            "name": ""
        }
    ], 
    "errno": 107, 
    "message": "\"[{\"op\": \"add\", \"path\": \"/data/aaa\", \"value\": \"aaa\"}]\n\" is not a mapping type: Does not implement dict-like functionality."
}
```

glasserc:
I was able to run this branch and make the request you posted without any problems. Is it maybe something silly like an outdated version of Cornice in your virtualenv? If you're running `make serve`, that should take care of it. Can you track down where something is being checked for mapping functionality?

leplatrem:
I can help, I'm the one who merged #790 and introduced some additional work on validation. I'd be glad to merge the PR myself providing the last bit of effort! You've done an amazing job here, and I wouldn't want you to get discouraged...

leplatrem:
I'll merge the PR myself please. You've provided a tremendous amount of work here, and you should now lean back for a while :)

gabisurita:
Great! Please let me know if you need anything else.

Issue #868, opened 2016-10-19
Documentation broken with Sphinx 1.5


leplatrem:
The `defindex.html` template was removed:
https://github.com/sphinx-doc/sphinx/issues/2986#issuecomment-252157323

PR #869, opened 2016-10-19
Pin Sphinx to 1.4 (ref #868)

leplatrem:
Temporary fix for #868 

We should inherit another template in `docs/_templates/indexcontent.html` since `defindex.html` was removed in Sphinx 1.5a2


Natim:
If possible I would rather fix it like that: https://github.com/Kinto/kinto/pull/870

PR #875, opened 2016-10-24
Add some docs about testing

leplatrem:
r? @glasserc

### DIFF
--- a/docs/community.rst
+++ b/docs/community.rst
@@ -152,9 +152,9 @@ The `tests are the specifications <http://blog.mathieu-leplatre.info/your-tests-
 * Each test class represents a situation or feature (e.g. ``class AnonymousCreationTest(unittest.TestCase):``)
 * Each test method represents an aspect of the specification (e.g. ``def test_creation_is_allowed_if_enabled_in_settings(self):``)
 * Each test method is independant
-* The assertions should only reflect the specifications of this test method
+* The assertions should only correspond to the aspect of the specification that this method targets
 * The ``setUp()`` method contains some initialization steps that are shared among the methods
-* If the methods have different initialization steps, they should probably bit split into different test classes
+* If the methods have different initialization steps, they should probably be split into different test classes
 
 When contributing a **bug fix**:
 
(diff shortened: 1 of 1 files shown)


docs/community.rst
```diff
@@ -144,6 +144,40 @@ Get started!
...
+-------------------
+
+The `tests are the specifications <http://blog.mathieu-leplatre.info/your-tests-as-your-specs.html>`_.
+
+* Each test class represents a situation or feature (e.g. ``class AnonymousCreationTest(unittest.TestCase):``)
+* Each test method represents an aspect of the specification (e.g. ``def test_creation_is_allowed_if_enabled_in_settings(self):``)
+* Each test method is independant
+* The assertions should only reflect the specifications of this test method
+* The ``setUp()`` method contains some initialization steps that are shared among the methods
+* If the methods have different initialization steps, they should probably bit split into different test classes
```
glasserc: `be split` or `get split`

docs/community.rst
```diff
@@ -144,6 +144,40 @@ Get started!
...
 
+Testing methodology
+-------------------
+
+The `tests are the specifications <http://blog.mathieu-leplatre.info/your-tests-as-your-specs.html>`_.
+
+* Each test class represents a situation or feature (e.g. ``class AnonymousCreationTest(unittest.TestCase):``)
+* Each test method represents an aspect of the specification (e.g. ``def test_creation_is_allowed_if_enabled_in_settings(self):``)
+* Each test method is independant
+* The assertions should only reflect the specifications of this test method
```
glasserc: I think I know what you mean here, but I think this sentence is kind of confusing. How about `The assertions should only correspond to the aspect of the specification that this test method is exercising`?

docs/community.rst
```diff
@@ -144,6 +144,40 @@ Get started!
...
+
+1. Write a test that reproduces the problem: it should fail because of the bug
+2. Fix the faulty piece of code
+3. The test should now pass
+
+When contributing a **new feature**:
+
+* Do not rush on the code
+* Step by step, you'll write tests for each aspect and each edge case of the feature
+* Start very small: one simple test for the most simple situation
```
glasserc: I think you're trying to propose a TDD-style approach where you write tests before writing code for your feature, but you didn't say that explicitly here? Also, `simplest`. I think some of these bullet points would be better as a paragraph of prose, or maybe a link to some other blog post?

Issue #877, opened 2016-10-25
Error views seem to happen after the pyramid-tm tween


glasserc:
See mozilla-services/kinto-fxa#33. We're hacking around this for now but it would be good to have a real fix.

leplatrem:
Shouldn't this be reported to pyramid-tm instead?

Or couldn't we prevent to do smart stuff in the 403 view? Like why are we transforming a 403 into a 401 anyway?

Natim:
> Why are we transforming a 403 into a 401 anyway?

That's to be consistent with the Kinto protocol. If you are not authenticated you should have a 401 and not a 403.

glasserc:
I'm not sure if it's a problem with `pyramid-tm` or if we're using it wrong.

leplatrem:
Tracked upstream: https://github.com/Pylons/pyramid_tm/issues/40

leplatrem:
I suggest that we close this, and make sure the code we put in those error views do not hit the backends

PR #878, opened 2016-10-25
Manually abort transaction in error view

glasserc:
Fixes mozilla-services/kinto-fxa#33.

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.

r? @Natim 

This fixes mozilla-services/kinto-fxa#33.

This is an ugly hack; the real fix involves fixing kinto#877.

### DIFF
--- a/kinto/core/views/errors.py
+++ b/kinto/core/views/errors.py
@@ -1,3 +1,5 @@
+import transaction
+
 from pyramid import httpexceptions
 from pyramid.httpexceptions import HTTPTemporaryRedirect
 from pyramid.settings import asbool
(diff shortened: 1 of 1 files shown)


Natim:
I agree with the fix, I wonder if we can add a test there.

leplatrem:
Good catch!

I think that the transaction is opened by this line: https://github.com/Kinto/kinto/blob/master/kinto/core/authorization.py#L35

Maybe we could put the fix there (something like: if no transaction is opened before the call on the backend, then close it just after there)

Natim:
How do you detect if the transaction is opened or not?

leplatrem:
> How do you detect if the transaction is opened or not?

Maybe with that ? https://github.com/zopefoundation/transaction/blob/master/transaction/_transaction.py#L85

Maybe the current patch is the best way to go after all :)

leplatrem:
After some investigation, we realized that kinto-fxa was not implemented properly. We'll propose a fix there. 

Let's keep #877 and #879 open, it does not hurt to fix them some day, but the root cause is fxa in this case (accessing `request.effective_principals` several times in the request cycle shouldn't induce several hits to auth server and cache backend)

Thanks for spotting the bug !! I close this but remain open to any objection ;)

glasserc:
This was eventually solved in mozilla-services/kinto-fxa#35.

PR #891, opened 2016-10-27
Add user principals to hello view

### LINKED ISSUE #874: Add user principals to hello view
Along `user` info, show the list of user principals (for groups notably)

gabisurita:
Fixes #874 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -11,12 +11,13 @@ API
 Changelog
 ---------
 
-1.13 (2016-10-20)
+1.13 (2016-10-27)
 '''''''''''''''''
 
 - Add support to JSON ``patch`` format when using 
   ``Content-Type: application/json-patch+json`` (as in RFC 6902).
   For more details, see `JSON-Patch Format <1.x/records.html#json-patch-operations>`_.
+- Add a list of ``principals`` to ``hello`` view.
 
 1.13 (2016-10-14)
 '''''''''''''''''
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -11,7 +11,7 @@ API
 Changelog
 ---------
 
-1.13 (2016-10-27)
+1.13 (Unreleased)
 '''''''''''''''''
 
 - Add support to JSON ``patch`` format when using 
--- a/tests/test_views_hello.py
+++ b/tests/test_views_hello.py
@@ -2,7 +2,8 @@ from kinto import __version__ as VERSION
 
 from kinto.core.testing import unittest
 
-from .support import BaseWebTest
+from .support import (BaseWebTest, MINIMALIST_BUCKET,
+                      MINIMALIST_GROUP)
 
 
 class HelloViewTest(BaseWebTest, unittest.TestCase):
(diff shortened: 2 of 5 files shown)


tests/test_views_hello.py
```diff
@@ -23,6 +24,17 @@ def test_returns_user_id_if_authenticated(self):
...
 
+    def test_returns_user_principals_if_authenticated(self):
+        group_url = '/buckets/beers/groups/users'
+        group = MINIMALIST_GROUP.copy()
+        group['data']['members'].append(self.principal)
+        self.app.put_json('/buckets/beers', MINIMALIST_BUCKET, headers=self.headers)
+        self.app.put_json(group_url, group, headers=self.headers)
+        response = self.app.get('/', headers=self.headers).json['user']['principals']
+        principals = ('system.Everyone', 'system.Authenticated',
+                      group_url, self.principal.split(':')[1])
```
Natim: Why `self.principal.split(':')[1]`?

Natim: It should be something like `basicauth:1234` right?

gabisurita: ``effective_principals`` actually omits ``basicauth`` prefix. Should the view include ``basicauth:1234`` or just ``1234``.

leplatrem: Yes, this is case. 

So, if we want we can put `request.effective_principals + [request.prefixed_userid]` in the hello view instead?

We do some similar black magics here: https://github.com/Kinto/kinto/blob/master/kinto/core/authorization.py#L59

leplatrem: I think it's alright since we already have prefixed_userid in `hello[user]`

Natim: Note that we also remove the other one: https://github.com/Kinto/kinto/blob/master/kinto/core/authorization.py#L64

leplatrem:
Could you add an entry in the API changelog please?

Thanks :)

leplatrem:
> Please at least remove the principal without the prefix from the list of principals.

Why is it so important? From the Pyramid point of view, it's there!

Natim:
> Why is it so important? From the Pyramid point of view, it's there!
Actually from Kinto point of view it is not and if it is it is an issue because you could have conflict between two Authorizaton policy that would give the same `user_id` principal while it shouldn't be possible

docs/api/index.rst
```diff
@@ -11,12 +11,13 @@ API
 Changelog
 ---------
 
-1.13 (2016-10-20)
+1.13 (2016-10-27)
```
Natim: Isn't this supposed to be `1.14 (unreleased)`

leplatrem: `1.13 unreleased`, but that's acceptable

Natim: It seems to me that 1.13 have been used already: https://github.com/Kinto/kinto/pull/891/files#diff-2c14b00f3393d6e7355fb480528f38e3R22

PR #899, opened 2016-11-03
CORS response for redirect view (fixes #852)

### LINKED ISSUE #852: An OPTIONS request on "/v1" is answered with a 307 instead of a 200
example query:

```
curl 'https://kinto.dev.mozaws.net/v1' -X OPTIONS -H 'Host: kinto.dev.mozaws.net' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:50.0) Gecko/20100101 Firefox/50.0' -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' -H 'Accept-Language: fr,fr-FR;q=0.8,en-US;q=0.5,en;q=0.3' --compressed -H 'Access-Control-Request-Method: GET' -H 'Access-Control-Request-Headers: authorization' -H 'Origin: http://localhost:8000' -H 'Connection: keep-alive' -H 'Cache-Control: max-age=0'
```

As this OPTIONS request is made/used by the browser for CORS validation, it breaks on the 307.

leplatrem:
Fixes #852 

What should we return here then?

capture d ecran de 2016-11-03 17-05-53

### DIFF
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -9,6 +9,7 @@ from pyramid.exceptions import ConfigurationError
 from pyramid.httpexceptions import (HTTPTemporaryRedirect, HTTPGone,
                                     HTTPBadRequest)
 from pyramid.renderers import JSON as JSONRenderer
+from pyramid.response import Response
 from pyramid.security import NO_PERMISSION_REQUIRED
 from pyramid.interfaces import IAuthenticationPolicy
 from pyramid.settings import asbool, aslist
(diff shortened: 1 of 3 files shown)


kinto/core/initialization.py
```diff
@@ -72,6 +72,10 @@ def setup_version_redirection(config):
         return
 
     def _redirect_to_version_view(request):
+        if request.method.lower() == 'options':
+            from pyramid.response import Response
+            return Response()
```
Natim: We need to make sure cors are applied there.
```
    def test_redirects_benefits_from_cors_setup(self):
        app = self.make_app()
        headers = {
            'Origin': 'lolnet.org',
            'Access-Control-Request-Method': 'GET'
        }
        resp = app.options('/v0', headers=headers, status=200)
        self.assertIn('Access-Control-Allow-Origin', resp.headers)
```

PR #904, opened 2016-11-08
Upgrade to cornice 2.1

leplatrem:
r? @Natim


leplatrem:
Pass `schema` as instance instead of class. Related to https://github.com/Cornices/cornice/pull/415

PR #911, opened 2016-11-10
Force server url admin

leplatrem:
Will require https://github.com/Kinto/kinto-admin/pull/310 to be released before it would work, but does not harm

r? @n1k0


leplatrem:
screenshot from 2016-11-10 15-42-07

Natim:
When I see your screenshot I remember that we want to be able to remove FxA and to rename Basic auth in LDAP for the kinto-writer.

I don't know if we can use plugin to handle authorization policies there.

leplatrem:
Please open an issue in kinto-admin to see what can be done there

PR #914, opened 2016-11-11
Permissions endpoint expose settings (fixes #909)

### LINKED ISSUE #909: Permissions endpoint does not expose perms specified in settings
Ex:
- Set `kinto.bucket_write_principals = system.Authenticated`
- `GET /permissions` should mention every buckets

Related #892 #350

leplatrem:
Fixes #909

- [x] basic implementation
- [ ] polish implem (eg. request in bucket loop)

### DIFF
--- a/kinto/views/permissions.py
+++ b/kinto/views/permissions.py
@@ -1,5 +1,6 @@
 import colander
 from pyramid.security import NO_PERMISSION_REQUIRED, Authenticated
+from pyramid.settings import aslist
 
 from kinto.authorization import PERMISSIONS_INHERITANCE_TREE
 from kinto.core import utils as core_utils, resource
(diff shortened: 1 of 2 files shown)


leplatrem:
@Natim for this patch to work efficiently we would need to *scan* every objects parent id (the same as webpush?).

For example:
- Give `collection_read_principals = system.Authenticated`
- Query the `/permissions` endpoint as any authenticated user
- You would expect to have the read permission on every collections (in every buckets)

For retrieving the list of every collections, if we use `get_all()` we don't know in which bucket they are (ie. we don't know the value of `parent_id`). We still can first iterate on buckets and then on their collections, but this is sub-optimal (*that's the current implementation of this patch*).

PR #923, opened 2016-11-17
Update collections.rst

joelmarty:
added an explicit mention to the fact that data may be used to store client-related fields


leplatrem:
Thanks for the precision!

PR #929, opened 2016-11-21
Fix failing tests on master

leplatrem:
* The charset was removed from content-type response (?)
* Transaction 2.0 is not fully supported by pyramid_tm: see https://github.com/Pylons/pyramid_tm/pull/52


Natim:
Apparently tagging transaction < 2 is not sufficient anymore. We need to also pin the `pyramid_tm` version too.

PR #935, opened 2016-11-22
Delete children permissions when deleting parent (fixes #898)

### LINKED ISSUE #898: The permissions endpoint returns deleted collections
In the stage kinto-writer while there are no collection in `/buckets/staging/collections` named `add-ons` the `/permissions` endpoint returns

```
        {
            "bucket_id": "staging", 
            "collection_id": "add-ons", 
            "id": "add-ons", 
            "permissions": [
                "read", 
                "read:attributes"
            ], 
            "resource_name": "collection", 
            "uri": "/buckets/staging/collections/add-ons"
        }, 
```

Is it because the permissions where not correctly removed when the bucket was deleted?

leplatrem:
Fixes #898

r? @Natim

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,15 @@ This document describes changes between each past release.
 5.0.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+**Bug fixes**
+
+- Permissions are now correctly removed from permission backend when a parent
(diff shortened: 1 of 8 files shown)


Natim:
Can we also add tests to make sure it works with records and groups when you delete the buckets and with records when you delete the collection?

PR #937, opened 2016-11-22
Expose the basicauth capability if activated.

### LINKED ISSUE #934: Expose a dedicated capability when basicauth is enabled
Refs https://github.com/Kinto/kinto-admin/issues/331

We need to know if the basicauth authentication method is available on a given server. We could detect if the server is configured with basicauth enabled and expose a dedicated capability in the hello endpoint; wdyt?

Natim:
Fixes #934 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem

### DIFF
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -101,6 +101,15 @@ def setup_authentication(config):
     from configuration.
     """
     config.include('pyramid_multiauth')
+    settings = config.get_settings()
+
+    policies = aslist(settings['multiauth.policies'])
+    if 'basicauth' in policies:
+        config.add_api_capability(
+            "basicauth",
+            description="Very basic authentication sessions. Not for production use.",
+            url="http://kinto.readthedocs.io/en/stable/api/1.x/authentication.html",
+        )
 
     # Track policy used, for prefixing user_id and for logging.
     def on_policy_selected(event):
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -105,11 +105,7 @@ def setup_authentication(config):
 
     policies = aslist(settings['multiauth.policies'])
     if 'basicauth' in policies:
-        config.add_api_capability(
-            "basicauth",
-            description="Very basic authentication sessions. Not for production use.",
-            url="http://kinto.readthedocs.io/en/stable/api/1.x/authentication.html",
-        )
+        config.include('kinto.core.authentication')
 
     # Track policy used, for prefixing user_id and for logging.
     def on_policy_selected(event):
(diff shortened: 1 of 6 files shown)


kinto/core/initialization.py
```diff
@@ -101,6 +101,15 @@ def setup_authentication(config):
     from configuration.
     """
     config.include('pyramid_multiauth')
+    settings = config.get_settings()
+
+    policies = aslist(settings['multiauth.policies'])
+    if 'basicauth' in policies:
```
Natim: The limitation is if somebody configure another policy with the same name, but it is a edge case we can ignore I guess.

Natim:
n1k0-basicauth-capability

tests/core/test_authentication.py
```diff
@@ -12,17 +12,26 @@ class AuthenticationPoliciesTest(BaseWebTest, unittest.TestCase):
...
 
     def test_basic_auth_is_declined_if_disabled_in_settings(self):
         app = self.make_app({
             'multiauth.policies': 'dummy',
             'multiauth.policy.dummy.use': ('pyramid.authentication.'
                                            'RepozeWho1AuthenticationPolicy')})
         app.get(self.collection_url, headers=self.headers, status=401)
+        # Check that the capability is exposed on the homepage.
+        resp = self.app.get('/')
+        assert 'basicauth' in resp.json['capabilities']
```
Natim: `assert not in`

kinto/core/initialization.py
leplatrem: Couldn't we have `includeme(config)` function in the authentication module instead? 

See https://github.com/mozilla-services/pyramid_multiauth/blob/master/pyramid_multiauth/__init__.py#L220

tests/core/test_views_hello.py
```diff
@@ -78,7 +81,11 @@ def test_capabilities_can_be_specified_via_config(self):
 
         response = app.get('/')
         capabilities = response.json['capabilities']
-        expected = {"cook-coffee": {"description": "", "url": ""}}
+        expected = {
+            "cook-coffee": {"description": "", "url": ""},
+            "basicauth": {
+                'description': 'Very basic authentication sessions. Not for production use.',
+                'url': 'http://kinto.readthedocs.io/en/stable/api/1.x/authentication.html'}}
```
leplatrem: I'm not super fan about the duplication of the basic auth capability here.

I guess we could just set `multiauth.policies` to an empty value for this test

kinto/core/initialization.py
Natim: That's a really interesting question, this mean adding it in kinto.plugins as well which doesn't feel right to me, I can see multiple way of activating the `include()` while disabling it in the `multiauth.policies` which is why I didn't took that path.

Natim: Seems interesting

leplatrem: Hmm, but then the policy name (userid prefix) won't be `basicauth`

leplatrem: just 
```py
if 'basicauth' in policies:
    config.include('kinto.authentication')
```
to at least keep the description along the implementation?

PR #940, opened 2016-11-24
Update Scalingo pricings.

Natim:
r? @magopian


Natim:
Oups @EtienneM was faster.

Issue #894, opened 2016-10-28
Principals from settings not taken into account when hiding permissions


leplatrem:
Since #837, the permissions are empty if the user has no `write` permission.

But the permissions from settings are not taken into account.

For example, with this conf:

```ini
# kinto.ini
kinto.bucket_create_principals = system.Authenticated
kinto.bucket_write_principals = system.Authenticated
```

Someone creates a bucket:

```http
$ http PUT :8888/v1/buckets/her --auth token:alice
HTTP/1.1 201 Created
Access-Control-Expose-Headers: Retry-After, Content-Length, Alert, Backoff
Content-Length: 154
Content-Type: application/json; charset=UTF-8
Date: Fri, 28 Oct 2016 15:27:25 GMT
Etag: "1477668445792"
Last-Modified: Fri, 28 Oct 2016 15:27:25 GMT
Server: waitress

{
    "data": {
        "id": "her", 
        "last_modified": 1477668445792
    }, 
    "permissions": {
        "write": [
            "basicauth:a06c2d0d82a192b5afab7695eb1159b75dd5a7732caa442db624e5c7c334bcb6"
        ]
    }
}
```

Because of the setting, every can write. However:

```http
$ http GET :8888/v1/buckets/her --auth token:bob  
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, ETag, Pragma, Cache-Control, Backoff
Cache-Control: no-cache, no-store
Content-Length: 68
Content-Type: application/json; charset=UTF-8
Date: Fri, 28 Oct 2016 15:27:35 GMT
Etag: "1477668445792"
Last-Modified: Fri, 28 Oct 2016 15:27:25 GMT
Server: waitress

{
    "data": {
        "id": "her", 
        "last_modified": 1477668445792
    }, 
    "permissions": {}
}

```

Once bob touches the object, a `write` permission is added, and he can see the permissions:

```http
$ http PUT :8888/v1/buckets/her --auth token:bob
```

Then only he can see them:

```http
$ http GET :8888/v1/buckets/her --auth token:bob
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, ETag, Pragma, Cache-Control, Backoff
Cache-Control: no-cache, no-store
Content-Length: 231
Content-Type: application/json; charset=UTF-8
[...]

Issue #892, opened 2016-10-28
Show permissions principals from settings in JSON responses


leplatrem:
For better consistency, permissions principals specified via settings could be presented in the response JSON along others.

Natim:
We hit that bug today when trying to remove records from a user default_bucket, they didn't show in the admin

PR #946, opened 2016-11-25
Refactor handling of prefixed user id among request principals

leplatrem:
Some background ground for #892 and #894 

r? @Natim

### DIFF
--- a/kinto/core/testing.py
+++ b/kinto/core/testing.py
@@ -34,9 +34,9 @@ class DummyRequest(mock.MagicMock):
         self.authn_type = 'basicauth'
         self.prefixed_userid = 'basicauth:bob'
         self.effective_principals = [
-            self.prefixed_userid,
             'system.Everyone',
-            'system.Authenticated']
+            'system.Authenticated',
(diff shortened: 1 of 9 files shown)


kinto/core/testing.py
```diff
@@ -34,9 +34,9 @@ def __init__(self, *args, **kwargs):
         self.authn_type = 'basicauth'
         self.prefixed_userid = 'basicauth:bob'
         self.effective_principals = [
-            self.prefixed_userid,
             'system.Everyone',
-            'system.Authenticated']
+            'system.Authenticated',
+            'bob']
```
Natim: Any idea why you want to change the behavior here? Is there a security risk not to have the prefix in the principal here.

leplatrem: This `effective_principals` value truely respects what Pyramid would return. The previous list was wrong, there was no way we could obtain a prefixed_userid in the principals with the previous code.

Natim: Ok I'll trust you on that :)

PR #953, opened 2016-11-29
Add limit for cache size for the memory backend.

### LINKED ISSUE #610: Add limit for cache size (especially memory)
original: https://github.com/mozilla-services/cliquet/issues/406

That's an improvement so we control the cache size. Otherwise it's a free wheel that eventually crashes the server. 

Which makes me think: if I do a load test on stage today, is the kinto process growing indefinitely in memory ?

Natim:
Fixes #610

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem

### DIFF
--- a/tests/core/test_cache.py
+++ b/tests/core/test_cache.py
@@ -1,5 +1,6 @@
 import mock
 import unittest
+import time
 
 from kinto.core.utils import sqlalchemy
 from kinto.core.cache import (CacheBase, memory as memory_backend,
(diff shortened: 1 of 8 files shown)


tests/core/test_cache.py
```diff
@@ -51,6 +53,33 @@ def test_ping_returns_false_if_unavailable(self):
...
+        assert 'foobar' in self.cache._ttl
+        assert 'foobar' in self.cache._created_at
+        time.sleep(0.02)
+        retrieved = self.cache._clean_expired()
+        assert 'foobar' not in self.cache._store
+        assert 'foobar' not in self.cache._ttl
+        assert 'foobar' not in self.cache._created_at
+        self.assertIsNone(retrieved)
+
+    def test_clean_oversized_expires_items(self):
```
Natim: Maybe we can just get rid of this one.

Issue #955, opened 2016-11-29
Use os.urandom rather than random.choice or random.random


Natim:
```
>> Issue: [B311:blacklist] Standard pseudo-random generators are not suitable for security/cryptographic purposes.
   Severity: Low   Confidence: High
   Location: kinto/views/__init__.py:15
14	        letters = [random.choice(ascii_letters + string.digits)]
15	        letters += [random.choice(alphabet) for x in range(7)]
16	        return ''.join(letters)
```

```
>> Issue: [B311:blacklist] Standard pseudo-random generators are not suitable for security/cryptographic purposes.
   Severity: Low   Confidence: High
   Location: kinto/views/__init__.py:14
13	        alphabet = ascii_letters + string.digits + '-_'
14	        letters = [random.choice(ascii_letters + string.digits)]
15	        letters += [random.choice(alphabet) for x in range(7)]
```

```
>> Issue: [B311:blacklist] Standard pseudo-random generators are not suitable for security/cryptographic purposes.
   Severity: Low   Confidence: High
   Location: kinto/core/storage/__init__.py:262
261	            else:
262	                if random.random() < _HEARTBEAT_DELETE_RATE:
263	                    backend.delete_all(_HEARTBEAT_COLLECTION_ID,
```

```
>> Issue: [B311:blacklist] Standard pseudo-random generators are not suitable for security/cryptographic purposes.
   Severity: Low   Confidence: High
   Location: kinto/core/cache/__init__.py:87
86	        try:
87	            if random.random() < _HEARTBEAT_DELETE_RATE:
88	                backend.delete(_HEARTBEAT_KEY)
```

m-charlton:
I can do this

Natim:
👍 Go ahead 💃

m-charlton:
Looks like I've caused a problem on one of the tests for Py3.4 - investigating

Natim:
It is because of a mock on ``kinto.core.storage.random.random``

m-charlton:
I see the problem, I will submit the fix tomorrow.

tusharmakkar08:
Is this issue closed?

Natim:
I think it is yes.

PR #971, opened 2016-12-01
Replaced insecure calls to random with more secure SystemRandom

m-charlton:
Fixes #

- [ ] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @

See - 'Use os.urandom rather than random.choice or random.random #955'
SystemRandom, according to the docs, calls os.urandom under the hood and
provides the same 'choice' and 'random' API calls that 'random' provides.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,9 @@ This document describes changes between each past release.
 - The ``delete_object_permissions()`` of the permission backend now supports
   URI patterns (eg. ``/bucket/id*``)
 - Refactor handling of prefixed user id among request principals
+- Replaced insecure use of ``random.random()`` and ``random.choice(...)`` with
+  more secure ``random.SystemRandom().random()`` and
+  ``random.SystemRandom().choice(...)``. (#955)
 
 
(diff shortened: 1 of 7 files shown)


m-charlton:
mocks now inject modified paths to the calls to `random` used in the implementation files.

Natim:
I tried that yesterday not sure it works.

m-charlton:
O.K - I can set multiple tests at my end or, come up with an alternative approach.

Natim:
Tests passes 💃 

Could you add a line in the Changelog in **Internal Changes**?

Thanks a lot :)

m-charlton:
Sure, will do. Apologies fro breaking the build earlier on.

Natim:
Don't apologies :) That's totally normal to break the build! It is there to break actually :)

Natim:
It looks good to me thanks :)

m-charlton:
No worries, I'm free to do some more fixes so I can get to know the 
code base better. Going to spend some time getting my test environment 
working, had some gyp with getting a postgres server woking correctly.

PR #972, opened 2016-12-01
Add a setting to exclude certain resources from being tracked by history (fixes #964)

### LINKED ISSUE #964: Add a configuration of collections that the history plugin needs to keep track on
Today the history plugin applies to all the collection but most of them don't need it.
For instance with the kinto-signer plugin we don't want to track history of changes in the preview and signed collection.
The same goes with the kinto-changes plugin when we don't want to track monitor changes modifications.

The same way we can configure the kinto-signer resources we want to track, we should be able to configure the list of collections we want the history plugin to track.

leplatrem:
Fixes #964 

r? @Natim

### DIFF
--- a/tests/plugins/test_history.py
+++ b/tests/plugins/test_history.py
@@ -681,3 +681,47 @@ class PermissionsTest(HistoryWebTest):
                             headers=self.headers)
         entries = [e['resource_name'] == 'history' for e in resp.json["data"]]
         assert not any(entries)
+
+
+class ExcludeResourcesTest(HistoryWebTest):
+
+    def get_app_settings(self, extras=None):
+        settings = super(ExcludeResourcesTest, self).get_app_settings(extras)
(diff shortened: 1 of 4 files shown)


tests/plugins/test_history.py
```diff
@@ -681,3 +681,47 @@ def test_history_entries_are_not_listed_in_permissions_endpoint(self):
...
+                            headers=self.headers)
+        entries = resp.json['data']
+        assert len(entries) > 0
+
+        resp = self.app.get('/buckets/b/history?collection_id=a',
+                            headers=self.headers)
+        entries = resp.json['data']
+        assert len(entries) == 0  # nothing.
+
+    def test_some_specific_object_can_be_excluded(self):
```
Natim: Should we add a test for records too?

leplatrem: There's no reason it wouldn't work as groups, no?

Issue #880, opened 2016-10-25
Validate json patch body with Colander


leplatrem:
Currently:
https://github.com/Kinto/kinto/blob/b777a0191a5b317c6f8d0b15efe9e1bbf2b63b1d/kinto/core/resource/__init__.py#L108-L109

Validation is performed in the underlying lib that applies json patch:
https://github.com/Kinto/kinto/blob/b777a0191a5b317c6f8d0b15efe9e1bbf2b63b1d/kinto/core/resource/__init__.py#L646-L649

Issue #873, opened 2016-10-20
Validate headers and querystring with cornice schemas


leplatrem:
Now that colander allows to define clear schemas for querystring and body, we could leverage it to validate querystrings (eg. `_since=wrong`) and headers (eg `If-Match: z`) values instead of our custom code:
- https://github.com/Kinto/kinto/blob/4.3.1/kinto/core/resource/__init__.py#L746-L751
- https://github.com/Kinto/kinto/blob/4.3.1/kinto/core/resource/__init__.py#L939-L951

Requires #790

PR #977, opened 2016-12-08
Add OpenAPI/Swagger spec (fixes #631)

### LINKED ISSUE #631: Generate Swagger API declaration
We talked about it at Djangocong, it would be really good to add automatic functional tests and documentation.

Refs https://github.com/mozilla-services/cliquet/issues/257

gabisurita:
Fixes #631 

Features:

- OpenAPI documentation resource at `/v1/swagger.json`
- Works with [Swagger codegen](https://github.com/swagger-api/swagger-codegen) and   [Bravado](https://github.com/Yelp/bravado)  automated client generators.
- It's compatible with   [Swagger UI](http://swagger.io/swagger-ui/) and   [Apiary](https://apiary.io/) interactive documentation generators. You can try them on  [Swagger Hub](https://app.swaggerhub.com/api/gabisurita/kinto/1.13)  or [Apiary.io](http://docs.kinto.apiary.io/) (although working with basicAuth there is tricky).
- Tests for semantic and logic using [bravado_core](https://github.com/Yelp/bravado-core).

Limitations:

- No support for schema free fields.
  - Treated like extensions for code generators (may raise warnings on other utilities)
  - No filtering
  - No field selection
  - No validation on collection schema
- No validation on OR required clauses (ex: provide data or permissions)
- No validation for fields that are required on response, but not on request
  (ex: id doesn't show as a required field), doing so would result in two different
  object definitions for each object, which would add unnecessary complexity to the clients IMO)
- Backoff headers

Pending:

- [x] Investigate memory usage on py27-raw TOX environment.
- [x] Allow setting the spec on configuration.
- [x] Allow plugins to register themselves to the spec.
- [x] Use YAML instead of JSON.
- [x] Allow other authentication methods. (may be defined as regular plugins)
- [x] Add Custom Headers (If-Match/If-None-Match/Etag)
- [x] Add utility tests
- [x] Investigate ways to test for coverage (we test all operations with `test_resources`)
- [x] Add more descriptions to the spec.
- [x] Add batch validation tests
- [x] Upgrade documentation.
- [x] Add a changelog entry.
- [x] Update the API changelog.

r? @glasserc @Natim @leplatrem


delijati:
There are two ways of creating a valid json api with openapi (swagger) support:

1.) swagger.yaml (use yaml, we need to read it) as master + some validation [1]
2.) cornice + colander as master and create a swagger.yaml file [2]

Every way has its advantages and disadvantages.

[1] https://github.com/pipermerriam/flex or https://github.com/striglia/pyramid_swagger
[2] https://github.com/Cornices/cornice.ext.swagger

gabisurita:
@delijati, thank you for the feedback!

1) When you say validation you're talking about validating every request/response against the spec or validating the spec itself? Unfortunately, from what I've tested, pyramid swagger request/response validation isn't suitable for all use cases (ex: extra parameters with any type on an object body) and it didn't detect `kinto.core` view paths. I'm also not sure about adding another validation layer. If you say about validating the spec itself, why not use bravado_core instead of flex or pyramid_swagger?

2) I'm not sure using cornice.ext.swagger is feasible, but I think it might add considerable complexity to the API code, which was something I want to avoid.

About YAML instead of JSON, I agree. It is listed as pending.

delijati:
@gabisurita 
  
1. Yes every request/ response against the spec. I looked ``bravado_core`` up, they seem also to 
     provide validation of requests and responses. (Sorry i assumed you have no validation at all)

2. ``cornice.ext.swagger`` is the other way (generating swagger.yaml) from ``colander`` schema. That 
    makes only sense if you already use or have plans to use ``cornice``. I'm also not sure if the 
    generated ``swagger.yaml`` is feasible for yours needs.

For the ``YAML`` part, ``swagger2.0`` understands json, yaml:

```python
class YamlRendererFactory(object):
    def __init__(self, info):
        pass

    def __call__(self, value, system):
        response = system['request'].response
        response.headers['Content-Type'] = 'application/x-yaml; charset=UTF-8'
        return yaml.dump(value).encode('utf-8')

def includeme(config):
    config.add_renderer(
        'yaml', 'pyramid_swagger.api.YamlRendererFactory'
    )
```

gabisurita:
@delijati i've used validation of requests/responses just on tests. I think using it in production doesn't make much sense (as we already validate them using colander). The idea is only to ensure that that spec is correct, not use it for validation.

delijati:
As you mentioned that kinto is already using ``cornice`` i integrated ``cornice.ext.swagger`` (with a minor patch) and the resulting ``swagger.json`` looks promising.
[swagger_json.txt](https://github.com/Kinto/kinto/files/643456/swagger_json.txt)

delijati:
@gabisurita I hacked a small plugin together, check it out if you like.

```bash
$ pip install git+https://github.com/delijati/kinto_swagger.git
$ vim config/kinto.ini
kinto.includes=kinto_swagger

$ firefox http://0.0.0.0:8888/v1/swagger 
```

gabisurita:
@delijati I've tried the plugin, but got a `500` on GET   `/v1/swagger_json`. Do you have a branch where you've been testing this? It would be easier to reproduce on my side.

Here's the traceback:

```
2016-12-10 11:00:06,578 ERROR [kinto.core.views.errors][waitress] "GET   /v1/swagger_json" ? (? ms) unbound method get() must be called with BatchRequest instance as first argument (got str instance instead) agent=HTTPie/0.9.4 authn_type=None errno=None exception=Traceback (most recent call last):
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/pyramid/tweens.py", line 22, in excview_tween
    response = handler(request)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 119, in tm_tween
    reraise(*exc_info)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/pyramid_tm/__init__.py", line 98, in tm_tween
    response = handler(request)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/pyramid/router.py", line 158, in handle_request
    view_name
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/pyramid/view.py", line 547, in _call_view
    response = view_callable(context, request)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/pyramid/viewderivers.py", line 442, in rendered_view
    result = view(context, request)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/pyramid/viewderivers.py", line 147, in _requestonly_view
    response = view(request)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/kinto_swagger/__init__.py", line 20, in swagger_json
    base_path=base_path, head=False)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/cornice_swagger/swagger.py", line 174, in generate_swagger_spec
    parameters, definition = schema_to_parameters(schema, service)
  File "/home/gabi/outreachy/kinto/.venv/lib/python2.7/site-packages/cornice_swagger/swagger.py", line 22, in schema_to_parameters
    location_schema = schema.get(location)
TypeError: unbound method get() must be called with BatchRequest instance as first argument (got str instance instead) lang=None uid=None
```

delijati:
@gabisurita i tried it on the ``master`` of ``kinto`` with only two plugins activated:

```bash
$ git clone https://github.com/Kinto/kinto.git
$ cd kinto
$ virtualenv env
$ env/bin/pip install -e .
$ env/bin/pip install git+https://github.com/delijati/kinto_swagger.git
$ kinto init
$ kinto migrate
$ vim config/kinto.ini
...
kinto.includes = kinto.plugins.default_bucket
                         kinto_swagger
...
$ kinto start
$ firefox http://0.0.0.0:8888/v1/swagger 
```
Btw. the error seams to be a older version of ``cornice_swagger``. I also fixed the path to the swagger js files in ``kinto_swagger``.

gabisurita:
Upgrading `cornice_swagger` did the work.
This looks great, but skimming through the docs I've got a few questions while following this approach:
 
1. How should responses be documented?
2. Patch requests should have a body, but that doesn't show on the documentation
3. How to document query strings?
4. Does it allows customizing the tags? I think putting all the operations under the tag `buckets` may be confusing, specially when using client generators.
5. How to set up authentication?
6. Finally it doesn't seems to generate valid swagger, but I understand it's just a sketch.

delijati:
*1. How should responses be documented?*

I think currently ``cornice`` does not support that @leplatrem ? Maybe this can be added as docstrings and extracted from there?

*2. Patch requests should have a body, but that doesn't show on the documentation*

Maybe the ``body`` schema is not defined right see 3. 

*3. How to document query strings?*

That is a ``cornice`` feature and they are extracted. (``cornice`` docs "Schema validation")

```python
class Query(MappingSchema):
    yeah = SchemaNode(String())
    mau = SchemaNode(String())

class RequestSchema(MappingSchema):
    body = Body(description="Defines a cornice body schema")
    querystring = Query()
```

*4. Does it allows customizing the tags? I think putting all the operations under the tag buckets may be confusing, specially when using client generators.*

Currently it only does ``tag_name = service.path.split("/")[1]`` so sure this can be improved somehow.

*5. How to set up authentication?*

You mean this? https://github.com/swagger-api/swagger-ui#header-parameters

*6. Finally it doesn't seems to generate valid swagger, but I understand it's just a sketch.*

Never checked that ;)

gabisurita:
@delijati, great! We may wait for @glasserc, @Natim and @leplatrem to discuss this,but for now my impressions are:

1. I think documenting responses with docstrings might be too much information on them. I think we should find a better option.
2. I'm not sure about that, it would need proper investigation.
3. The problem is that as far as I know the querystrings are not validated with `cornice` schemas, but on `kinto.core.resource`.
4. Great.
5. I mean `securityDefinitions` and `/path/{path}/{method}/security` fields on the swagger document.

glasserc:
Sorry I wasn't able to look at this for so long!

- I haven't had a chance to look at the memory issue you encountered. I'm glad you found a workaround. Maybe that's good enough, or maybe we should dig deeper?
- The tests look very thorough. Testing the definitions in isolation is a great idea.
- Adding a new kind of request means adding a new test to the swagger tests. Maybe it would be cool if there were a way to avoid this? For example, maybe we could have a test that just makes every request possible. Or is it at least possible to verify that all requests in the spec were exercised by the test suite, and fail if one wasn't? (This might not work for error responses, which I think we still have to write out explicitly.)
- If it makes sense to use `cornice_swagger`, then go ahead, but if it doesn't suit our use cases, then I don't want to spend a lot of time trying to force it to fit. In our meeting today, you mentioned that our Colander schemae are actually shared, so probably not a good match for a user-facing schema. In addition, some of our API involves custom headers, which I bet aren't captured in any Colander schema. Both of these seem like good reasons why `cornice_swagger` wouldn't fit.
- @gabisurita proposed making Swagger stuff into a plugin. @Natim didn't have any objections. @leplatrem ? We agreed that we should proceed as-is (i.e. in this repo) for now, since moving it should be easy once it exists. I think it should stay in this repo as long as we're maintaining it manually.
- @gabisurita asked where the Swagger view should live (`kinto` or `kinto.core`). I think that as long as it's being maintained manually, it should live in `kinto`, given that the views that are documented are there too. If we come up with some way of generating it automatically, it could live in `kinto.core` so that projects "inheriting" from `kinto.core` could get Swagger support too. What do you think @Natim @leplatrem ?

Custom headers like If-Match and If-Not-Match weren't there when I looked at this earlier, but I see that they're now. In a meeting today, @gabisurita said that support for headers has just landed in Swagger 2.0.

tests/swagger/support.py
```diff
@@ -0,0 +1,74 @@
...
+resources = build_resources(spec)
+
+
+class SwaggerTest(BaseWebTest, unittest.TestCase):
+
+    def setUp(self):
+        # FIXME: solve memory issues from generating the spec multiple times
+        self.spec_dict = spec_dict
+        self.spec = spec
+        self.resources = resources
```
leplatrem: If you want to instantiate the spec only once for the whole test suite, you can try to use `setUpClass()` instead of `setup()` https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUpClass

gabisurita: Great! Thanks!

leplatrem:
>> 1. How should responses be documented?
>>
> I think currently cornice does not support that @leplatrem ? Maybe this can be added as docstrings and extracted from there?
>

Indeed, there's nothing in Cornice right now. We could add extra attributes on Cornice services to specify the responses schemas.

As you noted already, in Kinto we still have some custom validation that should be moved to Colander schemas (see #873 #880).

> If it makes sense to use cornice_swagger, then go ahead, but if it doesn't suit our use cases, then I don't want to spend a lot of time trying to force it to fit.

We can keep this as a long term goal indeed, and use that use-case to help the library improve until we obtain an output that is good enough. (Be aware that some do not agree on the automated approach: https://larlet.fr/david/blog/2016/specifications-apis/)

> @gabisurita proposed making Swagger stuff into a plugin.

I would strongly suggested to keep it in the same repo. Especially because the spec must evolve very closely with the code. And the tests should break if the spec does not match the resulting API (and vice-versa :)). It could be built-in plugin but what are the arguments in favor of running the API without it?

> If we come up with some way of generating it automatically, it could live in kinto.core so that projects "inheriting" from kinto.core could get Swagger support too. 

Generating would be the goal of `cornice_swagger` I guess. The view itself does not contain anything specific to kinto. The .yml file does. So it's basically like `GET /__version__`. A project made with `kinto.core` can provide a swagger definition and the view serves it. So I don't have any objection with having it in `kinto.core`, but it may require extra efforts to have a clean separation with regards to testing (*or maybe not you already did everything the right way*)

I don't remember its name, but there's a Pyramid project that makes sure your requests/responses match the specifications you expose. That's something we could investigate to make sure our spec remains up-to-date with the code...

Awesome contribution again :) 👏

tests/swagger/support.py
gabisurita: I've tried deallocating them in the `tearDown()` method and leaving it for garbage collecting but it didn't work. Maybe this works different on pytest, but I guess there should be no problems in sharing the spec across the tests. I can't think of a test case we would need to modify these objects.

swagger.yaml
```diff
@@ -135,6 +143,10 @@ paths:
             type: object
         '404':
           description: version.json file not found.
+          schema:
+            $ref: '#/definitions/Error'
+        default:
+          $ref: '#/responses/default'
```
gabisurita: Yes, I'm sorry. I should have split this into different commits.

gabisurita:
> I would strongly suggested to keep it in the same repo. Especially because the spec must evolve very closely with the code. And the tests should break if the spec does not match the resulting API (and vice-versa :)). It could be built-in plugin but what are the arguments in favor of running the API without it?

I agree on keeping it in the same repo. I can't think of an use case were someone would like to get it disabled. On the current implementation if someone really wants to do it simple deleting file or removing it with `pkg_resources` would do the work. Also, should we show it as a capability?

> Generating would be the goal of cornice_swagger I guess. The view itself does not contain anything specific to kinto. The .yml file does. So it's basically like GET /__version__. A project made with kinto.core can provide a swagger definition and the view serves it. So I don't have any objection with having it in kinto.core, but it may require extra efforts to have a clean separation with regards to testing (or maybe not you already did everything the right way)

Keeping the spec as is think we could just move `test_view_swagger.py` to `kinto.core` and keep `tests_swagger` on the root of the tests. Also, imaging an use case like this we could also split the current spec into `kinto.core` and `kinto` specs and merge them during generation. What do you think?

> I don't remember its name, but there's a Pyramid project that makes sure your requests/responses match the specifications you expose. That's something we could investigate to make sure our spec remains up-to-date with the code.

The one I'm aware of is `pyramid_swagger`. Unfortunately it's not suitable for some use cases because of limitations on the spec. There's no way to define a schema with additional attributes without a fixed type, which means requests/responses of objects with additional attributes may fail. As I as commenting with @glasserc on our meeting today, we are testing some limits of the OpenAPI documentation. There are also some dubious definitions on the spec that fail `bravado_core` validation but work with documentation and code generation resources, which means I.M.O we can't trust `bravado_core` validation as our only source of truth.

Curiosity: `pyramid_swagger` user `bravado_core` for validation

leplatrem:
> Also, should we show it as a capability?

We keep capabilities for API features that can be disabled/enabled from settings. So, I would say that no capability would be required (like for `GET __version__`)

>  imaging an use case like this we could also split the current spec into kinto.core and kinto specs and merge them during generation. What do you think?

That's a nice idea! We could also apply the same strategy when plugins are included (e.g `config.add_openapi_entry(...)`). But let's keep that for another round, the scope of this PR is good enough IMO

> The one I'm aware of is pyramid_swagger. Unfortunately it's not suitable for some use cases because of limitations on the spec.

Take it easy, you'll even have time to contribute to the spec ;)

kinto/core/views/swagger.py
```diff
@@ -64,11 +64,23 @@ def swagger_view(request):
...
     schemes = [settings.get('http_scheme') or 'http']
 
+    # Default security match all security defs
+    auth_types = swagger_view.__json__.get('securityDefinitions', {})
+    security = swagger_view.__json__.get('security', [])
+
+    for name, prop in auth_types.items():
+        scopes = prop.get('scopes', [])
+        # Drop scope descriptions
+        security_def = {name: [scope.keys()[0] for scope in scopes]}
```
glasserc: Why just the first scope?

tests/core/test_views_swagger.py
```diff
@@ -37,3 +37,25 @@ def test_extensions(self):
...
+        content = {
+            'securityDefinitions': {
+                'fxa': {
+                    'type': 'oauth2',
+                    'authorizationUrl': 'https://oauth-stable.dev.lcip.org',
+                    'flow': 'implicit',
+                    'scopes': [{'kinto': 'Basic scope'}]
+                }
+            }
+        }
```
glasserc: I don't really understand the value of this change. It seems like in this test, the `type` would be something that a client might want to know?

Natim:
> We keep capabilities for API features that can be disabled/enabled from settings.

Yes so if this feature lands as a default plugins we should probably have a capability.

tests/swagger/test_resources.py
```diff
@@ -1,104 +1,44 @@
...
+from .support import SwaggerTest
 
 
-class SwaggerResourcesTest(SwaggerTest):
+class MetaResourcesTest(type):
+    """Metaprogram test cases for easy debugging."""
 
-    allowed_failures = ['version']
+    def __init__(cls, *args, **kargs):
+        unittest.case.TestCase = cls
```
glasserc: I just noticed this.. what's going on here?

tests/swagger/test_resources.py
```diff
@@ -1,104 +1,44 @@
...
-                requests=requests,
-                defaults=defaults
-            )
-            self.validate_request_call(op)
-
-    def test_resource_buckets(self):
-        resource = self.resources['Buckets']
+                if op_id not in cls.allowed_failures:
+                    setattr(cls, 'test_resource_%s' % op_id,
+                            lambda self: cls.validate_request_call(self, op))
```
glasserc: So what I was thinking when I suggested this was something a little different:

```python
def operation_test_method(op):
    def test_method(self):
        pass   # basically "validate_request_call"
    return test_method

def resource_test_class(resource):
    attrs = {}
    for op_id, op in resource.operations.items():
        attrs["test_{}".format(op_id)] = operation_test_method(op)
    return type("{}Test".format(resource.name), (SwaggerResourceTest,), attrs)

BucketsTest = resource_test_class(resources['Buckets'])   # or whatever the syntax is
CollectionsTest = resource_test_class(resources['Collections'])    # etc.
```

This approach means you don't have to explicitly handle metaclasses, which could be easier to read. Another approach might be to use a class decorator:

```python
def operation_test_method(op):
    def method(self):
        pass  # again, basically validate_request_call

def resource_test_case(resource):
    def decorator(cls):
        for op_id, op in resource.operations.items():
            setattr(cls, "test_{}".format(op_id), operation_test_method(op))
        return cls

    return decorator

@resource_test_case(resources['Buckets'])
class BucketsTest(SwaggerTest):
    pass
```

But having reread this a bunch of times, I think the hardest part to read was just the diff. Reading the new file, it isn't that confusing, except for a couple of quirks in how I think you are interfacing with Bravado.

tests/swagger/test_resources.py
gabisurita: This is the weirdest part, I think. We are forcing pytest to look at the Worker class while running this TestCase, without it it would just use the collected version. Whick means it doesn't run any test. This is the main motivation to use a metaclass. See http://doc.pytest.org/en/latest/example/pythoncollection.html and http://stackoverflow.com/questions/2798956/python-unittest-generate-multiple-tests-programmatically

tests/swagger/test_resources.py
```diff
@@ -1,104 +1,44 @@
...
-        for op in resource.operations.values():
-            self.setUp()
-            self.request.path = {
-                'bucket_id': 'b1',
-                'collection_id': 'c1',
-                'record_id': 'r1',
-            }
-            self.request.json = lambda: MINIMALIST_RECORD
-            self.validate_request_call(op)
+        request = self.request.__dict__
```
gabisurita: I will patch this and I think I'll make `self.request` always a dict. I couldn't understand why, but `marshal_param` expects the request to be a dict while `validate_request` expects it to be a `IncomingRequest` object with attributes matching the fields used on `marshal_param` request dict. My guess is that  `IncomingRequest` was implemented later and they didn't want to change the param handling part of the code. I was using `validate_request` on the last version of test resources, but with it we needed to specify where each parameter would go by hand (e.g. body, path, header) now we are using only `marshal_param` to validate and set the parameters, which actually looks for the param specification and see where each param should go.

tests/core/test_views_swagger.py
gabisurita: This test case (in comparison to the last one) checks if it's possible to change the default security (which means what the resources require as authentication) by only updating only `securityOptions` on an extension. It matches what happens on [this part](https://github.com/gabisurita/kinto/blob/775b67d6c78ae011286946f53bd99af0602b1e57/kinto/core/views/swagger.py#L67). I didn't test for type in this case because it should be the same for `test_extensions`.

docs/api/1.x/openapi.rst
```diff
@@ -0,0 +1,62 @@
...
+#####################
+
+The `OpenAPI Specification <https://github.com/OAI/OpenAPI-Specification>`_
+(formally known as Swagger specification)
+is a standard to describe REST APIs in a human and computer readable format
+and allow several features like interactive documentation and automated
+client generation.
+
+.. important::
+    OpenAPI 2.0 currently doensn't some features that are present on Kinto API.
```
Natim: > OpenAPI 2.0 currently doensn't some features that are present on Kinto API.

Do you mean:

> OpenAPI 2.0 currently doesn't support some features that are present on Kinto API.

docs/api/1.x/openapi.rst
```diff
@@ -0,0 +1,62 @@
...
+    Some known limitations are
+    lack of validation on or clauses (e.g. provide `data` or `permissions` in PATCH
+    operations), no support for non schema querystrings, as used on :ref:`filtering <filtering>`
+    with additional fields, and no optional response headers, as required for
+    `Backoff signs`<http://kinto.readthedocs.io/en/stable/api/1.x/backoff.html>`.
+    Also, :ref:`collection defined schemas <collection-json-schema>`,
+    are not validated, So if you're using client
+    generation, you may need to implement these features.
+
```
Natim: Can you add a note on how you choose to handle this limitations in the SWAGGER file?

docs/api/1.x/openapi.rst
```diff
@@ -0,0 +1,62 @@
...
+``kinto.swagger_file`` entry on your configuration file.
+
+OpenAPI extensions
+==================
+
+By default all includes on the configuration file will be prompt for a
+``swagger.yaml`` file on the package root containing an OpenAPI extension
+for the main specification. You may use it to include additional resources
+that the plugin provides or override the original definitions. You may disable
+Swagger extensions by setting ``kinto.swagger_extensions`` to ``False`` on the
```
Natim: I would rather have a default plugin then rather than yet another way to activate / deactivate features.

Natim: Note that if you can deactivate it we probably need a capability for it.
Any reason why you want to be able to deactivate it?

kinto/core/views/swagger.py
```diff
@@ -64,15 +64,20 @@ def swagger_view(request):
...
-    # Security definitions are JSON objects with a single key
+    # Security options are JSON objects with a single key
+    security = swagger_view.__json__.get('security', [])
     security_names = [next(iter(security_def)) for security_def in security]
 
-    # Include securityDefinitions that are not on default security
-    for name, prop in auth_types.items():
+    # include securityDefinitions that are not on default security options
+    for name, prop in security_defs.items():
         security_def = {name: prop.get('scopes', {}).keys()}
```
glasserc: I think some more comments would be helpful here to readers who don't already know much about the OpenAPI spec. How about: `The 'security' field lists all acceptable ways of authenticating, so no plugin should add something to 'securityDefinitions' without it being here too`, and then `A 'security' requirement is an object of the form {security_mechanism: ['list', 'of', 'oauth2', 'scopes']}  # or an empty list if oauth2 isn't used`.

requirements.txt
```diff
@@ -16,6 +16,7 @@ pyramid==1.7.3
 pyramid-multiauth==0.9.0
 pyramid-tm==1.1.1
 python-dateutil==2.6.0
+PyYAML==3.12
```
Natim: We should use `ruamel.yaml` rather than `PyYAML`

Natim: http://yaml.readthedocs.io/en/latest/pyyaml.html

Natim: And also because we had this discussion for kinto-wizard and we made this choice there: https://github.com/Kinto/kinto-wizard/blob/master/setup.py

setup.py
```diff
@@ -26,6 +27,7 @@ def read_file(filename):
     'jsonpatch',
     'python-dateutil',
     'pyramid_multiauth >= 0.8',  # User on policy selected event.
+    'PyYAML',
```
Natim: ruamel.yaml

kinto/swagger.yaml
```diff
@@ -0,0 +1,2183 @@
...
+      2. [Filtering](http://kinto.readthedocs.io/en/stable/api/1.x/filtering.html)
+      is supported on any field by using `?{prefix}{field_name}={value}`. But swagger
+      doesn't have support for patterned querystrings.
+      3. [Backoff headers](http://kinto.readthedocs.io/en/stable/api/1.x/backoff.html)
+      may occur with any response, but they are only present if the server is under
+      in heavy load, so we cannot validate them on every request. They are listed only
+      on the default error message.
+      4. [Collection schemas](http://kinto.readthedocs.io/en/stable/api/1.x/collections.html#collection-json-schema)
+      can be provided when defining a collection and should be a valid JSON Schema object,
+      but this is not validated by this specification.
```
glasserc: I think we should unify this list of limitations with the ones in the documentation, and provide a link to them.

docs/api/1.x/openapi.rst
```diff
@@ -0,0 +1,62 @@
...
+(formally known as Swagger specification)
+is a standard to describe REST APIs in a human and computer readable format
+and allow several features like interactive documentation and automated
+client generation.
+
+.. important::
+    OpenAPI 2.0 currently doensn't some features that are present on Kinto API.
+    Some known limitations are
+    lack of validation on or clauses (e.g. provide `data` or `permissions` in PATCH
+    operations), no support for non schema querystrings, as used on :ref:`filtering <filtering>`
```
glasserc: `, and provide request or default in /batch requests`

Also, I'd like to see a comment about whatever we ended up doing with additionalProperties and what bugs that fixes in what libraries, etc.

docs/api/1.x/openapi.rst
```diff
@@ -0,0 +1,62 @@
...
+    with additional fields, and no optional response headers, as required for
+    `Backoff signs`<http://kinto.readthedocs.io/en/stable/api/1.x/backoff.html>`.
+    Also, :ref:`collection defined schemas <collection-json-schema>`,
+    are not validated, So if you're using client
+    generation, you may need to implement these features.
+
+GET /swagger.json
+=================
+
+Returns the OpenAPI specification for the running instance of Kinto on JSON format.
```
glasserc: We should also make a note that this is the ONLY supported way to get the spec, since the running code does some transformations on the data, for example updating the host and listing the accepted security mechanisms.

docs/api/1.x/openapi.rst
```diff
@@ -0,0 +1,62 @@
...
+and allow several features like interactive documentation and automated
+client generation.
+
+.. important::
+    OpenAPI 2.0 currently doensn't some features that are present on Kinto API.
+    Some known limitations are
+    lack of validation on or clauses (e.g. provide `data` or `permissions` in PATCH
+    operations), no support for non schema querystrings, as used on :ref:`filtering <filtering>`
+    with additional fields, and no optional response headers, as required for
+    `Backoff signs`<http://kinto.readthedocs.io/en/stable/api/1.x/backoff.html>`.
```
glasserc: Something got messed up with your RST here. I think you have an extra backtick before your link. You might also need an underscore after the closing backtick.

gabisurita:
> I don't really understand why turning additionalProperties from {"description": "blah"} to {} fixes anything.

In terms of logic that's super weird, but following the traceback for the error, [unmarshal_model](https://github.com/Yelp/bravado-core/blob/master/bravado_core/unmarshal.py#L164), has a special case for empty objects, which skips [unmarshal_object](
https://github.com/Yelp/bravado-core/blob/master/bravado_core/unmarshal.py#L34), that explicitly checks for type. So there's a indeed a special case to handle empty objects. We may call it a bug on bravado, but at this point I don't really know what's a bug and what's not... ¯\_(ツ)_/¯

```
/home/travis/build/gabisurita/kinto/.tox/py34/lib/python3.4/site-packages/bravado_core/unmarshal.py:173: in unmarshal_model

    model_as_dict = unmarshal_object(swagger_spec, model_spec, model_value)

/home/travis/build/gabisurita/kinto/.tox/py34/lib/python3.4/site-packages/bravado_core/unmarshal.py:133: in unmarshal_object

    result[k] = unmarshal_schema_object(swagger_spec, prop_spec, v)

/home/travis/build/gabisurita/kinto/.tox/py34/lib/python3.4/site-packages/bravado_core/unmarshal.py:54: in unmarshal_schema_object

))))

E           bravado_core.exception.SwaggerMappingError: The following schema object is missing a type field: {'description': 'Any extra field that applies.'}
```

docs/api/1.x/openapi.rst
gabisurita: It may be unlikely, but I think it possible to have conflicts in plugins. But I think we can drop this keep it as a default, and reintroduce if we need it later. What do you think?

glasserc:
OK, I'm r+ on this. @Natim @leplatrem ?

leplatrem:
I'm in favor of merging it too!

There's one little detail that we discussed recently: the URL `/swagger.json`. We don't have this `.json` suffix in any of the other endpoints, and since the spec is evolving to *Open API*, maybe we should not insist too much on the former name *swagger*. 

In Shavar @tarekziade used the URL `/__api__` which is agnostic and consistent with the other ones like `/__version__` or `/__heartbeat__`.  https://github.com/mozilla-services/shavar/pull/78/

I don't want to delay the merging of this PR, but if you agree we can create an issue and think about it before we release the next Kinto version :)

Congrats!

gabisurita:
> There's one little detail that we discussed recently: the URL /swagger.json. We don't have this .json suffix in any of the other endpoints, and since the spec is evolving to Open API, maybe we should not insist too much on the former name swagger.

I agree. [According to spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#file-structure), there's still a convention in using `swagger.json` to refer to OpenAPI description files, but it doesn't say anything about resource paths.

tests/core/test_views_swagger.py
```diff
@@ -0,0 +1,65 @@
+import unittest
+import mock
+import yaml
```
glasserc: Do we have to change this to `ruamel.yaml` now that we've updated the dependency? Sorry if this is a stupid question, I've never used `ruamel.yaml` before...

gabisurita: Yes, we should. This should have raised on tests, but I guess that's another package that imports pyYAML.

glasserc:
🎊

PR #978, opened 2016-12-09
document 400 on get object requests

gabisurita:
400 is possible on GET objects with invalid headers, example:

```http
$ http get localhost:8888/v1/buckets/b1 If-Match:111 -a aaa:aaa
HTTP/1.1 400 Bad Request
Content-Length: 293
Date: Fri, 09 Dec 2016 01:11:30 GMT
Server: waitress

{
    "code": 400,
    "details": [
        {
            "description": "Invalid value for If-Match. The value should be integer between double quotes.",
            "location": "header",
            "name": null
        }
    ],
    "errno": 107,
    "message": "header: Invalid value for If-Match. The value should be integer between double quotes."
}
```

### DIFF
--- a/docs/api/1.x/_status-get-object.rst
+++ b/docs/api/1.x/_status-get-object.rst
@@ -3,6 +3,7 @@ HTTP Status Codes
 
 * |status-200|: The request was processed
 * |status-304|: Object did not change since value in ``If-None-Match`` header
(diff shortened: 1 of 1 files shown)


leplatrem:
Thanks for taking the time to fix that!

PR #981, opened 2016-12-14
Fix JSON-Merge validation (fixes #979)

### LINKED ISSUE #979: JSON Merge Patch deserialization is broken
Merge-patch operations are broken. It looks like a deserialization problem, which may have been introduced with  #790.  Also, we should definitely include some view tests for this content-type. My bad :/ 
  
```
echo '{"data": {"aaa": "bbb"}}' | http patch localhost:8888/v1/buckets/b1 Content-Type:application/merge-patch+json -a aaa:
HTTP/1.1 400 Bad Request
Content-Length: 331
Date: Sun, 11 Dec 2016 23:47:26 GMT
Server: waitress

{
    "code": 400,
    "details": [
        {
            "description": "\"{\"data\": {\"aaa\": \"bbb\"}}\n\" is not a mapping type: Does not implement dict-like functionality.",
            "location": "body",
            "name": ""
        }
    ],
    "errno": 107,
    "message": "\"{\"data\": {\"aaa\": \"bbb\"}}\n\" is not a mapping type: Does not implement dict-like functionality."
}
```

gabisurita:
Fixes #979 

I've hacked a quick fix for this issue. It may not be the most elegant solution, but it works. I've also added some view tests for JSON Merge to make sure this doesn't happen again.

- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem  @glasserc

### DIFF
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -95,6 +95,25 @@ def register_resource(resource_cls, settings=None, viewset=None, depth=1,
             argument_getter = getattr(viewset, '%s_arguments' % endpoint_type)
             view_args = argument_getter(resource_cls, method)
 
+            # We need to support JSON-merge for validation, so the approch
+            # is to validate the Content-Type and then cast it as regular JSON
+            # request.
+            def validate_json_content_type(req, *args, **kargs):
+                """
+                Cast JSON Content-Type variations as JSON for validation as
+                they are JSON anyways.
+                """
+
+                if re.match('^application/(.*?)json$', req.content_type):
+                    req.validated_content_type = req.content_type
+                    req.content_type = 'application/json'
+                else:
+                    req.validated_content_type = None
+
+                return req
+
+            view_args['validators'].insert(0, validate_json_content_type)
+
             view = viewset.get_view(endpoint_type, method.lower())
             service.add_view(method, view, klass=resource_cls, **view_args)
 
@@ -657,9 +676,8 @@ class UserResource(object):
             applied_changes = requested_changes.copy()
             updated = record.copy()
 
-            content_type = str(self.request.headers.get('Content-Type')).lower()
             # recursive patch and remove field if null attribute is passed (RFC 7396)
-            if content_type == 'application/merge-patch+json':
+            if self.request.validated_content_type == 'application/merge-patch+json':
                 recursive_update_dict(updated, applied_changes, ignores=[None])
             else:
                 updated.update(**applied_changes)
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -95,26 +95,6 @@ def register_resource(resource_cls, settings=None, viewset=None, depth=1,
             argument_getter = getattr(viewset, '%s_arguments' % endpoint_type)
             view_args = argument_getter(resource_cls, method)
 
-            # FIXME: We need to support JSON-merge validation, so the approch
-            # is to validate the Content-Type as a JSON variant and then cast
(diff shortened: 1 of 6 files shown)


tests/core/resource/test_record.py
```diff
@@ -228,9 +228,69 @@ def test_patch_record_returns_updated_fields(self):
...
+            result = self.resource.patch()['data']
+        self.assertDictEqual(result, {'position': 0})
+
+    def test_returns_changed_fields_if_behaviour_is_light(self):
+        self.resource.request.json = {'data': {'unread': True, 'position': 15}}
+        self.resource.request.headers['Response-Behavior'] = 'light'
+        with mock.patch.object(self.resource.model, 'update_record',
+                               return_value={'unread': True, 'position': 0}):
+            result = self.resource.patch()['data']
+        self.assertDictEqual(result, {'unread': True, 'position': 0})
```
glasserc: What are these changes about?

gabisurita: I just wrapped the Merge Patch tests in a new class, that moved up all the tests that were below. It's really weird on the diff.

kinto/core/resource/__init__.py
```diff
@@ -95,6 +95,26 @@ def register_service(endpoint_type, settings):
...
+                if re.match('^application/(.*?)json$', req.content_type):
+                    req.validated_content_type = req.content_type
+                    req.content_type = 'application/json'
+                else:
+                    req.validated_content_type = None
+
+                return req
+
+            if 'validators' in view_args:
+                view_args['validators'].insert(0, validate_json_content_type)
```
leplatrem: We can probably move this to `core.resource.viewset` (and handle a global refactor later with Kinto/kinto#880)

Or, better, contribute the fix to Cornice: https://github.com/Cornices/cornice/blob/master/cornice/validators/__init__.py#L36
(i.e. `re.match()` on the content_type header instead of equality)

gabisurita:
@glasserc 

What happens is that the default cornice validator doesn't understand (and thus don't deserialize variations of JSON). This may be a cornice/colander issue, but I'm not sure if it really is, or how to address it there.

I agree about using a list instead of regular expressions. Notice that we already check for acceptable content-types at `viewset.py`, so we shouldn't expect nothing there that isn't Merge-Patch, Json-Patch or regular Json. Actually in this solution we don't need validation at all, we just need to cast the Content-Type for validation and store the previous one in an attribute such as the patch method can know it is a merge request.

https://github.com/gabisurita/kinto/blob/25a67356dd6088912a1c3001c0526e2ee04a85e9/kinto/core/resource/viewset.py#L12

leplatrem:
I must say that beyond the elegance of this patch, I'm really pleased by the quality of the pull-request (tests notably etc.) 👍

leplatrem:
> This may be a cornice/colander issue

Yes, this can be fixed in cornice

gabisurita:
So I suggest dropping the patch here (keeping the tests) and fixing it on cornice. We can merge the tests after this has been fixed on cornice.

 @leplatrem Do you think we should keep using regular expressions? There are [several types](http://www.iana.org/assignments/media-types/media-types.xhtml) match `+json` that i think could be supported.

glasserc:
Ah, I understand [now](https://github.com/Cornices/cornice/blob/master/cornice/validators/__init__.py#L36)..

Yes, let's keep the tests, I agree they're good and valuable.

leplatrem:
By adding a `cornice >= 2.3` in `setup.py`, the tests should pass now :)

leplatrem:
:huge thanks:

PR #982, opened 2016-12-16
Fix crash on redirection when path contains unicode characters

leplatrem:


leplatrem:
You know what, I'm done with this. Life is too beautiful to ruin it with crazy sh$¤ like UnicodeDecodeError and bytes-like object required.

Natim:
It is not your fault @leplatrem this thing is shitty...

tests/core/test_views_errors.py
```diff
@@ -183,6 +183,12 @@ def test_redirection_does_not_allow_crlf(self):
     def test_redirection_does_not_allow_control_characters(self):
         self.app.get('/9l2j7%0A21m2n', status=404)
 
+    def test_redirection_allows_unicode_characters(self):
+        self.app.get('/crlftest%E5%98%8AXTest:crlftest/', status=307)
```
glasserc: Maybe add a comment showing the character that this represents? `嘊`

glasserc:
🎆 ❇️ 💥

leplatrem:
I was wrong about webob, at least regarding `request.path` (some other issues exist with parameters like https://github.com/Pylons/webob/issues/161).

The `request.matchdict` (Pyramid url dispatch) has unicode, whereas `request.path` hasn't.

Hence the great simplification.

PR #990, opened 2016-12-19
Add command to purge tombstones

leplatrem:
Related to #985

### DIFF
--- a/kinto/core/scripts.py
+++ b/kinto/core/scripts.py
@@ -29,6 +29,7 @@ def migrate(env, dry_run=False):
 
 
 def delete_collection(env, bucket_id, collection_id):
+    # XXX: this script contains specific notions of Kinto. Should not be in core.
     registry = env['registry']
     settings = registry.settings
     readonly_mode = asbool(settings.get('readonly', False))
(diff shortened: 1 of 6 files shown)


kinto/core/scripts.py
```diff
@@ -85,3 +85,18 @@ def delete_collection(env, bucket_id, collection_id):
...
+
+
+def purge_deleted(env, timestamp):
+    if timestamp is None:
+        logger.warning('Purge all tombstones.')
+    else:
+        logger.info('Purge tombstones before %r.' % timestamp)
+    registry = env['registry']
+
+    count = registry.storage.purge_deleted(parent_id='*', collection_id=None, before=timestamp)
```
Natim: I think we should allow people to configure the bucket and collection they want to purge, don't you think?

leplatrem: Do we have a use case :) ?

tests/test_main.py
```diff
@@ -134,6 +134,45 @@ def test_cli_delete_collection_run_delete_collection_script(self):
...
+            args, kwargs = purge_deleted.call_args
+            self.assertEqual(args[2], "/buckets/b")
+
+    def test_cli_purge_deleted_with_collection(self):
+        with mock.patch('kinto.__main__.scripts.purge_deleted') as purge_deleted:
+            purge_deleted.return_value = mock.sentinel.purge_deleted
+            main(['--ini', TEMP_KINTO_INI, 'init', '--backend', 'memory'])
+            main(['--ini', TEMP_KINTO_INI, 'purge-deleted', '--collection', 'c'])
+            args, kwargs = purge_deleted.call_args
+            self.assertEqual(args[2], "/buckets/*/collections/c")
```
Natim: wow powerfull

Natim:
This command is really too dangerous, for instance you cannot remove tombstones only for record or only for collection.

I would differentiate between:
 - `/buckets/bid/collections/cid`
and
 - `/buckets/bid/collections/cid/records`

Or between

- `/buckets/bid/collections/*`

And

- `/buckets/bid/collections/*/records`

leplatrem:
> This command is really too dangerous

This command purges tombstones. That's it! If this is *dangerous*, then let's not provide the feature, especially now that #991 was merged.

> I would differentiate between [...]

How? With an additional CLI parameter?

> /buckets/bid/collections/*/records

BTW there is not such parent. Only `parent_id='/buckets/bid/collections/*'`, `collection_id='record'`

leplatrem:
I realize that my previous was very straight to the point, and can be read with a very serious tone :| Sorry if it came out that way, didn't mean to be violent :]

leplatrem:
> This command is really too dangerous

BTW if your usage of Kinto does not rely on offline sync, then it is fine to remove the tombstones!

Natim:
> If your usage of Kinto does not rely on offline sync, then it is fine to remove the tombstones!

It is but the usecases we have do rely on it.

leplatrem:
Closing then!

PR #997, opened 2016-12-20
Paginated delete (fixes #987)

### LINKED ISSUE #987: Delete is not affected by limit query
The [current documentation](https://kinto.readthedocs.io/en/latest/api/1.x/pagination.html) says:

> Plural endpoints support limiting the number of elements returned. For example, you can retrieve a fixed number of records in a collection. To do this, provide a `_limit` query parameter specifying the number of records to return.
(...)
Pagination works on any plural endpoint.

It doesn't work on delete endpoints and I'm not sure if it should. I guess the behavior in this case might be a bit dubious. Would it limit the number of deleted entries or the number of returned deleted entries? 

An example:
```json
$ http delete localhost:8888/v1/buckets?_limit=1 -a a:HTTP/1.1 200 OK
Content-Length: 181
Date: Sun, 18 Dec 2016 18:53:46 GMT
Server: waitress

{
    "data": [
        {
            "deleted": true,
            "id": "b1",
            "last_modified": 1482087226574
        },
        {
            "deleted": true,
            "id": "b2",
            "last_modified": 1482087226575
        },
        {
            "deleted": true,
            "id": "b3",
            "last_modified": 1482087226576
        }
    ]
}


```

leplatrem:
Fixes #987 

- [x] Add limit/sorting to querystring
- [x] Add next-page header to plural delete responses
- [x] Plug model resource models
- [x] Add necessary parameters to storage backend
- [x] Check out documentation

### DIFF
--- a/kinto/core/resource/model.py
+++ b/kinto/core/resource/model.py
@@ -111,12 +111,22 @@ class Model(object):
             comparison (see `kinto.core.utils.COMPARISON`). All filters
             are combined using *AND*.
         :type filters: list of :class:`kinto.core.storage.Filter`
-        :param sorting: XXX
-        :type sorting: XXX
-        :param pagination_rules: XXX
-        :type pagination_rules: XXX
-        :param limit: XXX
-        :type limit: XXX
+
+        :param sorting: Optionnally sort the records by attribute.
+            Each sort instruction in this list refers to a field and a
+            direction (negative means descending). All sort instructions are
+            cumulative.
+        :type sorting: list of :class:`kinto.core.storage.Sort`
+
+        :param pagination_rules: Optionnally paginate the deletion of records.
+            This list of rules aims to reduce the set of records to the current
+            page. A rule is a list of filters (see `filters` parameter),
+            and all rules are combined using *OR*.
+        :type pagination_rules: list of list of
+            :class:`kinto.core.storage.Filter`
+
+        :param int limit: Optionnally limit the number of records to be
+           deleted.
 
         :param str parent_id: optional filter for parent id
 
--- a/kinto/core/resource/model.py
+++ b/kinto/core/resource/model.py
@@ -100,7 +100,8 @@ class Model(object):
             auth=self.auth)
         return records, total_records
 
-    def delete_records(self, filters=None, parent_id=None):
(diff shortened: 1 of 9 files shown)


kinto/core/resource/model.py
```diff
@@ -110,6 +111,12 @@ def delete_records(self, filters=None, parent_id=None):
             comparison (see `kinto.core.utils.COMPARISON`). All filters
             are combined using *AND*.
         :type filters: list of :class:`kinto.core.storage.Filter`
+        :param sorting: XXX
```
Natim: :param filters: Optionally filter the records by their attribute.
             Each filter in this list is a tuple of a field, a value and a
             comparison (see `kinto.core.utils.COMPARISON`). All filters
             are combined using *AND*.
         :type filters: list of :class:`kinto.core.storage.Filter`
 
         :param sorting: Optionnally sort the records by attribute.
             Each sort instruction in this list refers to a field and a
             direction (negative means descending). All sort instructions are
             cumulative.
         :type sorting: list of :class:`kinto.core.storage.Sort`
 
         :param pagination_rules: Optionnally paginate the list of records.
             This list of rules aims to reduce the set of records to the current
             page. A rule is a list of filters (see `filters` parameter),
             and all rules are combined using *OR*.
         :type pagination_rules: list of list of
             :class:`kinto.core.storage.Filter`
 
         :param int limit: Optionnally limit the number of records to be
            retrieved.

kinto/core/resource/model.py
```diff
@@ -111,6 +112,22 @@ def delete_records(self, filters=None, parent_id=None):
             are combined using *AND*.
         :type filters: list of :class:`kinto.core.storage.Filter`
 
+        :param sorting: Optionnally sort the records by attribute.
```
glasserc: `Optionally` 🐑

leplatrem:
@gabisurita so it should return the total records that will deleted accross all pages right?

gabisurita:
@leplatrem Yes, I think so. It should return the number of buckets that would be deleted without pagination.

leplatrem:
could you open an issue please :) ?
I think it's trivial because we do a get_all() just before performing the deletion...

PR #1001, opened 2016-12-22
add Total-Records to collection DELETE headers

### LINKED ISSUE #1000: Add a Total-Records header to delete on plural endpoints
For consistency with GET, now that we support pagination on DELETE.

gabisurita:
Fixes #1000 

- [x] Add tests.
- [x] Add a changelog entry.


CHANGELOG.rst
```diff
@@ -9,6 +9,14 @@ This document describes changes between each past release.
 - Nothing changed yet.
 
 
+5.1.1 (Unreleased)
+------------------
+
+**Bug fixes**
+
+- Add missing ``Total-Records`` headers on ``DELETE`` with plural endpoints (fixes #1000)
```
leplatrem: *header* (singular)

tests/core/resource/test_model.py
```diff
@@ -64,6 +64,13 @@ def test_delete_supports_collection_filters(self):
         records = result['data']
         self.assertEqual(len(records), 1)
 
+    def test_delete_collection_gives_number_of_deletable_records_in_headers(self):
+        self.resource.request.GET = {'_limit': '1'}
+        self.resource.collection_delete()
+        headers = self.last_response.headers
+        count = headers['Total-Records']
+        self.assertEquals(int(count), 2)
```
leplatrem: I don't think this should go in the model tests, since it is done in the resource. `PaginatedDeleteTest` seems more appropriate

gabisurita: I agree, i've included it there because there's a pretty similar test to check `Total-Records` on get operations. Maybe we should move both.

https://github.com/gabisurita/kinto/blob/4ba9cb6a5dacc015b39754a4b4f30dc26ec55a5e/tests/core/resource/test_model.py#L11

leplatrem: yep!

Issue #949, opened 2016-11-27
Change default listening address from 0.0.0.0 to 127.0.0.1 


julienrat:
Hi all !
I'm trying to run kinto in a local network without internet connexion, When i plug the internet on my kinto server, it works perfectly, but when i launch kinto with no internet connexion ! have got this errors : 

```
root@padaone:/home/padaone/formbuilder/kinto# make serve
.venv/bin/kinto --ini config/kinto.ini migrate
/home/padaone/formbuilder/kinto/kinto/core/initialization.py:568: UserWarning: HTTPS is not enabled
  warnings.warn('HTTPS is not enabled')
/home/padaone/formbuilder/kinto/kinto/core/storage/postgresql/client.py:88: UserWarning: Reuse existing PostgreSQL connection. Parameters permission_* will be ignored.
  warnings.warn(msg)
/home/padaone/formbuilder/kinto/kinto/core/storage/postgresql/client.py:88: UserWarning: Reuse existing PostgreSQL connection. Parameters cache_* will be ignored.
  warnings.warn(msg)
INFO   Running kinto 5.0.1.dev0. 
INFO   PostgreSQL cache schema is up-to-date. 
INFO   Detected PostgreSQL storage schema version 14. 
INFO   PostgreSQL storage schema is up-to-date. 
INFO   PostgreSQL permission schema is up-to-date. 
echo '{"name":"kinto","version":"5.0.0-19-g0cb4e9d","source":"https://github.com/Kinto/kinto.git","commit":"0cb4e9da6974a78f10d1a98ed0dd88f50a2417a3"}' > version.json
.venv/bin/kinto --ini config/kinto.ini start --reload
Starting subprocess with file monitor
/home/padaone/formbuilder/kinto/kinto/core/initialization.py:568: UserWarning: HTTPS is not enabled
  warnings.warn('HTTPS is not enabled')
/home/padaone/formbuilder/kinto/kinto/core/storage/postgresql/client.py:88: UserWarning: Reuse existing PostgreSQL connection. Parameters permission_* will be ignored.
  warnings.warn(msg)
/home/padaone/formbuilder/kinto/kinto/core/storage/postgresql/client.py:88: UserWarning: Reuse existing PostgreSQL connection. Parameters cache_* will be ignored.
  warnings.warn(msg)
2016-11-28 00:23:25,599 INFO  [kinto.core][MainThread] Running kinto 5.0.1.dev0. 
Starting server in PID 2685.
Traceback (most recent call last):
[...]

PR #1005, opened 2016-12-24
Amending default address - issue 949

PeriGK:
Hi there,

I have done some changes regarding issue #949. Not sure if this PR is complete, as I am havng some concerns that might be blockers.

1) The load tests failed(I have attached the log file). Not sure if there is something like "known issues" but I have to report it anyway.
2) There are few occurences of both IPs in the code, hardcoded(in test code). I believe they will not affect the app at all, but I would like to double check it.
3) In the contribution guidelines there is a not about always having unit tests. I suppose this is not applicable here, right?

Let me know if you need anything.

Thanks and Merry Christmas,
P.

[load_tests.txt](https://github.com/Kinto/kinto/files/671800/load_tests.txt)


PeriGK:
Hi @gabisurita ,

I have updated the changelog as well.

Thanks,
P.

Natim:
Waiting for #1009 before merging.

Natim:
Thank you @PeriGK :)

PR #1010, opened 2016-12-30
Run tests with Python 3.6

Natim:
I am not sure if we should keep py34/py35 I guess we could.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,9 @@ after_success:
 
 matrix:
   include:
-    - python: 3.5
+    - python: 3.6
       env:
-        - TOX_ENV=py35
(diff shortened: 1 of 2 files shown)


Natim:
In production we use 3.5 for webextension and Docker stacks and 2.7 for older stacks.

Natim:
Apparently tests are not running on Py3.6 yet :)

Natim:
I filled a PR upstream but it seems to be already fixed in master, just not released.

https://github.com/stefankoegl/python-json-patch/pull/58

Natim:
python-json-patch 1.15 has been released so it should work now. Re launching tests.

Issue #1006, opened 2016-12-27
Generate the OpenAPI specification from Python code.


gabisurita:
Follow up discussion from #977.

PR #1021, opened 2017-01-17
Validate header and querystring with cornice schemas (fixes #873)

### LINKED ISSUE #873: Validate headers and querystring with cornice schemas

gabisurita:
Fixes #873 
Related to #1006 

- [x] Fix unicode in headers
- [x] Fix JSON Patch (related to #880)
- [x] Investigate if there's anything else that can be validated (specially some filters)
- [x] Add a changelog entry

r? @


kinto/core/resource/__init__.py
```diff
@@ -542,7 +542,8 @@ def patch(self):
                 new_record[extra_field] = existing[extra_field]
 
         # Adjust response according to ``Response-Behavior`` header
-        body_behavior = self.request.headers.get('Response-Behavior', 'full')
+        body_behavior = self.request.validated.get('header',
+                                                   {}).get('Response-Behavior', 'full')
```
leplatrem: When does it happen that `validated` doesn't have a `header` entry? See https://github.com/Cornices/cornice/blob/9b73c5ae8dfbebede6413a007ffac7fe28e76401/cornice/validators/__init__.py#L60-L62

kinto/core/resource/__init__.py
```diff
@@ -812,36 +813,20 @@ def _raise_304_if_not_modified(self, record=None):
...
-        try:
-            if not (if_none_match[0] == if_none_match[-1] == '"'):
-                raise ValueError()
-            modified_since = int(if_none_match[1:-1])
-        except (IndexError, ValueError):
-            if if_none_match == '*':
-                return
-            raise_invalid(self.request, **error_details)
+        if if_none_match == '*':
+            return
```
leplatrem: ❤️

kinto/core/resource/__init__.py
```diff
@@ -918,9 +887,8 @@ def _raise_400_if_id_mismatch(self, new_id, record_id):
     def _extract_partial_fields(self):
         """Extract the fields to do the projection from QueryString parameters.
         """
-        fields = self.request.GET.get('_fields', None)
+        fields = self.request.validated.get('querystring', {}).get('_fields')
```
leplatrem: Ditto: could this happen that `querystring` is missing? If it's for the test, we could fix the test setUp instead I believe.

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,71 @@ class BookmarkSchema(ResourceSchema):
 
     def preparer(self, appstruct):
         return strip_whitespace(appstruct)
+
+
+class CSVQuerystring(colander.SchemaNode):
```
leplatrem: What do you think of `FieldList` or `StringList` instead ?

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,71 @@ class BookmarkSchema(ResourceSchema):
...
+
+    schema_type = colander.String
+    missing = colander.drop
+    error_message = "The value should be a list of comma separated field names"
+
+    def deserialize(self, cstruct=colander.null):
+        params = super(CSVQuerystring, self).deserialize(cstruct)
+        if params is colander.drop:
+            return params
+        else:
```
leplatrem: nit: superfluous `else`

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,71 @@ class BookmarkSchema(ResourceSchema):
...
+            return params.split(',')
```
leplatrem: I think there is the notion of `preparer` in Colander that could split, and then the notion of `Sequence`. In a second iteration, we can try to leverage that if possible

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,71 @@ class BookmarkSchema(ResourceSchema):
...
+    missing = colander.drop
+    error_message = "The value should be integer between double quotes"
+    validator = colander.Any(colander.Regex('^"([0-9]+?)"$', msg=error_message),
+                             colander.Regex('\*'))
+
+    def deserialize(self, cstruct=colander.null):
+        if isinstance(cstruct, six.string_types):
+            try:
+                cstruct = decode_header(cstruct)
+            except:
```
leplatrem: `except UnicodeDecodeError` ?

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,71 @@ class BookmarkSchema(ResourceSchema):
...
+
+
+class HeaderSchema(colander.MappingSchema):
+    if_match = HeaderQuotedInteger(name='If-Match', missing=colander.drop)
+    if_none_match = HeaderQuotedInteger(name='If-None-Match', missing=colander.drop)
+    response_behaviour = colander.SchemaNode(colander.String(),
+                                             name='Response-Behavior',
+                                             validator=colander.OneOf(
+                                                 ['full', 'light', 'diff']),
+                                             missing=colander.drop)
```
leplatrem: Is `decode_header` not necessary for this one? Maybe we could run `decode_header` here before deserailize each sub-node so that we have it one place only? Or do it in Cornice instead...

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,71 @@ class BookmarkSchema(ResourceSchema):
...
+        param = super(HeaderQuotedInteger, self).deserialize(cstruct)
+        if param is colander.drop or param == '*':
+            return param
+        else:
+            return int(param[1:-1])
+
+
+class HeaderSchema(colander.MappingSchema):
+    if_match = HeaderQuotedInteger(name='If-Match', missing=colander.drop)
+    if_none_match = HeaderQuotedInteger(name='If-None-Match', missing=colander.drop)
```
leplatrem: I see that you repeat `missing=drop`, maybe we can remove it from each schema then!

kinto/core/resource/viewset.py
```diff
@@ -120,13 +120,13 @@ def get_view_arguments(self, endpoint_type, resource_cls, method):
...
         if method.lower() in map(str.lower, self.validate_schema_for):
-            schema = PartialSchema()
+            schema = RequestSchema()
             record_schema = self.get_record_schema(resource_cls, method)
             record_schema.name = 'body'
             schema.add(record_schema)
             args['schema'] = schema
         else:
-            args['schema'] = SimpleSchema()
+            args['schema'] = RequestSchema()
```
leplatrem: One was `Partial` and the other one `Simple`. Is that ok to now give the same behaviour to both?

gabisurita: Well, I've got why `StrictSchema` and `SimpleSchema`were needed by `get_record_schema`but I actually didn't understand why we needed `PartialSchema` and `SimpleSchema` on this one, aren't we just setting the request schema here (in contrary to the body schema)? 

Also, this didn't break any tests and the API behavior looks ok, so I think it's ok to keep the same behavior to both.

kinto/core/resource/schema.py
gabisurita: I guess handling it on cornice is a good idea. I'm also not sure if it's still needed.

leplatrem: I can't remember why we had that, but if I remember well it's because of Webob (/cc @Natim)

kinto/core/resource/viewset.py
leplatrem: Ok then!

kinto/core/resource/schema.py
gabisurita: This can definitely be improved but I didn't understand how to use `preparer` to do it. Thinking through it, maybe defining it as `Sequence` we can do something as simple as:

```python
class FieldList(colander.SchemaNode):
    fields = colander.SchemaNode(colander.String(), missing=colander.drop)

    def deserialize(self, cstruct=colander.null):
        if isinstance(cstruct, six.string_types):
            cstruct = cstruct.split(',')
        return super(FieldList, self).deserialize(cstruct)

```

gabisurita:
Note: I'm probably going to address #880 on this PR as well.

glasserc:
Looks OK to me. Maybe I misunderstood what you were saying in our meeting earlier, but it sounded like you needed to fix the JSON thing to fix the build failures, but the build failures don't seem related to the JSON library validation *per se*. So what's the rationale for fixing the JSON thing in this PR? (Just curious...)

gabisurita:
> So what's the rationale for fixing the JSON thing in this PR?
Getting JSON Patch requests to be validated by cornice. We need this because now we are using colander for validating and deserializing other aspects of the request that are also needed on JSON Patch (sync headers, etc). Right now we only trust the external library for validation and ignore cornice validation for this content-type. https://github.com/Kinto/kinto/blob/master/kinto/core/resource/__init__.py#L101

The problem is that I just discovered colander doesn't accept JSON Arrays at the top level, so this may be a bit more tricky than it seems. https://github.com/Cornices/cornice/issues/433

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,142 @@ class BookmarkSchema(ResourceSchema):
...
+        return colander.Mapping(unknown='ignore')
+
+    def deserialize(self, cstruct=colander.null):
+        values = {}
+
+        schema_values = super(QuerySchema, self).deserialize(cstruct)
+        if schema_values is colander.drop:
+            return schema_values
+
+        # Deserialize filters
```
gabisurita: It does not look very good, but think it's better to handle the filter deserialization here than leave it to the `Resource._extract_filters` method.

Opinions?

glasserc: It doesn't look bad by any means IMHO, and I really like having an explicit, formal schema, so I think this is an improvement.

leplatrem: Me neither, well done!

I have one remark though: the above comment could be more explicit about `filters` (field filters?) to help understand what the code does. A couple of minimalist examples with boolean or list value could help for example (`?deleted=true` -> `{"deleted": True}`)

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,142 @@ class BookmarkSchema(ResourceSchema):
...
+        values.update(schema_values)
+        return values
+
+
+class JsonPatchOperationSchema(colander.MappingSchema):
+    """Single JSON Patch Operation."""
+
+    op = colander.SchemaNode(colander.String(),
+                             validator=colander.OneOf(
+                                 ['test', 'add', 'remove', 'replace', 'move', 'copy']))
```
leplatrem: please move this list elsewhere (or whole validator) to improve indentation ;) #IndentationFreak

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,142 @@ class BookmarkSchema(ResourceSchema):
...
+    path = colander.SchemaNode(colander.String())
+    from_ = colander.SchemaNode(colander.String(), name='from', missing=colander.drop)
+
+    @staticmethod
+    def schema_type():
+        return colander.Mapping(unknown='preserve')
```
leplatrem: shouldn't we allow `value` only instead ?

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,142 @@ class BookmarkSchema(ResourceSchema):
...
+
+
+class JsonPatchOperationSchema(colander.MappingSchema):
+    """Single JSON Patch Operation."""
+
+    op = colander.SchemaNode(colander.String(),
+                             validator=colander.OneOf(
+                                 ['test', 'add', 'remove', 'replace', 'move', 'copy']))
+    path = colander.SchemaNode(colander.String())
+    from_ = colander.SchemaNode(colander.String(), name='from', missing=colander.drop)
```
leplatrem: cherry on cake: you could have a regex to make sure those look valid :)

kinto/core/resource/schema.py
```diff
@@ -150,3 +151,142 @@ class BookmarkSchema(ResourceSchema):
...
+
+class RequestSchema(colander.MappingSchema):
+    """Baseline schema for kinto requests."""
+
+    header = HeaderSchema(missing=colander.drop)
+    querystring = QuerySchema(missing=colander.drop)
+
+
+class JsonPatchRequestSchema(RequestSchema):
+    body = JsonPatchBodySchema()
```
leplatrem: Technically we don't support as many querystring values for a patch. But maybe we can keep it simple as it is, they don't harm.

Your call!

kinto/core/resource/schema.py
gabisurita: The problem is that value don't have an specific type, so we can't set it on the schema (or at least I don't know how to do it). Maybe we could check this at `deserialize`?

kinto/core/resource/schema.py
gabisurita: Yeah, that was something I was thinking about earlier... this doesn't apply only for JSON Patch.

Should we use the same schema for all requests and validate all parameters, even the ones we won't use it, or set individual request schemas for each method with only the expected params? 

I think it's more explicit to have multiple request schemas, and it's better for documentation purposes, but that also means more code to maintain.

kinto/core/resource/schema.py
gabisurita: Or maybe define a new colander type that `deserialize` returns `ctruct`?

kinto/core/resource/schema.py
gabisurita: That would be cool, but I'm not sure what we can validate here. I know they all have to start with `/`, but I'm not sure if there's anything else to check here. Maybe if there is anything between two `/`? IDK

kinto/core/resource/schema.py
leplatrem: Just check its presence in deserialize maybe then

kinto/core/resource/schema.py
leplatrem: Yeah something like `^(\/\w)+$` ? .

kinto/core/resource/schema.py
leplatrem: As a step 1, what we have here is fine. If you struggle with that when working on #1006 then you can do a second round.

leplatrem:
The remaining changes to do are nice to have. You can merge as soon as you feel good about the code :)

GG !

leplatrem:
(*are you sure you added a changelog entry*?)

PR #1028, opened 2017-01-20
Upgrade Kinto Admin to 1.8

leplatrem:
⚠️  *Did not test locally*


Natim:
Did not test locally too bad...

PR #1033, opened 2017-01-25
[WIP] Use Cornice Swagger to generate the OpenAPI specification

### LINKED ISSUE #1006: Generate the OpenAPI specification from Python code.

gabisurita:
Fixes #1006 

This is still early WIP but I need some feedback on the architecture. If you think the changes are "too breaking" we can rollback ASAP and find another approach. The more controversial change IMO is to move parts of the schema from resource to a separate module. I've did this to allow imports from other parts of the code that are imported before Service is defined (like errors), and most of the schemas there are not exclusive to resources. Opinions? 

Todos:

- [x] Release cornice_swagger 0.4
- [x] Add basicauth security.
- [x] Add missing Tags , operation ids and responses
- [ ] Investigate security extensions. (https://github.com/Cornices/cornice.ext.swagger/issues/34)
- [ ] Fix broken tests.
- [ ] Add/Modify schema tests.
- [ ] Add/Modify swagger tests.
- [ ] Sort operations properly.
- [ ] Add documentation.
- [ ] Add a changelog entry.
- [ ] Cleanup .yaml files

Extras (can be finished at other PRs):

- [ ] Add response details.
- [ ] Add descriptions everywhere.
- [ ] Investigate using the response schemas to actually build responses.
- [ ] Use references to reduce document size (https://github.com/Cornices/cornice.ext.swagger/issues/57)

r? @glasserc


kinto/core/resource/schema.py
```diff
@@ -1,316 +1,4 @@
...
-class RequestSchema(colander.MappingSchema):
-    """Baseline schema for kinto requests."""
-
-    header = HeaderSchema(missing=colander.drop)
-    querystring = QuerySchema(missing=colander.drop)
-
-
-class JsonPatchRequestSchema(RequestSchema):
-    body = JsonPatchBodySchema()
+__all__ = ["ResourceSchema", "PermissionsSchema", "TimeStamp", "URL"]
```
gabisurita: Should we preserve backward compatibility here?

kinto/core/views/batch.py
```diff
@@ -68,13 +69,39 @@ class BatchRequest(colander.MappingSchema):
     body = BatchPayloadSchema()
 
 
+class BatchResponseSchema(colander.MappingSchema):
```
gabisurita: We could also move this to schema.

glasserc:
I didn't have time to do a thorough review, but from what I see, this looks OK to me. I don't have a problem with concentrating fundamental or common schemas in a `schemas` module. (In Haskell, sometimes projects will have a `Types.hs` module where all the models are defined but most of the operations they support omitted. This seems like that to me.) On the other hand, I like better how you put e.g. `VersionResponseSchema` in the same module as the view that serves versions -- it's closer, so it seems to me like it will be more likely to be modified at the same time. So I would say that we should try to do that as much as we can, and fall back to putting things in `schemas` if we have to. At first I thought you put things in `schemas` to re-use them, but I see that e.g. `CollectionRequestSchema` is only used once?

The `ResourceResponses` thing seems a little complicated.. I guess this is to simplify shared error responses? It feels like it's tied too tightly to the different views we support.

glasserc:
I'd love to hear @leplatrem 's thoughts on this!

tests/swagger/test_validation.py
```diff
@@ -113,7 +113,7 @@ def test_validate_headers(self):
         for head in headers:
             self.request.headers = head
             self.assertRaises(ValidationError, unmarshal_request,
-                              self.request, self.resources['Buckets'].get_buckets)
+                              self.request, self.resources['Bucket'].get_buckets)
```
glasserc: Ouch.. too bad we have to make dozens of changes like this. Maybe we should try to find a few minutes in a subsequent PR to extract this expression as an attribute on the test case, something like `self.swagger_resource`?

gabisurita: That would be good. I mean, now we are generating it automatically I don't think we should test it this way. We can probably think of something better.

gabisurita:
> The ResourceResponses thing seems a little complicated.. I guess this is to simplify shared error responses? It feels like it's tied too tightly to the different views we support.

Hmm, these are actually all the responses for the Resource class, and we are actually changing the record schemas (e.g allowed permissions) for each view [during initialization here](https://github.com/gabisurita/kinto/blob/1f7192660b472918095867817a007284c7c32228/kinto/core/resource/schema.py#L357). 

Changing the error responses is not something we would probably do in the server repository, because all storage views have the exactly same error messages. We did the same with the handwritten spec. But there may be some extensions that would like to change it. I just realized we can make responses a `Resource` attribute rather than a `ViewSet` attribute, than we could just change the responses messages by setting `response_schemas` on each Resource derived class.

gabisurita:
> So I would say that we should try to do that as much as we can, and fall back to putting things in schemas if we have to. At first I thought you put things in schemas to re-use them, but I see that e.g. CollectionRequestSchema is only used once?

`CollectionRequestSchema` is used only on `Resource`, and that's why it's on `kinto.core.resource.schema`. `Any` and `QueryField` are reused, and that's why they are on `kinto.core.schema`. What I did was to move some non resource specific schemas from `kinto.core.resource.schema` to `kinto.core.schema`.

gabisurita:
Closed in favor of #1078

PR #1036, opened 2017-01-26
Retries on cache.set IntegrityError

### LINKED ISSUE #1035: Setting value in cache should never fail
The cache is a key value store. The transaction isolation and integrity constraints are details of implementation. Setting a value in the cache should just never fail.

Natim:
Fixes #1035

- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,10 @@ This document describes changes between each past release.
 5.4.0 (unreleased)
 ------------------
 
+**Bug fixes**
+
+- Retries to set value in cache in case of IntegrityError (fixes #1035)
+
 **Internal changes**
 
 - Remove JSON Patch content-type from accepted types on the viewset, since it is handled
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 **Bug fixes**
 
-- Retries to set value in cache in case of IntegrityError (fixes #1035)
+- Retries to set value in PostgreSQL cache backend in case of IntegrityError (fixes #1035)
 
 **Internal changes**
 
--- a/kinto/core/cache/postgresql/__init__.py
+++ b/kinto/core/cache/postgresql/__init__.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+from functools import wraps
 
 import os
 
@@ -6,6 +7,17 @@ from kinto.core import logger
 from kinto.core.cache import CacheBase
 from kinto.core.storage.postgresql.client import create_from_config
 from kinto.core.utils import json
+
+
+def retry_on_failure(func):
+    @wraps(func)
+    def wraps_func(self, *args, **kwargs):
+        import psycopg2
+        try:
+            return func(self, *args, **kwargs)
+        except psycopg2.IntegrityError:
+            return func(self, *args, **kwargs)
+    return wraps_func
 
 
 class Cache(CacheBase):
@@ -121,6 +133,7 @@ class Cache(CacheBase):
         with self.client.connect() as conn:
             conn.execute(query, dict(ttl=ttl, key=self.prefix + key))
 
+    @retry_on_failure
     def set(self, key, value, ttl=None):
         if ttl is None:
             logger.warning("No TTL for cache key %r" % key)
--- a/kinto/core/cache/postgresql/__init__.py
+++ b/kinto/core/cache/postgresql/__init__.py
@@ -15,11 +15,11 @@ MAX_RETRIES = 5
 
 
 def retry_on_failure(func):
+    import psycopg2
 
     @wraps(func)
     def wraps_func(self, *args, **kwargs):
         tries = kwargs.pop('tries', 0)
-        import psycopg2
         try:
             return func(self, *args, **kwargs)
         except psycopg2.IntegrityError as e:
--- a/kinto/core/cache/postgresql/__init__.py
+++ b/kinto/core/cache/postgresql/__init__.py
@@ -10,12 +10,16 @@ from kinto.core.storage.postgresql.client import create_from_config
 from kinto.core.storage.exceptions import BackendError
 from kinto.core.utils import json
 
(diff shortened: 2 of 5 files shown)


kinto/core/cache/postgresql/__init__.py
```diff
@@ -8,6 +9,17 @@
...
 
 
+def retry_on_failure(func):
+    @wraps(func)
+    def wraps_func(self, *args, **kwargs):
+        import psycopg2
+        try:
+            return func(self, *args, **kwargs)
+        except psycopg2.IntegrityError:
+            return func(self, *args, **kwargs)
```
Natim: Should we retry more than once here?

kinto/core/cache/postgresql/__init__.py
```diff
@@ -8,6 +9,17 @@
 from kinto.core.utils import json
 
 
+def retry_on_failure(func):
+    @wraps(func)
+    def wraps_func(self, *args, **kwargs):
+        import psycopg2
```
leplatrem: nit: this could be imported from the top of file

kinto/core/cache/postgresql/__init__.py
leplatrem: We could retry 10 times with a small increasing delay between each call after the second call

CHANGELOG.rst
```diff
@@ -6,6 +6,10 @@ This document describes changes between each past release.
 5.4.0 (unreleased)
 ------------------
 
+**Bug fixes**
+
+- Retries to set value in cache in case of IntegrityError (fixes #1035)
```
leplatrem: cache PostgreSQL backend

kinto/core/cache/postgresql/__init__.py
Natim: I we put it on the top of the file it will fail in case the psycopg2 driver is not installed.

Natim:
```
――――――――――――――――――――――――――――― ERROR collecting tests/core/test_cache.py ―――――――――――――――――――――――――――――
ImportError while importing test module '/home/rhubscher/mozilla/kinto/tests/core/test_cache.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/core/test_cache.py:6: in <module>
    from kinto.core.cache import (CacheBase, memory as memory_backend,
kinto/core/cache/postgresql/__init__.py:5: in <module>
    import psycopg2
E   ImportError: No module named psycopg2

―――――――――――――――――――――― ERROR collecting tests/core/test_storage_migrations.py ―――――――――――――――――――――――
ImportError while importing test module '/home/rhubscher/mozilla/kinto/tests/core/test_storage_migrations.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/core/test_storage_migrations.py:8: in <module>
    from kinto.core.cache import postgresql as postgresql_cache
kinto/core/cache/postgresql/__init__.py:5: in <module>
    import psycopg2
E   ImportError: No module named psycopg2
```

kinto/core/cache/postgresql/__init__.py
```diff
@@ -1,12 +1,36 @@
...
 import os
+import time
 
 from kinto.core import logger
 from kinto.core.cache import CacheBase
 from kinto.core.storage.postgresql.client import create_from_config
+from kinto.core.storage.exceptions import BackendError
 from kinto.core.utils import json
 
+DELAY_BETWEEN_RETRIES_IN_SECONDS = 0.05
```
leplatrem: I think that `0.005` is enough

kinto/core/cache/postgresql/__init__.py
```diff
@@ -1,12 +1,36 @@
...
+        tries = kwargs.pop('tries', 0)
+        try:
+            return func(self, *args, **kwargs)
+        except psycopg2.IntegrityError as e:
+            if tries < MAX_RETRIES:
+                delay = (tries - 1) * DELAY_BETWEEN_RETRIES_IN_SECONDS
+                if delay > 0:
+                    time.sleep(delay)
+                return wraps_func(self, tries=(tries + 1), *args, **kwargs)
+            raise BackendError(original=e)
```
leplatrem: Why don't we raise the original exception? (please add comment if particular reason)

kinto/core/cache/postgresql/__init__.py
```diff
@@ -1,12 +1,36 @@
...
+    @wraps(func)
+    def wraps_func(self, *args, **kwargs):
+        tries = kwargs.pop('tries', 0)
+        try:
+            return func(self, *args, **kwargs)
+        except psycopg2.IntegrityError as e:
+            if tries < MAX_RETRIES:
+                delay = (tries - 1) * DELAY_BETWEEN_RETRIES_IN_SECONDS
+                if delay > 0:
+                    time.sleep(delay)
```
Natim: It is not `time.sleep(0)` the problem but `time.sleep(-0.5)`

kinto/core/cache/postgresql/__init__.py
```diff
@@ -1,12 +1,36 @@
...
+                return wraps_func(self, tries=(tries + 1), *args, **kwargs)
```
Natim: I tried without success and it felt more natural not to use a global state.

kinto/core/cache/postgresql/__init__.py
Natim: Actually the backends are not supposed to raise something else than `BackendError` because we want the user of the cache backend to be able to catch all exceptions regardless of the backend. I think we might have another bug to file about handling this for the cache backend.

kinto/core/cache/postgresql/__init__.py
leplatrem: Please add a comment.

kinto/core/cache/postgresql/__init__.py
leplatrem: Hum, I suggest that we either fix it everywhere or wait for another PR no?

Natim: Yes ok

PR #1037, opened 2017-01-26
Fix docs warning 

sunakshi96:
Fixes #

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.

r? @


docs/kinto-admin.rst
```diff
@@ -0,0 +1,260 @@
+Kinto Web Administration Console
+################################
+
+.. image:: images/kinto-admin-login.png
+    :align: center
+    :width: 800px
+
+
+
+A Web admin UI to manage data from a *Kinto* `<https://kinto.readthedocs.io/>`_ server: *Demo* `<http://kinto.github.io/kinto-admin/>`_.
```
gabisurita: You can probably point to https://kinto.dev.mozaws.net/v1/admin/ instead.

docs/kinto-admin.rst
```diff
@@ -0,0 +1,260 @@
...
+  (using fine-grained permissions)
+- Take advantage of
+  :ref:`schema validation <collection-json-schema>`
+  if you need it
+- :ref:`Track changes <api-history>` on your application data
+
+
+.. _enable:
+
+Prerequisites
```
gabisurita: I guess the next sections make more sense if we are developing Kinto Admin. If we are only using the plugin that is shipped with the server, which is already precompiled, I would simply give instructions on how to enable using a config file and a link to kinto admin repository if someone wants to customize or contribute.

docs/index.rst
```diff
@@ -11,6 +11,7 @@ Table of content
    :maxdepth: 1
 
    overview
+   kinto-admin
```
gabisurita: I think kinto-admin isn't that relevant to be in the second position of the documentation. I would probably put in the tutorial section, IDK.

sunakshi96:
Hey 
I I have just opened the another issue. And deleted the kinto-admin.rst file 
Thanks gabisurita :-)

sunakshi96:
I think we can handle the issue 'Document built-in plugin for kinto-admin" to describe the enabling and features of plugin. 
.....and the  plugin awaits to be rescued  :-)

PR #1041, opened 2017-01-27
Expose Kinto Admin settings built from kinto.ini

### LINKED ISSUE #1042: Configure Kinto Admin authentication policy from server settings.

Natim:
Fixes #1042 
Refs https://github.com/Kinto/kinto-admin/issues/313

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ This document describes changes between each past release.
 - Remove JSON Patch content-type from accepted types on the viewset, since it is handled
   in a separate view (#1031).
 - Upgraded to Kinto-Admin 1.8.1
+- Configure the Kinto Admin activated auth methods from the server configuration (#1042)
 
 
 5.3.0 (2017-01-20)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ This document describes changes between each past release.
 - Remove JSON Patch content-type from accepted types on the viewset, since it is handled
   in a separate view (#1031).
 - Upgraded to Kinto-Admin 1.8.1
-- Configure the Kinto Admin activated auth methods from the server configuration (#1042)
+- Configure the Kinto Admin auth methods from the server configuration (#1042)
 
 
 5.3.0 (2017-01-20)
--- a/kinto/core/decorators.py
+++ b/kinto/core/decorators.py
@@ -1,4 +1,5 @@
 from functools import update_wrapper
+from pyramid.response import Response
 
 
 class cache_forever(object):
@@ -10,5 +11,9 @@ class cache_forever(object):
     def __call__(self, request, *args, **kwargs):
         if self.saved is None:
             self.saved = self.wrapped(request, *args, **kwargs)
+            if isinstance(self.saved, Response):
+                self.saved = None
+                raise ValueError('cache_forever cannot cache Response only its body')
+
         request.response.write(self.saved)
         return request.response
--- a/kinto/core/decorators.py
+++ b/kinto/core/decorators.py
@@ -1,0 +1,14 @@
+from functools import update_wrapper
+
+
+class cache_forever(object):
+    def __init__(self, wrapped):
+        self.wrapped = wrapped
+        self.saved = None
+        update_wrapper(self, wrapped)
+
+    def __call__(self, request, *args, **kwargs):
+        if self.saved is None:
+            self.saved = self.wrapped(request, *args, **kwargs)
+        request.response.write(self.saved)
+        return request.response
--- a/kinto/core/decorators.py
+++ b/kinto/core/decorators.py
@@ -1,3 +1,4 @@
+import threading
 from functools import update_wrapper
 from pyramid.response import Response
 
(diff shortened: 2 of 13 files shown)


tests/plugins/test_admin.py
```diff
@@ -27,6 +27,12 @@ def test_admin_index_cat_be_reached(self):
         resp = self.app.get('/admin/')
         assert "html" in resp.body.decode('utf-8')
 
+    def test_admin_index_contains_authMethods_configurations(self):
+        self.maxDiff = None
+        resp = self.app.get('/admin/')
+        assert "authMethods" in resp.body.decode('utf-8')
+        assert "basicauth" in resp.body.decode('utf-8')
```
n1k0: µNit: could we decode the body once and for all?

kinto/plugins/admin/views.py
```diff
@@ -0,0 +1,26 @@
...
+    settings = {
+        "authMethods": aslist(request.registry.settings.get('multiauth.policies'))
+    }
+    globalSettings = "<script>window.globalSettings = %s;</script>" % json.dumps(settings)
+
+    with open(os.path.join(HERE, 'build/index.html')) as f:
+        RESPONSE_CACHED = f.read().replace('<script', globalSettings + '<script')
+
+    request.response.write(RESPONSE_CACHED)
+    return request.response
```
n1k0: Nit: it feels like we could factor altering and sending the response in a small reused function

kinto/plugins/admin/views.py
```diff
@@ -0,0 +1,26 @@
...
+
+from pyramid.settings import aslist
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+RESPONSE_CACHED = None
+
+
+# Configured home page
+def admin_home_view(request):
+    global RESPONSE_CACHED
```
n1k0: It's been a long time I've not been doing any Python but it's odd we need a global here...

kinto/plugins/admin/views.py
```diff
@@ -0,0 +1,26 @@
+import json
+import os
+
+from pyramid.settings import aslist
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+RESPONSE_CACHED = None
```
n1k0: As suggested while pairing, I'd rather use a template here.

```python
INDEX_TEMPLATE = open(os.path.join(HERE, 'build/index.html'), 'r').read()
# Note: I don't remember how you close a file pointer in one call, if it's even possible

def admin_home_view(request):
    # ...
    request.response.write(INDEX_TEMPLATE.replace('<script', globalSettings + '<script'))
    return request.response
```

By the way, it's really odd to mutate `request.response` and return it; is it actually it's supposed to be used?

kinto/plugins/admin/views.py
leplatrem: Yes, it would be a lot nicer with a decorator (that may already exist in pyramid)

CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ This document describes changes between each past release.
 
 - Remove JSON Patch content-type from accepted types on the viewset, since it is handled
   in a separate view (#1031).
+- Configure the Kinto Admin activated auth methods from the server configuration (#1042)
```
leplatrem: nit: `activated` sounds superfluous here

kinto/plugins/admin/views.py
```diff
@@ -0,0 +1,26 @@
...
+    if RESPONSE_CACHED is not None:
+        request.response.write(RESPONSE_CACHED)
+        return request.response
+
+    settings = {
+        "authMethods": aslist(request.registry.settings.get('multiauth.policies'))
+    }
+    globalSettings = "<script>window.globalSettings = %s;</script>" % json.dumps(settings)
+
+    with open(os.path.join(HERE, 'build/index.html')) as f:
```
leplatrem: Maybe we should rename `index.html` into `index.tpl` or `_index.html`, otherwise it can be confusing

kinto/plugins/admin/views.py
Natim: That's a way of doing it, we could also recreate an object.

kinto/plugins/admin/views.py
Natim: Actually that's the name of the file that is generated by `react-scripts`.

kinto/plugins/admin/views.py
Natim: To be honest, I am a bit reluctant to add http://docs.pylonsproject.org/projects/pyramid_beaker/en/latest/ just for that...

kinto/core/decorators.py
```diff
@@ -0,0 +1,19 @@
+from functools import update_wrapper
+from pyramid.response import Response
+
+
+class cache_forever(object):
```
leplatrem: note: we had the `synchronized` decorator in `kinto.core.utils`

Natim: Do you want me to move all decorators inside the new file or are you suggesting that we should move this class in the utils file? I put it there because that's an exposed feature of kinto.core rather than an utility for kinto.core

kinto/core/decorators.py
```diff
@@ -0,0 +1,19 @@
...
+        self.wrapped = wrapped
+        self.saved = None
+        update_wrapper(self, wrapped)
+
+    def __call__(self, request, *args, **kwargs):
+        if self.saved is None:
+            self.saved = self.wrapped(request, *args, **kwargs)
+            if isinstance(self.saved, Response):
+                self.saved = None
+                raise ValueError('cache_forever cannot cache Response only its body')
```
leplatrem: Seeing the level of care here, I assume you want to introduce this as part of the core API. You could hence add it the API docs of kinto.core.

PR #1046, opened 2017-01-28
Set permission fields on initialization rather than request time

gabisurita:
Related to #1033

r? @glasserc @leplatrem


kinto/core/resource/schema.py
```diff
@@ -74,34 +74,38 @@ class PermissionsSchema(colander.SchemaNode):
...
 
-            # Add a String list child node with the name of ``perm``.
-            self.add(self._get_node_principals(perm))
+        if cstruct in (colander.null, colander.drop):
+            return super(PermissionsSchema, self).deserialize(cstruct)
 
-        # End up by deserializing a mapping whose keys are now known.
-        return super(PermissionsSchema, self).deserialize(permissions)
+        # If permissions are listed, check fields and produce fancy error messages
+        elif self.known_perms:
```
leplatrem: superfluous `elif` 'cuz of the return above

kinto/core/resource/schema.py
```diff
@@ -74,34 +74,38 @@ class PermissionsSchema(colander.SchemaNode):
...
+            for perm in cstruct:
+                colander.OneOf(choices=self.known_perms)(self, perm)
```
leplatrem: Why is this necessary if we have `unknown: raise` ?

kinto/core/resource/schema.py
```diff
@@ -74,34 +74,38 @@ class PermissionsSchema(colander.SchemaNode):
...
+            return super(PermissionsSchema, self).deserialize(cstruct)
+
+        # Else deserialize the fields that are not on the schema
+        permissions = {}
+        for perm, principals in cstruct.items():
+            node = colander.SequenceSchema(colander.SchemaNode(colander.String()),
+                                           missing=colander.drop).deserialize(principals)
```
leplatrem: Why `missing: drop` ?

It's better if you instantiate the schema once out of the loop IMO

kinto/core/resource/schema.py
```diff
@@ -74,34 +74,38 @@ class PermissionsSchema(colander.SchemaNode):
...
-            # Add a String list child node with the name of ``perm``.
-            self.add(self._get_node_principals(perm))
+        if cstruct in (colander.null, colander.drop):
+            return super(PermissionsSchema, self).deserialize(cstruct)
 
-        # End up by deserializing a mapping whose keys are now known.
-        return super(PermissionsSchema, self).deserialize(permissions)
+        # If permissions are listed, check fields and produce fancy error messages
+        elif self.known_perms:
+            for perm in cstruct:
```
leplatrem: Don't we need to check that `cstruct` is iterable?

kinto/core/resource/schema.py
```diff
@@ -74,34 +74,38 @@ class PermissionsSchema(colander.SchemaNode):
...
+                colander.OneOf(choices=self.known_perms)(self, perm)
+            return super(PermissionsSchema, self).deserialize(cstruct)
+
+        # Else deserialize the fields that are not on the schema
+        permissions = {}
+        for perm, principals in cstruct.items():
```
leplatrem: Are we sure here that `cstruct` is a mapping?

kinto/core/resource/schema.py
gabisurita: This preserve the error message used previously. e.g:
"permissions in body: \"foo\" is not one of read, write, collection:create, group:create"

Without it we just get an unrecognized fields message. We can set the error massage manually, but then this can override other validation messages.

kinto/core/resource/schema.py
gabisurita: Yes, we do. I should probably add a test for that too.

PR #1047, opened 2017-01-29
Split request schemas for each method/endpoint type

gabisurita:
Related to #1033

This PR changes the schema-viewset interaction:
- Request schemas are now explicitly declared for each method-endpoint type
- We introduce deferred fields on schemas that allow binding. 
   e.g: we now have a generic RequestSchema where we can bind different QueryString Schemas.

Todos

- [x] Add tests.
- [x] Add a changelog entry.
- [x] Wait for colander release

r? @glasserc @leplatrem


kinto/core/resource/schema.py
```diff
@@ -215,36 +221,41 @@ def deserialize(self, cstruct=colander.null):
...
-    response_behaviour = HeaderField(colander.String(), name='Response-Behavior',
-                                     validator=response_behavior_validator())
 
     @staticmethod
     def schema_type():
         return colander.Mapping(unknown='preserve')
 
 
+class PatchHeaderSchema(HeaderSchema):
+    """Header schema used with patch requests."""
```
leplatrem: nitpick: http verbs are usually written upercase

kinto/core/resource/schema.py
```diff
@@ -305,12 +338,43 @@ class JsonPatchBodySchema(colander.SequenceSchema):
...
+
+    querystring = CollectionQuerySchema()
+
+
+class CollectionGetRequestSchema(CollectionRequestSchema):
+    """Collection get request schema."""
+
+    @colander.instantiate()
+    class querystring(CollectionQuerySchema, GetQuerySchema):
+        pass
```
leplatrem: I had never seen this :) Interesting... It took me a while to figure out why `_fields` was not listed in `CollectionQuerySchema`!

Maybe there is a matter of wording or comments to really distinguish : general (both endpoints), plural and single enpoints

kinto/core/resource/viewset.py
```diff
@@ -118,14 +128,17 @@ def get_view_arguments(self, endpoint_type, resource_cls, method):
         endpoint_args = getattr(self, by_method, {})
         args.update(**endpoint_args)
 
+        request_schema = args.get('schema', RequestSchema)
+
+        # Instantiate request schema
+        request_schema = request_schema()
+
         if method.lower() in map(str.lower, self.validate_schema_for):
```
leplatrem: Note: is this `validate_schema_for` still relevant with the changes of this PR? Maybe we can get rid of it.

kinto/core/resource/viewset.py
```diff
@@ -118,14 +128,17 @@ def get_view_arguments(self, endpoint_type, resource_cls, method):
...
-            schema = RequestSchema()
             record_schema = self.get_record_schema(resource_cls, method)
             record_schema.name = 'body'
-            schema.add(record_schema)
-            args['schema'] = schema
-        else:
-            args['schema'] = RequestSchema()
+            request_schema.add(record_schema)
```
leplatrem: With all the explicit declarations that we do elsewhere, I wish those few lines become extremely simple

kinto/core/resource/schema.py
Natim: I call it the Mixin design pattern :)

kinto/core/resource/viewset.py
gabisurita: With the current implementation it still is, the `body` field is still added just on the viewset. I've tried to preserve this for backward compatibility, but we can indeed solve this on the schemas.

kinto/core/resource/viewset.py
gabisurita: An interesting approach IMO would be to handle the `record_schema` binding directly on the schema (instead of the viewset), what do you think?

gabisurita:
> The class declarations seem a bit verbose but I don't have any alternative proposition. Except maybe have some kind of helper

An interesting approach for that would be to set the header, query and body on when instantiating a request. e.g:
```python
 request_schema = RequestSchema(header=HeaderSchema(), querystring=QuerySchema())
```

gabisurita:
Also, do we have to keep SimpleSchema, PartialSchema and StrictSchema on the viewset?

kinto/core/resource/schema.py
gabisurita: `_fields` is not accepted in plural DELETE, so that's why it's not there. I've tried to omit "General" and just add the Collection or Record according to each type. Turns out Record was not used. :P

kinto/core/resource/schema.py
gabisurita: Do you think the verb on the class name should be uppercase as well?

leplatrem:
> better to make this explicit.

Works for me!

> keep SimpleSchema, PartialSchema and StrictSchema on the viewset?

If not used, and not mentionned in docs, I'm not again removing them :)

kinto/core/resource/schema.py
leplatrem: No, in comments at least

kinto/core/resource/schema.py
leplatrem: Don't hesitate to add as many comments as you can for this kind of details :)

kinto/core/resource/viewset.py
leplatrem: I didn't understand your proposition :(

gabisurita:
@leplatrem, I tried this different approach using schema binding, I think it makes the viewset a lot simpler, what do you think?

tests/core/resource/test_schema.py
```diff
@@ -213,3 +213,84 @@ def test_querystring_preserve_unkown_fields(self):
     def test_drops(self):
         deserialized = self.schema.deserialize({})
         self.assertEquals(deserialized, {})
+
+
+class GetRequestSchemaTest(unittest.TestCase):
+
+    def setUp(self):
+        self.schema = schema.RequestSchema().bind(
+            querystring=schema.RecordGetQuerySchema())
```
leplatrem: nitpick: prefer an intermediary variable to avoid indentation

kinto/core/resource/schema.py
```diff
@@ -278,6 +289,62 @@ def deserialize(self, cstruct=colander.null):
...
+
+    _fields = FieldList()
+
+
+# Body Schemas
+
+
+class RecordSchema(colander.MappingSchema):
+
+    data = colander.deferred
```
leplatrem: Add a comment to explain that we decide later what to put in data based on the resource schema?

kinto/core/resource/schema.py
```diff
@@ -305,12 +372,38 @@ class JsonPatchBodySchema(colander.SequenceSchema):
...
-    """Baseline schema for kinto requests."""
+    """Base schema for kinto requests."""
+
+    header = HeaderSchema()
+    querystring = QuerySchema()
+
+    def bind(self, header=None, querystring=None, **kwargs):
+        binded = super(RequestSchema, self).bind(**kwargs)
+        if header:
+            binded['header'] = header.bind(**kwargs)
```
leplatrem: Do we really want to support that `.bind(**kwargs)`. I find it cool and flexible, but it looks like we don't use it and it does not ease the understanding of the code.

kinto/core/resource/schema.py
```diff
@@ -305,12 +372,38 @@ class JsonPatchBodySchema(colander.SequenceSchema):
...
 
+# Request schemas
+
+
 class RequestSchema(colander.MappingSchema):
-    """Baseline schema for kinto requests."""
+    """Base schema for kinto requests."""
+
+    header = HeaderSchema()
+    querystring = QuerySchema()
```
leplatrem: Why are they not deferred in this case?

gabisurita:
> I just don't know if bind() was made for such a use case. Maybe we can ask somewhere?

I've asked #pyramid channel on IRC and they said that it was, and that there should be an easier way to do it. I've submitted a patch to colander which should be on the next release. https://github.com/Pylons/colander/pull/280

kinto/core/resource/schema.py
gabisurita: Because they have a default.

tests/core/resource/test_viewset.py
```diff
@@ -119,12 +114,12 @@ def test_class_parameters_are_used_for_collection_arguments(self, mocked):
...
         )
 
-        arguments = viewset.collection_arguments(mock.MagicMock, 'get')
+        arguments = viewset.collection_arguments(mock.MagicMock(), 'get')
 
         self.assertDictEqual(
             arguments,
             {
-                'schema': mocked(),
+                'schema': mocked().bind(),
```
gabisurita: This is the return value if we call `bind()` on a mocked schema. If we want to avoid this, we should also mock patch the bind value. egs.
```python
>>> m = mock.MagicMock()
>>> m.method()
<MagicMock name='mock.method()' id='4417348560'>
```
And a proposed solution:
```
>>> m = mock.MagicMock()
>>> m.method = lambda: m
>>> m.method()
<MagicMock id='4417096080'>
```

tests/core/resource/test_viewset.py
```diff
@@ -355,18 +349,26 @@ def test_is_endpoint_enabled_returns_true_for_head_if_readonly(self):
...
 
 class ShareableViewSetTest(unittest.TestCase):
+
     def test_permission_dynamic_is_set_by_default(self):
         viewset = ShareableViewSet()
-        args = viewset.collection_arguments(mock.sentinel.resource, 'GET')
+        resource = mock.sentinel.resource
+        resource.schema = ResourceSchema
+        resource.permissions = ('read', 'write')
+        args = viewset.collection_arguments(resource, 'GET')
```
gabisurita: It failed on the second bind as one call from colander that inserts a node would not accept a mock object, so I've set the `ResourceSchema` manually. I'll try to find a better approach tough.

tests/core/resource/test_viewset.py
```diff
@@ -53,43 +55,36 @@ def test_arguments_are_merged_on_initialization(self):
...
         viewset = ViewSet()
-        args = viewset.collection_arguments(mock.sentinel.resource, 'GET')
+        args = viewset.collection_arguments(mock.MagicMock(), 'GET')
         self.assertEquals(args['permission'], 'private')
 
     def test_schema_is_added_when_method_matches(self):
-        viewset = ViewSet(
-            validate_schema_for=('GET', )
-        )
+        viewset = ViewSet()
```
gabisurita: Yes, we are. It was used to tell if we added a body to the request schema or not, but it doesn't make sense anymore as we split the requests into `RequestSchema` and `PayloadRequestSchema`.

We could consider this a breaking change because it can be changed from a resource, but I can't see any reason why someone would change this attribute, because probably the requests would fail if it was different.

tests/core/resource/test_viewset.py
gabisurita: Anyways, it's not a problem anymore with the new bind implementation.

leplatrem:
This is excellent work. You never stop raising the bar 👍 👍

tests/core/resource/test_viewset.py
glasserc: Sorry, I had misread this test the first time around!

PR #1049, opened 2017-01-30
Remove pypy support

Natim:
Remove PYPY supports.

This is the first step before removing supports for py27 alltogether.


gabisurita:
I'm Just curious... Has anyone tried kinto with pypy3?

Natim:
No I didn't

PR #1054, opened 2017-02-01
Move generic schemas from Resource to kinto.core

gabisurita:
Related to #1033 

This allow importing generic schemas on other parts of kinto core without breaking the import order.

- [x] Add a CHANGELOG entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,8 +10,17 @@ This document describes changes between each past release.
 
 - Remove pypy supports. (#1049)
 
+**Internal changes**
 
-5.3.2 (unreleased)
+- Permission schema children fields are now set during initialization instead of on
+  deserialization (#1046).
+- Request schemas (including validation and deserialization) are now isolated by method
+  and endpoint type (#1047).
+- Move generic API schemas (e.g TimeStamps and HeaderFields) from `kinto.core.resource.schema`
+  to a sepate file on `kinto.core.schema`.
+
+
+5.3.2 (2017-01-31)
 ------------------
 
 **Bug fixes**
@@ -30,12 +39,8 @@ This document describes changes between each past release.
 
 - Remove JSON Patch content-type from accepted types on the viewset, since it is handled
   in a separate view (#1031).
-- Permission schema children fields are now set during initialization instead of on
-  deserialization (#1046).
 - Upgraded to Kinto-Admin 1.8.1
 - Configure the Kinto Admin auth methods from the server configuration (#1042)
-- Request schemas (including validation and deserialization) are now isolated by method
-  and endpoint type (#1047).
 
 5.3.0 (2017-01-20)
 ------------------
--- a/kinto/core/schema.py
+++ b/kinto/core/schema.py
@@ -1,8 +1,7 @@
-"""
-This module contains generic schemas used by the core. You should declare schemas that
+"""This module contains generic schemas used by the core. You should declare schemas that
 may be reused across the `kinto.core` here.
 
-.. note::
+.. note:: This module is reserve for generic schemas only.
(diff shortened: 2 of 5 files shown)


CHANGELOG.rst
```diff
@@ -10,8 +10,17 @@ This document describes changes between each past release.
...
 - Remove pypy supports. (#1049)
 
+**Internal changes**
+
+- Permission schema children fields are now set during initialization instead of on
+  deserialization (#1046).
+- Request schemas (including validation and deserialization) are now isolated by method
+  and endpoint type (#1047).
+- Move generic API schemas (e.g TimeStamps and HeaderFields) from `kinto.core.resource.schema`
+  to a sepate file on `kinto.core.schema`.
```
leplatrem: I remember we had that before, and we changed... What does the docs say? Will the old location still be accesible? deprecated?

leplatrem: Yes, see https://github.com/mozilla-services/cliquet/blob/master/cliquet/schema.py

gabisurita: Hmm, that's interesting. I agree `ResourceSchema` should be on `kinto.core.resource` (and I'm keeping it there), but I'm not so sure about other things like Timestamps or Header fields that could be reused in other parts of the core.

Most of these schemas are imported on `kinto.core.resource` so they are still accessible. Only Timestamp and URL aren't, and they aren't used anywhere on Kinto. Do you know why they are there?

leplatrem: > Do you know why we have them there?

Originally ([archeology](https://github.com/mozilla-services/readinglist/blob/master/readinglist/views/article.py)) they were in core because we thought they were common types of fields. But now they're just used in third parties or [plugins](https://github.com/mozilla-services/kinto-fxa/blob/master/kinto_fxa/views/relier.py#L19). We can deprecated them, no worries.

gabisurita: > We can deprecated them, no worries.

OK. I don't have a strong opinion about it. We can keep them too. In case we deprecate or move them, do you think we should add warning messages or we can just make it a breaking change?

gabisurita: About the documentation: I couldn't fond mentions to these schemas anywhere. Also [in some parts](http://kinto.readthedocs.io/en/stable/core/resource.html?highlight=kinto.core.resource.schema#full-example) we use a colander String with url validator to implement an url instead of the URL schema.

kinto/core/schema.py
```diff
@@ -0,0 +1,110 @@
+import six
```
glasserc: It might be nice to have a module-level docstring explaining what's supposed to go in this file vs. what goes in `kinto.core.resource.schema`.

gabisurita: I don't see many module-level docstrings on `kinto.core` (`__init__.py` is the only exception), that's why I didn't use them, but I agree it would be nice to have this written in the code somewhere.

CHANGELOG.rst
```diff
@@ -35,12 +44,8 @@ This document describes changes between each past release.
 
 - Remove JSON Patch content-type from accepted types on the viewset, since it is handled
   in a separate view (#1031).
-- Permission schema children fields are now set during initialization instead of on
-  deserialization (#1046).
 - Upgraded to Kinto-Admin 1.8.1
 - Configure the Kinto Admin auth methods from the server configuration (#1042)
-- Request schemas (including validation and deserialization) are now isolated by method
-  and endpoint type (#1047).
```
leplatrem: Are these deletions expected?

gabisurita: Yeah, these are unreleased. They probably got there by a problem in some auto merge.

PR #1056, opened 2017-02-01
Upgrade to PostgreSQL 9.5

### LINKED ISSUE #1055: IntegrityError with Posgresql cache backend
I would like to update the current request with one of http://stackoverflow.com/a/1109198/186202

Natim:
Fixes #1055

r? @leplatrem

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,16 +6,15 @@ This document describes changes between each past release.
 6.0.0 (unreleased)
 ------------------
 
+**Breaking changes**
+
+**PostgreSQL 9.5**
+
+- Update the upsert query to use an INSERT or UPDATE on CONFLICT behavior (fixes #1055)
+
 **Internal changes**
 
 - Remove pypy supports. (#1049)
-
-5.3.3 (unreleased)
-------------------
-
-**Bug fixes**
-
-- Update the upsert query to use an INSERT or UPDATE on CONFLICT behavior (fixes #1055)
 
 
 5.3.2 (2017-01-31)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,12 +8,11 @@ This document describes changes between each past release.
 
 **Breaking changes**
 
-**PostgreSQL 9.5**
-
-- Update the upsert query to use an INSERT or UPDATE on CONFLICT behavior (fixes #1055)
+- Upgraded to PostgreSQL 9.5
 
 **Internal changes**
 
+- Update the upsert query to use an INSERT or UPDATE on CONFLICT behavior (fixes #1055)
 - Remove pypy supports. (#1049)
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,8 +10,15 @@ This document describes changes between each past release.
 
 - Remove pypy supports. (#1049)
 
+5.3.3 (unreleased)
+------------------
 
-5.3.2 (unreleased)
+**Bug fixes**
+
+- Update the upsert query to use an INSERT or UPDATE on CONFLICT behavior (fixes #1055)
+
+
+5.3.2 (2017-01-31)
 ------------------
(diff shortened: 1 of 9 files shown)


README.rst
```diff
@@ -37,4 +37,4 @@ Requirements
 ------------
 
 * **Python**: 2.7, 3.4+
-* **Backends**: In-memory (development), Postgresql 9.4+ (production)
+* **Backends**: In-memory (development), Postgresql 9.5+ (production)
```
leplatrem: This is a breaking change. I suggest that we keep two queries depending on the version of postgresql being used instead of breaking compat

Natim: Do we have a way to check that programatically at the Python level?

leplatrem: If we look [at the logs](https://gist.github.com/leplatrem/54637f78e6297ba8d5ba0b50d7447042), a part of the code already introspects it (see `select version()`).
I'll loook into SQLAlchemy...

leplatrem: `engine.server_version_info`
https://github.com/zzzeek/sqlalchemy/blob/bc4a96836d5bfb911da26f7e77119f3a7b356f2e/lib/sqlalchemy/engine/default.py#L244-L245

leplatrem:
My proposition (instead of breaking compatibility with 9.4) would be to:
* Take your fix using `ON CONFLICT` when running PostgreSQL 9.5
* Keep an issue opened and mention the known problem with cache and 9.4
* Later, when releasing version 6.0 we can drop 9.4 support (there are other things to change, like create index if not exists etc.)

Natim:
Can we do your strategy for 5.3.3 and keep the 9.5+ version only for kinto 6.0?

CHANGELOG.rst
```diff
@@ -6,12 +6,18 @@ This document describes changes between each past release.
 6.0.0 (unreleased)
 ------------------
 
+**Breaking changes**
+
+**PostgreSQL 9.5**
```
leplatrem: Shouldn't we be more explicit like `PostgreSQL 9.5 is now required for the cache backend`?

CHANGELOG.rst
```diff
@@ -6,12 +6,18 @@ This document describes changes between each past release.
...
+
+- Update the upsert query to use an INSERT or UPDATE on CONFLICT behavior (fixes #1055)
```
leplatrem: This one is technically a bug fix ;)

kinto/core/cache/postgresql/__init__.py
```diff
@@ -149,13 +149,11 @@ def set(self, key, value, ttl=None):
...
-            UPDATE cache SET value = :value, ttl = sec2ttl(:ttl)
-             WHERE key=:key
-            RETURNING *)
         INSERT INTO cache (key, value, ttl)
-        SELECT :key, :value, sec2ttl(:ttl)
-        WHERE NOT EXISTS (SELECT * FROM upsert)
+        VALUES (:key, :value, sec2ttl(:ttl))
+        ON CONFLICT (key) DO UPDATE
+        SET value = :value,
+            ttl = sec2ttl(:ttl);
```
leplatrem: We could remove the retry decorator I think, because it hasn't proven to be useful...

leplatrem:
Should we rename that PR into *Upgrade to PostgreSQL 9.5* ? And change the index creation too?

Natim:
> Should we rename that PR into Upgrade to PostgreSQL 9.5 ? And change the index creation too?

Yes, can I let you update the Index creation?

README.rst
Natim: How do you get the engine from the client?

Natim:
@leplatrem final r+?

leplatrem:
> @leplatrem final r+?

I would just improve the CHANGELOG a bit with a sentence instead of just *Postgresql 9.5*, no? This is not a blocker though, so r+ :)

PR #1058, opened 2017-02-02
Fix IntegrityError for 5.3.X branch

### LINKED ISSUE #1055: IntegrityError with Posgresql cache backend

Natim:
Fixes #1055

### DIFF
--- a/kinto/core/cache/postgresql/__init__.py
+++ b/kinto/core/cache/postgresql/__init__.py
@@ -169,10 +169,10 @@ class Cache(CacheBase):
         params = dict(key=self.prefix + key, value=value, ttl=ttl)
         with self.client.connect() as conn:
             server_version = conn.connection().dialect.server_version_info
-            if server_version < (9, 5):
-                conn.execute(query_pg_94, params)
-            else:
+            if server_version >= (9, 5):
                 conn.execute(query, params)
+            else:
+                conn.execute(query_pg_94, params)
 
     def get(self, key):
         purge = "DELETE FROM cache WHERE ttl IS NOT NULL AND now() > ttl;"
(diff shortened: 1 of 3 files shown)


kinto/core/cache/postgresql/__init__.py
```diff
@@ -157,10 +158,24 @@ def set(self, key, value, ttl=None):
...
+        INSERT INTO cache (key, value, ttl)
+        SELECT :key, :value, sec2ttl(:ttl)
+        WHERE NOT EXISTS (SELECT * FROM upsert);
+        """
         value = json.dumps(value)
+        params = dict(key=self.prefix + key, value=value, ttl=ttl)
         with self.client.connect() as conn:
-            conn.execute(query, dict(key=self.prefix + key,
-                                     value=value, ttl=ttl))
+            server_version = conn.connection().dialect.server_version_info
```
leplatrem: **nitpick:** here https://github.com/zzzeek/sqlalchemy/blob/bc4a96836d5bfb911da26f7e77119f3a7b356f2e/lib/sqlalchemy/testing/exclusions.py#L384 they use `getattr` so maybe we should use it too and compare `version >= (9, 5)`

leplatrem:
Quick note on maintainance: I'd rather continue to submit patches to the `master` branch. Then, when we want to release 5.3.X, we *backport* the pull-request (usually with a simple `git cherry-pick -m 1 <merge-commit>`). Duplicating the PRs like this and #1056 can end-up being confusing IMO

Natim:
I agree with you generally but for this specific case the patch are different between the master 6.X and the 5.X branch

kinto/core/cache/postgresql/__init__.py
Natim: I don't think we need to use getattr there because we are always using the pgdialect that will always define it.

leplatrem:
> patch are different between the master 6.X and the 5.X branch

Gotcha, but maybe we can handle the situation in the content of the commit that backports the change? (*not sure what's best, let's see next time...*).

PR #1064, opened 2017-02-10
Fix crash on If-Match: * (fixes #1063)

### LINKED ISSUE #1063: If-Match:* behaviour is unclear (and crashes) with existing record
On 5.2 it would raise a "400" on existing record.

On current master, it will crash in "500" as it's not handled by `kinto.core.resource` anymore, but it's considered a valid header value by deserialization.

I think we should set the behavior to just ignore it. and process the request anyways. Opinions?

gabisurita:
Fixes #1063 

Fix crash and previous '400' by always ignoring precondition If-Match:*

- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ This document describes changes between each past release.
 **Bug fixes**
 
 - Prevent injections in the PostgreSQL permission backend
+- Fix crash on ``If-Match: *``
 
 **Internal changes**
 
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -838,7 +838,7 @@ class UserResource(object):
                 # Tombstones should not prevent creation.
                 return
             modified_since = -1  # Always raise.
-        elif if_match:
+        elif if_match and if_match != '*':
             modified_since = if_match
         else:
             # In case _raise_304_if_not_modified() did not raise.
--- a/tests/core/resource/test_preconditions.py
+++ b/tests/core/resource/test_preconditions.py
@@ -196,6 +196,14 @@ class ModifiedMeanwhileTest(BaseTest):
                 'field': 'new'}}
         self.resource.collection_post()  # not raising.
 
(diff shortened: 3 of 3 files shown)


gabisurita:
@Natim Hmm. I don't get it. Why `If-Match:*` is invalid? It's just a convention used in Kinto? I thought we should accept the valid values for If-Match on RFC 2616.

leplatrem:
Yes indeed it is official:

```

   The meaning of "If-Match: *" is that the method SHOULD be performed
   if the representation selected by the origin server (or by a cache,
   possibly using the Vary mechanism, see section 14.44) exists, and
   MUST NOT be performed if the representation does not exist.
```

This spec is so hard to grasp...

Natim:
You are right my mistake.

glasserc:
Sorry to ask stupid questions, but does this change follow the spec then? Or is it just to stop a crash and we'll try to conform to the spec later? It seems like just ignores `If-Match: *` rather than checking whether the record exists or doesn't exist.

Natim:
It does follow the spec. If you want to check if the records exists or not your would use `If-None-Match: *` Here we are talking about `If-Match: *` which means any value.

gabisurita:
> but does this change follow the spec then?

I don't think it does. From my understanding of the following paragraph it just fixes the crash. IMO if we want to follow the spec correctly we should raise 412 on non existing resource, but not only on `*`, but also on other values of `If-Match`.

> If none of the entity tags match, or if "*" is given and no current entity exists, the server MUST NOT perform the requested method, and MUST return a 412 (Precondition Failed) response.

Natim:
That behaviour was never implemented so I guess we can merge this and create a new PR to implement it.

Natim:
I have created https://github.com/Kinto/kinto/issues/1069

gabisurita:
> That behaviour was never implemented so I guess we can merge this and create a new PR to implement it.

Agree. It would also mean a breaking change on the API.

Natim:
> It would also mean a breaking change on the API.

Yes but it would still be backward compatible I guess.

PR #1071, opened 2017-02-13
Add permissions backend SQL injections tests.

### LINKED ISSUE #1070: Add a unit test to assert #1061 has been fixed.

Natim:
Fixes #1070

- [x] Add tests.

### DIFF
--- a/tests/test_views_objects_permissions.py
+++ b/tests/test_views_objects_permissions.py
@@ -109,6 +109,7 @@ class CollectionPermissionsTest(PermissionsTest):
         self.app.get("/buckets/beer'/groups/barley", headers=self.headers, status=403)
 
         self.app.get("/buckets/beer/collections/barley'", headers=self.headers, status=400)
+        # XXX: We should validate the collection ID on the records collection endpoint. #1077
         self.app.get("/buckets/beer/collections/barley'/records", headers=self.headers, status=404)
 
         self.app.get("/buckets/beer/groups/barley'", headers=self.headers, status=400)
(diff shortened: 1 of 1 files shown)


tests/test_views_objects_permissions.py
```diff
@@ -103,6 +103,16 @@ def test_cannot_write_if_not_allowed(self):
                           headers=self.alice_headers,
                           status=403)
 
+    def test_permission_backend_prevent_sql_injections(self):
+        self.app.get("/buckets/beer'", headers=self.headers, status=403)
+        self.app.get("/buckets/beer'/collections/barley", headers=self.headers, status=403)
+        self.app.get("/buckets/beer'/groups/barley", headers=self.headers, status=403)
+
+        self.app.get("/buckets/beer/collections/barley'", headers=self.headers, status=400)
+        self.app.get("/buckets/beer/collections/barley'/records", headers=self.headers, status=404)
```
leplatrem: interesting difference that would a deserve a comment or XXX : )

Natim: Yes it looks like a bug to me actually.

Natim: Oh no its not a bug because we have the permissions on the beer bucket so it makes sense.

leplatrem: We can add a comment for mental health of the reader

Natim: Oh no you are right we should probably say that the ID is wrong even if we are on the collection record endpoint? I don't know.

PR #1076, opened 2017-02-14
Prevent 500 if timestamp is to big in QueryString.

### LINKED ISSUE #1067: _since overflow error: long too big to convert
When since is too big with the Postgresql backend it fails with an OverflowError: https://sentry.prod.mozaws.net/operations/kinto-prod-1/issues/382250/

Natim:
Fixes #1067 

- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem

### DIFF
--- a/kinto/core/resource/schema.py
+++ b/kinto/core/resource/schema.py
@@ -212,18 +212,19 @@ class QuerySchema(colander.MappingSchema):
 class CollectionQuerySchema(QuerySchema):
     """Querystring schema used with collections."""
 
-    _limit = QueryField(colander.Integer())
+    _limit = QueryField(colander.Integer(),
+                        validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
     _sort = FieldList()
     _token = QueryField(colander.String())
     _since = QueryField(colander.Integer(),
-                        validator=colander.Range(max=POSTGRESQL_MAX_INTEGER_VALUE))
+                        validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
     _to = QueryField(colander.Integer(),
-                     validator=colander.Range(max=POSTGRESQL_MAX_INTEGER_VALUE))
+                     validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
     _before = QueryField(colander.Integer(),
-                         validator=colander.Range(max=POSTGRESQL_MAX_INTEGER_VALUE))
+                         validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
     id = QueryField(colander.String())
     last_modified = QueryField(colander.Integer(),
-                               validator=colander.Range(max=POSTGRESQL_MAX_INTEGER_VALUE))
+                               validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
 
 
 class RecordGetQuerySchema(QuerySchema):
--- a/kinto/core/resource/schema.py
+++ b/kinto/core/resource/schema.py
@@ -6,6 +6,8 @@ from kinto.core.schema import (Any, HeaderField, QueryField, HeaderQuotedInteger
                                FieldList, TimeStamp, URL)
 from kinto.core.utils import native_value
 
+POSTGRESQL_MAX_INTEGER_VALUE = 9223372036854775807
+
 
 class TimeStamp(TimeStamp):
     """This schema is deprecated, you shoud use `kinto.core.schema.TimeStamp` instead."""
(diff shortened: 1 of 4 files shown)


tests/core/resource/test_schema.py
```diff
@@ -288,3 +288,52 @@ def test_body_supports_binding_after_other_binds(self):
         deserialized = bound.deserialize({})
         self.assertEquals(deserialized['querystring'], {'foo': 'bar'})
         self.assertEquals(deserialized['body'], {'foo': 'beer'})
+
+
+class CollectionQuerySchemaTest(unittest.TestCase):
+
+    def setUp(self):
+        self.schema = schema.CollectionQuerySchema()
+        self.record = {
```
gabisurita: I wouldn't call this record, maybe `querystring`?

kinto/core/resource/schema.py
```diff
@@ -213,11 +215,15 @@ class CollectionQuerySchema(QuerySchema):
     _limit = QueryField(colander.Integer())
```
gabisurita: Shouldn't we do the same for limit?

Natim: Well limit is bound by the config max limit so it shouldn't be a problem, also I don't know if LIMIT is supposed to be a BIGINT or of another kind. I might worth to have a look.

gabisurita: Also, with the current implementation we also accept negative values, which can overflow as well.

kinto/core/resource/schema.py
```diff
@@ -210,14 +212,19 @@ def deserialize(self, cstruct=colander.null):
...
+    _since = QueryField(colander.Integer(),
+                        validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
+    _to = QueryField(colander.Integer(),
+                     validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
+    _before = QueryField(colander.Integer(),
+                         validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
     id = QueryField(colander.String())
-    last_modified = QueryField(colander.Integer())
+    last_modified = QueryField(colander.Integer(),
+                               validator=colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE))
```
leplatrem: Instantiate this validator only once (e.g. `positiveInteger`)

kinto/core/resource/schema.py
```diff
@@ -6,6 +6,8 @@
                                FieldList, TimeStamp, URL)
 from kinto.core.utils import native_value
 
+POSTGRESQL_MAX_INTEGER_VALUE = 9223372036854775807
```
leplatrem: `2**64 / 2`

kinto/core/resource/schema.py
```diff
@@ -6,7 +6,9 @@
                                FieldList, TimeStamp, URL)
 from kinto.core.utils import native_value
 
-POSTGRESQL_MAX_INTEGER_VALUE = 9223372036854775807
+POSTGRESQL_MAX_INTEGER_VALUE = 2**64 / 2
+
+positiveBigInteger = colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE)
```
leplatrem: It shouldn't be camel case no ? I'm sorry to have suggested that name :(

Natim: Do you want it as a CONSTANT uppercase?

leplatrem: Hum, now I don't know anymore :D 

A constant seems weird for a reusable instance no ? 
Maybe like a class name ? Like `colander.Length` ? Or just `positive_integer` like a variable?

leplatrem: This is bikeshed, it's fine whatever way

kinto/core/resource/schema.py
```diff
@@ -6,6 +6,10 @@
                                FieldList, TimeStamp, URL)
 from kinto.core.utils import native_value
 
+POSTGRESQL_MAX_INTEGER_VALUE = 2**64 / 2
+
+POSITIVE_BIG_INTEGER = colander.Range(min=0, max=POSTGRESQL_MAX_INTEGER_VALUE)
```
gabisurita: In other parts of code we used lowercase for validators, but anything is fine, i guess.

kinto/core/resource/schema.py
```diff
@@ -6,6 +6,10 @@
                                FieldList, TimeStamp, URL)
 from kinto.core.utils import native_value
 
+POSTGRESQL_MAX_INTEGER_VALUE = 2**64 / 2
```
gabisurita: The tests are probably failing because Python3 implement default float casting on division. Try `2**64 // 2` or `2**32`

kinto/core/resource/schema.py
```diff
@@ -6,6 +7,10 @@
                                FieldList, TimeStamp, URL)
 from kinto.core.utils import native_value
 
+POSTGRESQL_MAX_INTEGER_VALUE = 2**64 // 2
```
glasserc: Isn't this the same as `2 ** 63` ?

Natim: Yes it is .:)

Natim: Fixed with 81a86d7

Issue #1077, opened 2017-02-14
bucket_id and collection_id are not validated on the records collection url.


Natim:
```self.app.get("/buckets/beer/collections/barley'/records", headers=self.headers, status=404)```

While it should probably be
```self.app.get("/buckets/beer/collections/barley'/records", headers=self.headers, status=400)```

gabisurita:
What is considered a valid id?

leplatrem:
We should get the respective validators from `request.registry.id_generators` and makes sure it matches.

Note: this also applies to groups and collections.

Issue #707, opened 2016-07-06
Add endpoint info in OPTIONS 


leplatrem:
With django rest framework, that's what we obtain with `OPTIONS`:

``` http
➜  ~ http OPTIONS https://x.x.x.x/api/fr/reports/report
HTTP/1.1 200 OK
Allow: POST, OPTIONS
Connection: close
Content-Language: fr
Content-Type: application/json
Date: Wed, 06 Jul 2016 15:01:46 GMT
Server: nginx/1.4.6 (Ubuntu)
Transfer-Encoding: chunked
Vary: Accept, Accept-Language, Cookie

{
    "actions": {
        "POST": {
            "category": {
                "label": "Catégorie", 
                "read_only": false, 
                "required": false, 
                "type": "field"
            }, 
            "comment": {
                "label": "Commentaire", 
                "read_only": false, 
                "required": false, 
                "type": "string"
            }
        }
    }, 
    "description": "Send report", 
    "name": "Report", 
    "parses": [
        "application/json", 
        "application/x-www-form-urlencoded", 
        "multipart/form-data"
    ], 
    "renders": [
        "application/json", 
        "application/json", 
        "text/html"
    ]
}
```

On Kinto (via Cornice), we have a 400 if we don't provide CORS headers:

``` http
➜  ~ http OPTIONS https://kinto.dev.mozaws.net/v1/buckets                              
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 324
Content-Type: application/json; charset=UTF-8
Date: Wed, 06 Jul 2016 15:13:05 GMT
Server: nginx

{
    "code": 400, 
    "details": [
        {
            "description": "this header is mandatory", 
            "location": "header", 
            "name": "Origin"
        }, 
        {
            "description": "this header is mandatory", 
            "location": "header", 
            "name": "Access-Control-Request-Method"
        }
    ], 
    "errno": 107, 
    "error": "Invalid parameters", 
    "message": "Access-Control-Request-Method in header: this header is mandatory"
}

```

Providing the required headers, we only receive `null`:

``` http
➜  ~ http OPTIONS https://kinto.dev.mozaws.net/v1/buckets "Origin:http://localhost" "Access-Control-Request-Method:GET"
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Length,Expires,Backoff,Retry-After,Last-Modified,Total-Records,ETag,Pragma,Cache-Control,Alert,Next-Page
[...]

PR #1078, opened 2017-02-14
Use cornice swagger to generate the OpenAPI documentation

gabisurita:
Fixes: #1006 and #707
Replaces: #1033

- [x] Add security information.
- [x] Fix/add tests.
- [x] Add documentation.
- [x] Cleanup YAML files.

Next steps (which I don't plan to do on this PR):
- Validate ids (related to #1077)
- Add single resource documentation in options (#707)
- Use pyramid_swagger for testing
- Improve schema details/descriptions


kinto/core/resource/schema.py
```diff
@@ -331,3 +332,130 @@ class JsonPatchRequestSchema(RequestSchema):
...
+
+        return bound
+
+
+class ShareableResourseResponses(ResourceReponses):
+    """Class that wraps and handles SharableResource responses."""
+
+    def __init__(self):
+        super(ShareableResourseResponses, self).__init__()
+        self.default_schemas.update({
```
glasserc: I think this will mutate the shared `default_schemas` dictionary.

kinto/core/api.py
```diff
@@ -69,3 +106,19 @@ def default_op_ids(self, service, method):
...
+    def default_security(self, service, method):
+        """Provides OpenAPI security properties based on kinto policies."""
+
+        definitions = service.definitions
+
+        # Get method view arguments
+        for definition in definitions:
+            met, view, args = definition
+            if met == method:
+                break
```
gabisurita: This is never supposed to happen. The ``method`` that is passed to the ``default_security`` is actually fetched the same way that is done here (by inspecting service definitions), but the interface for default_xxx calls on cornice swagger is designed to take only ``service`` as a parameters (so we can have the same interface for all defaults), but we can change the api to take a method definition as an optional argument.

gabisurita:
Final r? @Natim @leplatrem

kinto/core/views/batch.py
```diff
@@ -67,13 +68,43 @@ class BatchRequest(colander.MappingSchema):
...
+
 batch = Service(name="batch", path='/batch',
                 description="Batch operations")
 
 
 @batch.post(schema=BatchRequest,
             validators=(colander_validator,),
-            permission=NO_PERMISSION_REQUIRED)
+            permission=NO_PERMISSION_REQUIRED,
+            tags=['Batch'], operation_id='batch',
```
leplatrem: nit: shouldn't these go to the service definition ?

gabisurita: Tags could be defined on the service, but cornice swagger doesn't support it yet (There's an issue for that https://github.com/Cornices/cornice.ext.swagger/issues/61). Operation ids need to be isolated by method/verb.

PR #1016, opened 2017-01-06
Update to Pyramid 1.8

birdbrained:
Pyramid moved this out of the class and into its own function.

Fixes #1008

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.
- [ ] Add your name in the contributors file.
- [ ] If you changed the HTTP API, update the [API_VERSION](https://github.com/Kinto/kinto/blob/master/kinto/__init__.py#L15) constant and add an API changelog entry [in the docs](https://github.com/Kinto/kinto/blob/master/docs/api/index.rst)

r? @

PR #1087, opened 2017-02-16
Update to Pyramid 1.8

### LINKED ISSUE #1008: Port Kinto for Pyramid 1.8
Pyramid 1.8 is breaking a number of things:

- BasicAuth authn policy
- pserve http_port config.

Right now it is still an alpha release but it gets installed on time to time.

Natim:
Fixes #1008 
Revamp #1016 

- [x] Add changelog

### DIFF
--- a/docs/core/resource.rst
+++ b/docs/core/resource.rst
@@ -198,7 +198,7 @@ As an example, let's build a code that will copy a collection into another:
     from pyramid import Configurator
 
 
-    config = Configurator(settings=DEFAULT_SETTINGS)
+    config = Configurator(settings={**DEFAULT_SETTINGS})
     config.add_settings({
         'kinto.storage_backend': 'kinto.core.storage.postgresql'
         'kinto.storage_url': 'postgres://user:pass@db.server.lan:5432/dbname'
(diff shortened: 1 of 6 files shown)


Natim:
The failing tests are because of https://github.com/Pylons/pyramid/issues/2958

Natim:
Note: The next release of Pyramid should not alter the incoming settings.

docs/core/resource.rst
```diff
@@ -198,7 +198,7 @@ As an example, let's build a code that will copy a collection into another:
     from pyramid import Configurator
 
 
-    config = Configurator(settings=DEFAULT_SETTINGS)
+    config = Configurator(settings={**DEFAULT_SETTINGS})
```
gabisurita: Is a shallow copy here enough? Maybe it's a good idea to note with XXX that this is a bug on pyramid.

Natim: Our default_settings aren't nested so yes I think it is enough.
Maybe we don't need to update the documentation about it.
The bug in Pyramid will probably be merged on Sunday but at first they said it was a feature and not a bug...

glasserc:
So wait, finally we're not doing a shallow copy or anything?

Natim:
> So wait, finally we're not doing a shallow copy or anything?

We don't have to because the last release of Pyramid 1.8 has my patch with a fix there.

Issue #1091, opened 2017-02-17
kinto.core why isn't ID in the Colander Schema?


Natim:
In our case we are using kinto.core on a resource where we want to enforce the ID.
[This](https://github.com/Kinto/kinto/blob/master/kinto/core/resource/__init__.py#L322-L323) prevent us to do that in the deserialize of the colander method.

gabisurita:
First I think it should be. Actually I was trying to make it be part of the schemas for #1077, but the main problem  for this is that the `id_generator` can be defined at any Resource subclass or at the configuration file, so we can't know at the `ViewSet` what the id pattern is. One idea I had was to bind the `id_generator` to the schemas at the `register_resource` method.

gabisurita:
Currently the id is validated by the resource https://github.com/Kinto/kinto/blob/master/kinto/core/resource/__init__.py#L784

leplatrem:
I remember that we removed `id` and `last_modified` from the schemas in the early weeks of Cliquet. 
See it was there https://github.com/mozilla-services/cliquet/blob/1.3.0/cliquet/schema.py#L31
Like an old grandpa, I don't always remember the details of the stories... but with some git bisect and grep black magic you could find the commit/PR that removed it and possibly find some explanation! It does mean that what you'll find will still be relevant ;)

Now, I see no objection for moving the id generation/validation to the schema instead of resource. That's the purpose of schemas.

gabisurita:
I've found the changes here: 
https://github.com/mozilla-services/cliquet/commit/6e23e82997d11ab2cad1481553e195f533259f36

And the motivation here:
https://github.com/mozilla-services/readinglist/issues/170

I agree with removing `last_modified` from schema as the original issue suggests, but I don't get why id was removed as well. Maybe because it was already validated by the `Resource` class?

leplatrem:
TBH I don't know. 

Even `last_modified` could be present with a `missing=drop`. Currently what happens when `last_modified` is specified with `"abc"` ?

Do not hesitate to restore them if it solves your issues ;)

gabisurita:
> Currently what happens when last_modified is specified with "abc" ?

Nothing. It's just ignored. To be honest I thought this was a feature because there are tests that check if `last_modified` is properly ignored. I think ids should be in the schema, but I don't have an strong opinion about last_modified. As the issue said, we will set it as a default timestamp anyways.

PR #1097, opened 2017-02-20
[WIP] Validate resource ids with colander schemas (fixes #1077 and #1091)

### LINKED ISSUE #1077: bucket_id and collection_id are not validated on the records collection url.

gabisurita:
Fixes #1077 and #1091

- [x] Fix broken tests.
- [x] Add schema validation tests.
- [x] Add binding tests.
- [ ] Add OpenAPI tests.
- [ ] Add a changelog entry.

Follow ups:
- Support fake requests (where `request.current_service` is not defined)
- Validate path using standard colander validation (don't fetch the model until needed)


kinto/core/resource/__init__.py
```diff
@@ -148,6 +161,14 @@ class UserResource:
     """Schema to validate records."""
 
     def __init__(self, request, context=None):
+
+        self.request = request
+        self.context = context
+
+        # XXX: Check if the path is valid before doing anything. This have to be done
+        # before fetching any model, so we can't wait for regular colander validation.
+        self._raise_400_if_invalid_path()
```
leplatrem: If that helps we could make `self.model` lazy, ie. instantiate at first use. Which should be late enough to let colander path validation happen before?

gabisurita: Yeah, that's a way to go. My first plan was to set model as a `@property`, but this lead to several recursions to solve and I just gave up for some time hahaha.

leplatrem:
Gabi, could you give us a summary about the current state of this PR so that we can go forward please? Thanks!

gabisurita:
> Gabi, could you give us a summary about the current state of this PR so that we can go forward please? Thanks!

The current status is: the validation works as is but the main problem at the moment we fetch the model before validation so we need an auxiliary method to validate the path prior to regular validation. 

I've tried some approaches to make the model lazy (e.g. make it an attribute decorated method), but it looks like it still gets fetched when we access request or context (I don't remember each). This would need some deeper investigation. I still want to fix this at some point, but fell free to finish this earlier if you need it. :)

Issue #1027, opened 2017-01-19
Adding the ETag value in the response body


Natim:
We have seen that using the ETag header can cause a number of problems with CDN and cache handling.

A proposition would be to add its value as part as the response:

```
{
  "data": [...],
  "last_modified": "1480331301530"
}

**Comment by Natim** [comment]
Alternatively we could use a meta namespace:

```
{
  "data": [...],
  "meta": {
    "last_modified": "1480331301530"
  }
}
```

**Comment by Natim** [comment]
We could probably do the same thing with all the other meta information:

```
{
    "meta": {
      "total_records": 357,
      "next_page": "https://firefox.settings.services.mozilla.com/v1/buckets/blocklists/collections/certificates/records?_limit=3&_token=eyJvZmZzZXQiOjMsImxhc3RfcmVjb3JkIjp7Imxhc3RfbW9kaWZpZWQiOjE0ODQ3MDQ1ODA4OTR9fQ%3D%3D",
      "last_modified": 1484704581273,
      "date": "1484838228331"
    }
}
```

leplatrem:
Agree. 

This is what JSONApi recommands too: http://jsonapi.org/format/#document-meta

Maybe, the `next_page` could go to `"links": {"next": "https://"}` so that we stick to their spec too: http://jsonapi.org/format/#document-links

glasserc:
For those of you following along at home, one specific problem we've seen is that the CDN will gzip a response and then strip the ETag. Since we're using ETags for concurrency control, this breaks syncing and everything else.

There's an argument that we're using ETag wrong, since the RFC says an ETag corresponds to a *representation* rather than the resource itself. That would make the CDN behavior correct, since gzip'ing a response means producing a new representation. There's also the `Last-Modified` / `If-Unmodified-Since` headers. We could also perform a test with "weak" ETags, which only change if the resource is no longer the same "semantically". Maybe the CDN would preserve a "weak" ETag, even if it gzips the response. On the other hand, this might break all our existing clients, which assume ETags are just string-encoded integers :(

@Natim proposed in #1102 getting rid of ETags completely. I'm not crazy about the idea of removing ETag, since it will break our old clients. I'd be in favor of adding new concurrency-control information, whether that's within a `meta` tag or as `Last-Modified` or something else.

leplatrem:
Just for the reference: we used to rely on `If-Modified-Since` but got rid of them because they are not opaque, in the sense that they should contain an HTTP date, with a precision *to second*.

[...]

PR #1102, opened 2017-02-22
Clarify If-Match behavior (fixes #1072)

### LINKED ISSUE #1072: If-Match: * does not match the spec
Per https://github.com/Kinto/kinto/pull/1064#issuecomment-279351031, the behavior of `If-Match: *` is the opposite of `If-None-Match: *`. `If-None-Match: *` should succeed if no resource exists at the given URL and fail if any resource exists at the given URL, but `If-Match: *` should fail if no resource exists at the given URL and succeed if any resource exists at the given URL. Currently `If-Match: *` does nothing (it always succeeds).

gabisurita:
Fixes #1072

This tries to clarify some inaccuracies on the `If-Match` and  `If-None-Match` custom headers behavior and make them a little more similar to what the spec expects.

Proposed changes:
1. If-Match raise on with non-existing records (including Tombstones)
2. If-Match:* succeeds if the record exist.
3. If-Match and If-None-Match a assume a collection is a resource that always exists.

Some implications:
1. 1 and 3 implies that `If-Match: *` on a collection will always succeed.
2. 3 and the current behavior for `If-None-Match: *` implies this on collection will always fail (the current implementation just ignores it)

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/docs/api/1.x/timestamps.rst
+++ b/docs/api/1.x/timestamps.rst
@@ -81,27 +81,42 @@ the ``If-None-Match: *`` request header.
 
 The following table gives a summary of the expected behaviour of a resource:
 
-+-----------------------------+-------------+--------------+---------------+---------------+
-|                             | POST        | PUT          | PATCH         | DELETE        |
-+=============================+=============+==============+===============+===============+
-|| **If-Match: "timestamp"**                                                               |
-+-----------------------------+-------------+--------------+---------------+---------------+
-| Changed meanwhile           | ``HTTP 412``| ``HTTP 412`` | ``HTTP 412``  | ``HTTP 412``  |
-+-----------------------------+-------------+--------------+---------------+---------------+
-| Not changed                 | Create      | Overwrite    | Modify        | Delete        |
-+-----------------------------+-------------+--------------+---------------+---------------+
-|| **If-None-Match: ***                                                                    |
-+-----------------------------+-------------+--------------+---------------+---------------+
-| Id exists                   | ``HTTP 412``| ``HTTP 412`` | No effect     | No effect     |
-+-----------------------------+-------------+--------------+---------------+---------------+
-| Id unknown                  | Create      | Create       | No effect     | No effect     |
-+-----------------------------+-------------+--------------+---------------+---------------+
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+|                             | GET         | POST        | PUT         | PATCH       | DELETE      |
++=============================+=============+=============+=============+=============+=============+
+|| **If-Match: "timestamp"**                                                                        |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Not changed                 | Fetch       | Fetch       | Overwrite   | Modify      | Delete      |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Changed meanwhile           | ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``|
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+|| **If-Match: ***                                                                                  |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Id exists                   | Fetch       | Fetch       | Overwrite   | Modify      | Delete      |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Id unknown                  | ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``|
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+|| **If-None-Match: ***                                                                             |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Id exists                   | ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``|
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Id unknown                  | No effect   | Create      | Create      | No effect   | No effect   |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+
 
 When the client receives a |status-412|, it can then choose to:
 
 * overwrite by repeating the request without concurrency control;
 * reconcile the resource by fetching, merging and repeating the request.
 
+
+.. note::
+
+    ``If-Match`` and ``If-None-Match`` headers on Kinto work slightly different
+    from what `the specification expects <https://tools.ietf.org/html/rfc7232#section-3>`_
+    for these precondition headers. Instead of comparing if ``ETag`` is exactly the
+    same, as the specification suggests, it just checks if the ``ETag`` is newer
+    or older in order to simplify the synchronization behavior.
 
 Replication
 ===========
--- a/docs/api/1.x/timestamps.rst
+++ b/docs/api/1.x/timestamps.rst
@@ -110,15 +110,6 @@ When the client receives a |status-412|, it can then choose to:
 * reconcile the resource by fetching, merging and repeating the request.
 
 
-.. note::
-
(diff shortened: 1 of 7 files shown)


docs/api/1.x/timestamps.rst
```diff
@@ -81,28 +81,44 @@ the ``If-None-Match: *`` request header.
...
 * reconcile the resource by fetching, merging and repeating the request.
 
 
+.. note::
+
+    ``If-Match`` and ``If-None-Match`` headers on Kinto work slightly different from what
+    `their RFC specification expects <https://tools.ietf.org/html/rfc7232#section-3>`_.
+    Instead of comparing if the current
+    ``ETag`` is exactly the same as the value provided on the header,
+    the current implementation just checks if the ``ETag`` is newer or older
```
leplatrem: I think there is something wrong with this sentence, since we contrast *exactly the same* with *newer or older*, and that sounds weird. In other words, how *«newer or older»* is not equivalent to *«not exactly the same»*?

By the way, personnally, I would change the behaviour to match the spec. I don't see any implication in the synchronization process. The clients are not supposed to have timestamps that are superior to the server ones, so I wouldn't even consider this a breaking change.

kinto/core/resource/__init__.py
```diff
@@ -260,7 +260,7 @@ def collection_get(self):
         self._add_timestamp_header(self.request.response)
         self._add_cache_header(self.request.response)
         self._raise_304_if_not_modified()
-        self._raise_412_if_modified()
+        self._raise_412_if_modified(record={})  # ignore If-Match on non-existing
```
leplatrem: I would adjust the comment, your explanation on the PR description was a lot better: for example, something like *collection is a resource that always exists"

kinto/core/resource/__init__.py
```diff
@@ -452,8 +452,7 @@ def put(self):
             if len(tombstones) > 0:
                 existing = tombstones[0]
         finally:
-            if existing:
-                self._raise_412_if_modified(existing)
+            self._raise_412_if_modified(existing)
```
leplatrem: nit: we could specify the keyword `record=`

kinto/core/resource/__init__.py
```diff
@@ -829,18 +828,27 @@ def _raise_412_if_modified(self, record=None):
         if_match = self.request.validated['header'].get('If-Match')
         if_none_match = self.request.validated['header'].get('If-None-Match')
 
+        # Check if record exists and it's not a Tombstone
```
leplatrem: nit: useless upercase on tombstone

kinto/core/resource/__init__.py
```diff
@@ -829,18 +828,27 @@ def _raise_412_if_modified(self, record=None):
...
+        # If no precondition headers, just ignore
         if not if_match and not if_none_match:
             return
 
-        if record and if_none_match == '*':
-            if record.get(self.model.deleted_field, False):
-                # Tombstones should not prevent creation.
-                return
+        # If-None-Match: * should always raise if a record exist
+        elif if_none_match == '*' and record_exists:
```
leplatrem: nit: `elif` is not necessary since we `return` above

kinto/core/resource/__init__.py
```diff
@@ -829,18 +828,27 @@ def _raise_412_if_modified(self, record=None):
...
+
+        # If no precondition headers, just ignore
         if not if_match and not if_none_match:
             return
 
-        if record and if_none_match == '*':
-            if record.get(self.model.deleted_field, False):
-                # Tombstones should not prevent creation.
-                return
+        # If-None-Match: * should always raise if a record exist
```
leplatrem: nit: exists

kinto/core/resource/__init__.py
```diff
@@ -829,18 +828,27 @@ def _raise_412_if_modified(self, record=None):
...
+        elif if_none_match == '*' and record_exists:
             modified_since = -1  # Always raise.
+
+        # If-Match should always raise if a record don't exist
```
leplatrem: nit: doesn't

tests/core/resource/test_preconditions.py
```diff
@@ -200,13 +200,55 @@ def test_post_if_none_match_star_succeeds_if_record_does_not_exist(self):
                 'field': 'new'}}
         self.resource.collection_post()  # not raising.
 
-    def test_post_if_match_star_succeeds_if_record_does_exist(self):
```
leplatrem: should we have the tests on collection get/delete that always succeed with `If-Match:*` or `If-None-Match: *` ?

gabisurita:
> I am favor in favor or fixing the ETag comparison too. Let's gather opinions...

So let's do it them. If we want to match the spec strong comparison requirements and also keep the current behavior, we could allow specifying week entities as the spec suggests, but in case we don't need it lets just keep it simple.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.3.3

Natim:
If we want to fix the behaviour we should get rid of ETag's all together I guess because we are not using Etag as OPS understand they should work.
Refs #1027

leplatrem:
> If we want to fix the behaviour we should get rid of ETag's all together I guess because we are not using Etag as OPS understand they should work.

Huh, what?

docs/api/1.x/timestamps.rst
```diff
@@ -81,21 +81,28 @@ the ``If-None-Match: *`` request header.
...
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Id unknown                  | ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``|
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+|| **If-None-Match: ***                                                                             |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Id exists                   | ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``| ``HTTP 412``|
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+| Id unknown                  | No effect   | Create      | Create      | No effect   | No effect   |
++-----------------------------+-------------+-------------+-------------+-------------+-------------+
+
```
leplatrem: We could mention those changes in the API changelog (and increment the HTTP API number maybe)

Natim:
> Huh, what?

Well forget it we will handle that with #1027

docs/api/index.rst
```diff
@@ -11,7 +11,14 @@ API
 Changelog
 ---------
 
-1.14 (Unreleased)
+1.15 (Unreleased)
+'''''''''''''''''
+- ``If-Match`` and ``If-None-Match`` precondition headers now check if ``ETag`` is
+  equal or not equal to the provided value (as in RFC 2616). Previous versions
+  asserted then to be greater or lower.
```
glasserc: I agree with @leplatrem's previous comment about this verbiage. How about `If-Match and If-None-Match precondition headers now check the ETag for strict equality. Previous versions would allow requests if they seemed to be more recent than the current version.`

tests/core/resource/test_preconditions.py
```diff
@@ -200,30 +204,91 @@ def test_post_if_none_match_star_succeeds_if_record_does_not_exist(self):
...
+
+    def test_put_if_match_star_fails_if_record_does_not_exist(self):
+        self.validated['header']['If-Match'] = '*'
+        self.resource.request.json = {'data': {'field': 'new'}}
+        self.validated['body'] = self.resource.request.json
+        self.resource.record_id = self.resource.model.id_generator()
+        self.assertRaises(httpexceptions.HTTPPreconditionFailed,
+                          self.resource.put)
+
+    def test_put_if_fails_if_record_does_not_exist(self):
```
glasserc: Should this be `put_if_match` ? (Or was that too long?)

glasserc:
I'm also in favor of fixing the ETag thing, which I'm glad you did.

PR #1105, opened 2017-02-23
Get the previous version of a record in the past.

### LINKED ISSUE #1104: [history] State of objects at particular timestamp
There is a need to be able to see the state of an object (or list of objects) at a particular timestamp (when the history plugin is enabled).

This would consist in applying the history entries backwards to obtain the same response (+headers) as it used to be at that timestamp. The endpoint would be read-only.

There are some specific aspects that may be difficult to handle like permissions (is that user allowed to see those old records?).

An alternative approach is to do it on the client side using the existing history endpoint.

See also #826

Natim:
Fixes #1104 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.
- [ ] Handle permissions
- [ ] Refactor test setup step to avoid duplicating code.
- [ ] Split tests in short well-scoped methods.

f? @leplatrem


CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ Protocol is now at version **1.15**. See `API changelog`_.
 - Remove Python 2.7 support and upgrade to Python 3.5. (#1050)
 - Upgraded minimal PostgreSQL support to PostgreSQL 9.5 (#1056)
 
+**New features**
+
+- Add a `_at` parameter to a collection to grab it's history. (#1104)
```
Natim: We change the API so we should change the changelog.

docs/api/1.x/history.rst
```diff
@@ -196,3 +196,32 @@ It is possible to exclude certain resources from being tracked by history using
...
+-  ``GET /buckets/(bucket_id)/collections/(collection_id)/records/version/(timestamp)``
+-  ``GET /buckets/(bucket_id)/collections/(collection_id)/records/(record_id)/version/(timestamp)``
+
+
+.. note::
+
+    The history plugin does not record bucket deletion because the
+    whole bucket history is deleted on bucket deletion. That's why we
+    cannot provide a version history on the buckets collection
+    resource because we are not tracking it.
```
leplatrem: nit: buckets collection → buckets plural endpoint

docs/api/1.x/history.rst
```diff
@@ -196,3 +196,32 @@ It is possible to exclude certain resources from being tracked by history using
 
     kinto.history.exclude_resources = /buckets/preview
                                       /buckets/signed/collections/certificates
+
+
+Time traveling
+==============
+
+It is possible to retrieve a previous state of a collection of records
+or a record (a snapshot) by using the ``/version/(timestamp)``
```
leplatrem: a previous state of [...] → of *objects*

docs/api/1.x/history.rst
```diff
@@ -196,3 +196,32 @@ It is possible to exclude certain resources from being tracked by history using
 
     kinto.history.exclude_resources = /buckets/preview
                                       /buckets/signed/collections/certificates
+
+
+Time traveling
+==============
+
+It is possible to retrieve a previous state of a collection of records
```
leplatrem: «state» → «version» ?

docs/api/1.x/history.rst
```diff
@@ -196,3 +196,32 @@ It is possible to exclude certain resources from being tracked by history using
...
+.. note::
+
+    The history plugin does not record bucket deletion because the
+    whole bucket history is deleted on bucket deletion. That's why we
+    cannot provide a version history on the buckets collection
+    resource because we are not tracking it.
+
+.. note::
+
+   Currently it is not possible to filter snapshot list resource.
```
leplatrem: snapshot list resource → what is this? we were talking about state then version...  we should stick to a reduced vocabulary to avoid confusion.

docs/api/index.rst
```diff
@@ -17,6 +17,8 @@ Changelog
   strict equality. Previous versions would allow requests if they seemed
   to be more recent than the current version.
 - ``If-Match`` now raises ``412`` if a record doesn't exist.
+- Add a version endpoint to resource URL to retrieve previous state of
+  records and collections from the past.
```
leplatrem: µnit: collections and records ;)

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,85 @@ def _extract_filters(self):
...
 
         return filters_str_id
+
+
+version_view = Service(name='version_view',
+                       description='Handle retrieving object from the past.',
+                       path='{subpath:.*}/version/{last_modified:[0-9]{13}}')
+
+
+@version_view.get(permission=NO_PERMISSION_REQUIRED)
```
leplatrem: Does this make every snapshot public?

The permission aspect is crucial.
Example scenario: I have access to a collection, then I am removed the perm, then some data is created. Can I access a snapshot of some data that was edited after I was removed the permission?

It's not impossible to handle, but not trivial.

The history endpoint already handles this, by providing history entries only on data that is readable/writable. That's why I suggested to do this reconciliation on the client side.

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,85 @@ def _extract_filters(self):
...
+def get_version_view(request):
+    last_modified = int(request.matchdict['last_modified'])
+    subpath = '/{}'.format(request.matchdict['subpath'])
+
+    if not subpath.startswith('/buckets'):
+        raise httpexceptions.HTTPNotFound()
+
+    bucket_id = request.matchdict['subpath'].split('/', 2)[1]
```
leplatrem: Why all of this instead of putting `/buckets/{bucket_id}/{subpath:.*}` in the path?

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,85 @@ def _extract_filters(self):
...
+    bucket_uri = '/buckets/{}'.format(bucket_id)
+
+    is_collection = [True for collection_type in ['collections', 'groups', 'records']
+                     if subpath.endswith('/{}'.format(collection_type))]
```
leplatrem: This is not idiomatic, use `any()`

tests/plugins/test_history_at.py
```diff
@@ -0,0 +1,182 @@
...
+        self.record_uri = '/buckets/test/collections/col/records/rec'
+        body = {'data': {'foo': 42}}
+        resp = self.app.put_json(self.record_uri, body, headers=self.headers)
+        self.record = resp.json['data']
+
+    def build_version_object_uri(self, object_url):
+        return object_url + '/version'
+
+    def test_version_doesnt_crash_on_wrong_integer(self):
+        self.app.get('/buckets/version/toto', status=404)
```
leplatrem: You said above that «*we cannot provide a version history on the buckets collection*»

tests/plugins/test_history_at.py
```diff
@@ -0,0 +1,182 @@
+from .test_history import HistoryWebTest
```
leplatrem: The filename `test_history_at` does not match the vocabulary elsewhere (version, state, snapshot, ...)

tests/plugins/test_history_at_collections.py
```diff
@@ -0,0 +1,355 @@
...
+        resp = self.app.get('{}/{}'.format(bucket_records_uri, record3_creation_last_modified),
+                            headers=self.headers)
+        records = resp.json['data']
+        assert records == [record3, record2, self.record]
+
+        resp = self.app.get('{}/{}'.format(bucket_records_uri,
+                                           record_update_last_modified),
+                            headers=self.headers)
+        records = resp.json['data']
+        assert records == [record, record3, record2]
```
leplatrem: Those tests look like huge copy-paste of the same steps. Plus the test methods are long and contains several steps. It's usually easier to have short well-scoped test methods

docs/api/1.x/history.rst
leplatrem: If we do it on the client side, the filtering aspect is out of scope.

tests/plugins/test_history_at.py
Natim: buckets collection which is `/buckets/version/1234567890123` we can on the bucket record which is `/buckets/test/version/1234567890123`

tests/plugins/test_history_at.py
Natim: This file is the one from the history plugin which this feature is a part of.

kinto/plugins/history/views.py
Natim: > The history endpoint already handles this, by providing history entries only on data that was readable/writable at the time it was edited.

If this is the case it is really great ! We can have the exact same behaviour here.

docs/api/1.x/history.rst
Natim: > we should stick to a reduced vocabulary to avoid confusion.

I agree but kinto collection and collections of records is already really confusing...

Natim: > If we do it on the client side, the filtering aspect is out of scope.

Yes the filtering can appen on the client side.

Natim:
> Without the permission handling, I would not be in favor of merging.

Yes I agree we weren't sure about how the permission where handled on the history endpoint.

> I'm sorry that you already wrote so much code :|

Why are you sorry? Isn't it great? It is just a matter of handling permission right?

Natim:
Test are broken because of some swagger specification issue. @gabisurita can you help me out there?

gabisurita:
> Test are broken because of some swagger specification issue. @gabisurita can you help me out there?

Yeah. Cornice swagger doesn't support the `*subpath` notation (I didn't know cornice supported it as well). I'll try to find a fix for this or we can just leave it out the documentation for now.

Natim:
> or we can just leave it out the documentation for now.

How can I do that?

leplatrem:
>> I'm sorry that you already wrote so much code :|
>
> Why are you sorry? Isn't it great? It is just a matter of handling permissions right?

Yeah «*just*» :) 

Also, does it provide ETag etc as if it would be the usual endpoint?

Natim:
>  does it provide ETag etc as if it would be the usual endpoint?

Actually it is not a usual endpoint and it doesn't plan to be one. What would be the use case?

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,83 @@ def _extract_filters(self):
...
+def get_version_view(request):
+    last_modified = int(request.matchdict['last_modified'])
+    bucket_id = request.matchdict['bucket_id']
+    resource_path = '/buckets/{}{}'.format(request.matchdict['bucket_id'],
+                                           request.matchdict['subpath'])
+
+    bucket_uri = '/buckets/{}'.format(bucket_id)
+
+    is_collection = any([resource_path.endswith('/{}'.format(collection_type))
+                         for collection_type in ['collections', 'groups', 'records']])
```
leplatrem: nit: `.endswith('/{}s')`

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,83 @@ def _extract_filters(self):
...
+    return handle_version_on_records(request, last_modified, bucket_uri)
+
+
+def get_resource_name(request):
+    subpath = '/{}'.format(request.matchdict['subpath'])
+    resources = {
+        '/collections': "collection",
+        '/groups': 'group',
+        '/records': 'record'
+    }
```
leplatrem: why not `endswith('{}s')` like above?

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,83 @@ def _extract_filters(self):
...
+        if record_id not in already_processed:
+            already_processed.add(record_id)
+            if record['target']['data'].get('deleted') is not True:
+                results.append(record['target']['data'])
+
+    return {"data": results}
+
+
+def handle_version_on_records(request, last_modified, bucket_uri):
+    parent_id = '{}{}'.format(bucket_uri, request.matchdict['subpath'])
```
leplatrem: nit: why not `+` ?

kinto/plugins/history/views.py
leplatrem: Is adding the `/` necessary?

tests/plugins/test_history_version_records.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def build_version_object_uri(self, object_url):
+        return object_url + '/version'
+
+    def test_version_doesnt_crash_on_wrong_integer(self):
+        self.app.get('/buckets/version/toto', status=404)
+
+    def test_version_doesnt_crash_when_bucket_is_missing_from_the_uri(self):
+        self.app.get('/undefined/version/1487870672412', status=404)
+
```
leplatrem: Maybe add a test where you have an unknown resource name? `/buckets/abc/labels/12/version/1234` ?

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,83 @@ def _extract_filters(self):
...
+
+    # Only process records once
+    already_processed = set()
+    results = []
+    for record in records:
+        record_id = record['target']['data']['id']
+        if record_id not in already_processed:
+            already_processed.add(record_id)
+            if record['target']['data'].get('deleted') is not True:
+                results.append(record['target']['data'])
```
leplatrem: Why not use an ordered dict to deduplicate instead of set + list?

kinto/plugins/history/views.py
```diff
@@ -57,3 +61,83 @@ def _extract_filters(self):
...
+        filters=filters, sorting=sorting)
+
+    # Only process records once
+    already_processed = set()
+    results = []
+    for record in records:
+        record_id = record['target']['data']['id']
+        if record_id not in already_processed:
+            already_processed.add(record_id)
+            if record['target']['data'].get('deleted') is not True:
```
leplatrem: nit: why `is not True` and not just `not` ?

tests/plugins/test_history_version_records.py
```diff
@@ -0,0 +1,182 @@
...
+        self.record_uri = '/buckets/test/collections/col/records/rec'
+        body = {'data': {'foo': 42}}
+        resp = self.app.put_json(self.record_uri, body, headers=self.headers)
+        self.record = resp.json['data']
+
+    def build_version_object_uri(self, object_url):
+        return object_url + '/version'
+
+    def test_version_doesnt_crash_on_wrong_integer(self):
+        self.app.get('/buckets/version/toto', status=404)
```
gabisurita: Shouldn't it be 400 here?

tests/plugins/test_history_at.py
```diff
@@ -0,0 +1,80 @@
...
+        self.record_uri = '/buckets/test/collections/col/records/rec'
+        body = {'data': {'foo': 42}}
+        resp = self.app.put_json(self.record_uri, body, headers=self.headers)
+        self.record = resp.json['data']
+
+    def build_version_object_uri(self, object_url):
+        return object_url + '/version'
+
+    def test_version_doesnt_crash_on_wrong_integer(self):
+        self.app.get('/buckets/version/toto', status=404)
```
Natim: A missing resource return a `404`, if it was an invalid parameter we would return a `400` but here the resource just doesn't exists because no route are matching it.

kinto/plugins/history/views.py
Natim: Yes because you don't want to match `/buckets/twittercollections` for instance

kinto/plugins/history/views.py
Natim: I can see some use cases where deleted is a property of a record and not a tombstones: `"deleted": 3`

kinto/plugins/history/views.py
Natim: Sure how do you do that?

kinto/plugins/history/views.py
Natim: Because generally you don't want to concat strings with + 
In that specific case we probably could.
https://waymoot.org/home/python_string/

leplatrem: > Because generally you don't want to concat strings with +

Why ?

kinto/plugins/history/views.py
leplatrem: Using `dict.values()` no?

kinto/plugins/history/views.py
leplatrem: Interesting, ok

kinto/plugins/history/views.py
leplatrem: Good point! Please add some comments to the code, the conversations on github don't last

leplatrem:
>>  does it provide ETag etc as if it would be the usual endpoint?
> Actually it is not a usual endpoint and it doesn't plan to be one. What would be the use case?

'cuz @n1k0 mentioned use-cases like initial synchronization or replication, but appart from building a diff from a list of changes I don't have any use-case for this endpoint anyway :D

kinto/plugins/history/views.py
Natim: http://softwareengineering.stackexchange.com/a/304447

Natim:
I don't think you need ETag for replication or initial synchronization. Using the last_modified should be enough to get the state before a certain timestamp and then build on top of it or doing three way merge.

n1k0:
> 'cuz @n1k0 mentioned use-cases like initial synchronization or replication, but appart from building a diff from a list of changes I don't have any use-case for this endpoint anyway :D

OK fair enough, no valid identified use cases apart from what I should be working on. So we should probably give a shot at a client side implementation of the feature in kinto-http.js. You mentioned it was trivial, while I feel some complexity here, so my question is, do you want to pair on this? 😊

leplatrem:
> So we should probably give a shot at a client side implementation of the feature in kinto-http.js.

Yes, at least give it a try 🎉 

> You mentioned it was trivial, while I feel some complexity here, so my question is, do you want to pair on this? 😊

If I'm not wrong, it's just a matter of writing this part in JS, which replays the history from the start:

```py
    # We want to get the record at a certain time.
    filters = [Filter('target.data.last_modified', last_modified, COMPARISON.MAX),
               Filter('resource_name', resource_name, COMPARISON.EQ),
               Filter('uri', parent_id, COMPARISON.LIKE)]
    sorting = [Sort('last_modified', -1)]

    records, count = request.registry.storage.get_all(
        collection_id='history', parent_id=bucket_uri,
        filters=filters, sorting=sorting)

    # Only process records once
    already_processed = set()
    results = []
    for record in records:
        record_id = record['target']['data']['id']
        if record_id not in already_processed:
            already_processed.add(record_id)
            if record['target']['data'].get('deleted') is not True:
                results.append(record['target']['data'])

    return {"data": results}
```

Which would give something like that:

```js
entries = yield client.listHistory({
  'max_target.data.last_modified': timestamp,
  resource_name: record,
  uri: collection_uri
})

results = entries.reduce((entry, acc) => {
  const {target: {data}} = entry;
  const {id, deleted} = data;
  if (id in acc || deleted) {
    return acc;
  }
  acc[id] = data;
  return acc;
}, {})
.sort(({last_modified: a}, {last_modified: b}) => a - b);

```

There is another approach, which consist in applying the history backwards. The code would be a bit more complex, but it would be cheaper in resources (especially on collections with frequent changes). 

Note: In our use case, we don't have to deal with permissions since the obtained history entries already take them into account. 

We can pair on this on Monday, but I always feel bad when lots of code/efforts are at stake :|

n1k0:
> Yes, at least give it a try :tada:

https://github.com/Kinto/kinto-http.js/pull/161

Natim:
Handle on client side on the kinto-admin. Feel free to reopen if this is a feature that you want to push forward on the server side.

PR #1119, opened 2017-03-01
Prevent duplicate_key error on INSERT

### LINKED ISSUE #602: [cliquet] Timestamp uniqueness by parent_id under heavy load

Natim:
Fixes  #602

- [ ] Add tests.
- [ ] Add a changelog entry.

f? @leplatrem

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -245,7 +245,7 @@ class Storage(StorageBase):
                 from_epoch(:last_modified))
         ON CONFLICT (id, parent_id, collection_id) DO UPDATE
         SET data = (:data)::JSONB,
-            last_modified = from_epoch(:last_modified)
+            last_modified = find_timestamp(:parent_id, :collection_id, from_epoch(:last_modified))
         RETURNING id, as_epoch(last_modified) AS last_modified;
         """
         placeholders = dict(object_id=record[id_field],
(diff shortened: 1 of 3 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -243,6 +243,9 @@ def create(self, collection_id, parent_id, record, id_generator=None,
         VALUES (:object_id, :parent_id,
                 :collection_id, (:data)::JSONB,
                 from_epoch(:last_modified))
+        ON CONFLICT (id, parent_id, collection_id) DO UPDATE
```
leplatrem: we don't mention `last_modified` here?

kinto/core/storage/postgresql/__init__.py
```diff
@@ -243,6 +243,9 @@ def create(self, collection_id, parent_id, record, id_generator=None,
...
+        SET data = (:data)::JSONB,
+            last_modified = from_epoch(:last_modified)
```
leplatrem: I'm not sure, but if `last_modified` is not provided and there is a conflict, then this could fail (`last_modified` can not be null, but maybe the trigger will initialize it, we have to check)

kinto/core/storage/postgresql/__init__.py
```diff
@@ -304,15 +307,9 @@ def update(self, collection_id, parent_id, object_id, record,
...
-
-        query_update = """
-        UPDATE records SET data=(:data)::JSONB,
-                           last_modified=from_epoch(:last_modified)
-        WHERE id = :object_id
-           AND parent_id = :parent_id
-           AND collection_id = :collection_id
+        ON CONFLICT (id, parent_id, collection_id) DO UPDATE
+        SET data = (:data)::JSONB,
+            last_modified = from_epoch(:last_modified)
```
leplatrem: If tests pass, this is cool :)

Natim: We need somehow to bump_timestamp one more time in case of conflict.

Natim:
Superseeded by #1120

PR #1120, opened 2017-03-01
Prevent IntegrityError on INSERT.

### LINKED ISSUE #602: [cliquet] Timestamp uniqueness by parent_id under heavy load

Natim:
Fixes #602 
Superseed #1120

- [ ] Add tests.
- [ ] Add a changelog entry.

paired with @leplatrem

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -65,7 +65,7 @@ class Storage(StorageBase):
 
     """  # NOQA
 
-    schema_version = 14
+    schema_version = 15
 
     def __init__(self, client, max_fetch_size, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -243,6 +243,9 @@ class Storage(StorageBase):
         VALUES (:object_id, :parent_id,
                 :collection_id, (:data)::JSONB,
                 from_epoch(:last_modified))
+        ON CONFLICT (id, parent_id, collection_id) DO UPDATE
+        SET data = (:data)::JSONB,
+            last_modified = find_timestamp(:parent_id, :collection_id, from_epoch(:last_modified))
         RETURNING id, as_epoch(last_modified) AS last_modified;
         """
         placeholders = dict(object_id=record[id_field],
@@ -304,6 +307,9 @@ class Storage(StorageBase):
         VALUES (:object_id, :parent_id,
                 :collection_id, (:data)::JSONB,
                 from_epoch(:last_modified))
+        ON CONFLICT (id, parent_id, collection_id) DO UPDATE
+        SET data = (:data)::JSONB,
+            last_modified = find_timestamp(:parent_id, :collection_id, from_epoch(:last_modified))
         RETURNING as_epoch(last_modified) AS last_modified;
         """
 
--- a/kinto/core/storage/postgresql/schema.sql
+++ b/kinto/core/storage/postgresql/schema.sql
@@ -95,6 +95,34 @@ END;
 $$ LANGUAGE plpgsql;
 
 --
+-- Find timestamp in case of conflict
+--
+
+CREATE OR REPLACE FUNCTION find_timestamp(pid VARCHAR, cid VARCHAR, ts TIMESTAMP)
+RETURNS TIMESTAMP AS $$
+DECLARE
+    previous_collection_ts TIMESTAMP;
+    record_ts TIMESTAMP;
+BEGIN
+	record_ts := ts;
+
+	previous_collection_ts := NULL;
+    SELECT last_modified INTO previous_collection_ts
+      FROM timestamps
+     WHERE parent_id = pid
+       AND collection_id = cid;
+
+    IF ts IS NULL OR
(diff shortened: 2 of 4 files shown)


kinto/core/storage/postgresql/schema.sql
```diff
@@ -122,33 +149,33 @@ BEGIN
...
     WITH upsert AS (
-        UPDATE timestamps SET last_modified = current
+        UPDATE timestamps SET last_modified = next_collection_ts
          WHERE parent_id = NEW.parent_id AND collection_id = NEW.collection_id
         RETURNING *
     )
     INSERT INTO timestamps (parent_id, collection_id, last_modified)
-    SELECT NEW.parent_id, NEW.collection_id, current
+    SELECT NEW.parent_id, NEW.collection_id, next_collection_ts
     WHERE NOT EXISTS (SELECT * FROM upsert);
```
leplatrem: nit: This could be rewritten using `ON CONFLICT`

kinto/core/storage/postgresql/schema.sql
```diff
@@ -95,6 +95,34 @@ END;
...
+
+CREATE OR REPLACE FUNCTION find_timestamp(pid VARCHAR, cid VARCHAR, ts TIMESTAMP)
+RETURNS TIMESTAMP AS $$
+DECLARE
+    previous_collection_ts TIMESTAMP;
+    record_ts TIMESTAMP;
+BEGIN
+	record_ts := ts;
+
+	previous_collection_ts := NULL;
```
leplatrem: µnit: indentation

kinto/core/storage/postgresql/schema.sql
```diff
@@ -95,6 +95,34 @@ END;
...
+     WHERE parent_id = pid
+       AND collection_id = cid;
+
+    IF ts IS NULL OR
+       (previous_collection_ts IS NOT NULL AND as_epoch(ts) = as_epoch(previous_collection_ts)) THEN
+        -- If record does not carry last-modified, or if the one specified
+        -- is equal to previous, assign it to current (i.e. bump it).
+	  record_ts := previous_collection_ts;
+	END IF;
+RETURN record_ts;
```
leplatrem: µnit: identation

kinto/core/storage/postgresql/schema.sql
```diff
@@ -95,6 +95,34 @@ END;
...
+	previous_collection_ts := NULL;
+    SELECT last_modified INTO previous_collection_ts
+      FROM timestamps
+     WHERE parent_id = pid
+       AND collection_id = cid;
+
+    IF ts IS NULL OR
+       (previous_collection_ts IS NOT NULL AND as_epoch(ts) = as_epoch(previous_collection_ts)) THEN
+        -- If record does not carry last-modified, or if the one specified
+        -- is equal to previous, assign it to current (i.e. bump it).
```
leplatrem: This comment is not really in-sync anymore (eg. we don't bump here)

kinto/core/storage/postgresql/schema.sql
```diff
@@ -95,6 +95,34 @@ END;
...
+	record_ts := ts;
+
+	previous_collection_ts := NULL;
+    SELECT last_modified INTO previous_collection_ts
+      FROM timestamps
+     WHERE parent_id = pid
+       AND collection_id = cid;
+
+    IF ts IS NULL OR
+       (previous_collection_ts IS NOT NULL AND as_epoch(ts) = as_epoch(previous_collection_ts)) THEN
```
leplatrem: If I'm not wrong, `previous_collection_ts` will never be NULL because it would have been set during the first operation before the conflict occured.


Also, if I understand correctly, this will re-run the trigger with `record.last_modified = collection_ts`, which will re-bump the timestamp (as if a user would have had specified it in a normal PUT). Hence, shouldn't we compare with `as_epoch(ts) <= as_epoch(previous_collection_ts))`

kinto/core/storage/postgresql/__init__.py
```diff
@@ -243,6 +243,9 @@ def create(self, collection_id, parent_id, record, id_generator=None,
         VALUES (:object_id, :parent_id,
                 :collection_id, (:data)::JSONB,
                 from_epoch(:last_modified))
+        ON CONFLICT (id, parent_id, collection_id) DO UPDATE
```
leplatrem: Here we are resolving only one kind of conflict: two users trying to create objects with the same id in overlapping transactions.

What we face in #602 is different IMO: we create objects with different ids, but when we fail in assigning unique timestamps. And I don't think this `ON CONFLICT` will catch it.

Natim: Ok it makes sense. Let's find out how we can create a test to reproduce that. I guess we had one with loadtest simulation before.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -243,6 +243,9 @@ def create(self, collection_id, parent_id, record, id_generator=None,
...
+        SET data = (:data)::JSONB,
+            last_modified = find_timestamp(:parent_id, :collection_id, from_epoch(:last_modified))
```
leplatrem: I just read that `EXCLUDED.last_modified` would give us access to the rejected inserted value

Natim:
Handled by #1122 and #1125

PR #1122, opened 2017-03-02
Add an ignore_conflict argument to create for the default_bucket use case

Natim:
Refs #602 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/kinto/core/storage/__init__.py
+++ b/kinto/core/storage/__init__.py
@@ -86,6 +86,7 @@ class StorageBase:
         :param str collection_id: the collection id.
         :param str parent_id: the collection parent.
         :param dict record: the object to create.
+        :param boolean ignore_conflict: Do not raise the UnicityError on conflict.
 
         :returns: the newly created object.
         :rtype: dict
--- a/kinto/core/storage/__init__.py
+++ b/kinto/core/storage/__init__.py
@@ -72,7 +72,7 @@ class StorageBase:
     def create(self, collection_id, parent_id, record, id_generator=None,
                id_field=DEFAULT_ID_FIELD,
                modified_field=DEFAULT_MODIFIED_FIELD,
-               auth=None):
+               auth=None, ignore_conflict=False):
         """Create the specified `object` in this `collection_id` for this `parent_id`.
         Assign the id to the object, using the attribute
         :attr:`kinto.core.resource.model.Model.id_field`.
--- a/kinto/core/storage/__init__.py
+++ b/kinto/core/storage/__init__.py
@@ -86,7 +86,7 @@ class StorageBase:
         :param str collection_id: the collection id.
         :param str parent_id: the collection parent.
         :param dict record: the object to create.
-        :param boolean ignore_conflict: Do not raise the UnicityError on conflict.
+        :param bool ignore_conflict: Do not raise the UnicityError on conflict.
 
         :returns: the newly created object.
         :rtype: dict
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -243,17 +243,20 @@ class Storage(StorageBase):
         VALUES (:object_id, :parent_id,
                 :collection_id, (:data)::JSONB,
                 from_epoch(:last_modified))
+        %(on_conflict)s
+        RETURNING id, as_epoch(last_modified) AS last_modified;
         """
+
+        safe_holders = {"on_conflict": ""}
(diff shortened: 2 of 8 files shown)


kinto/core/storage/testing.py
```diff
@@ -187,6 +187,15 @@ def test_create_raise_unicity_error_if_provided_id_exists(self):
...
                           record=record)
 
+    def test_create_does_not_raise_unicity_error_if_ignore_conflict_is_set(self):
+        record = {**self.record, self.id_field: RECORD_ID}
+        self.create_record(record=record)
+        record = {**self.record, self.id_field: RECORD_ID}
+        try:
+            self.create_record(record=record, ignore_conflict=True)
+        except exceptions.UnicityError:  # pragma: no cover
+            self.fail('It should not raise a UnicityError here.')
```
leplatrem: nit: Elsewhere we just add a trailing inline comment `# not raising`  :)

kinto/core/storage/postgresql/__init__.py
```diff
@@ -243,8 +243,18 @@ def create(self, collection_id, parent_id, record, id_generator=None,
...
+
+        if ignore_conflict:
+            # We use DO UPDATE so that the RETURNING clause works
+            # but we don't update anything and keep the previous
+            # last_modified value already stored.
+            query += """ON CONFLICT (id, parent_id, collection_id) DO UPDATE
+        SET last_modified = EXCLUDED.last_modified
+        """
+
+        query += "RETURNING id, as_epoch(last_modified) AS last_modified;"
```
leplatrem: Instead of concatenating, elsewhere in the code base we use a `safeholder`, like `%(on_conflics)s`. Here providing an empty string by default for example.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -243,8 +243,18 @@ def create(self, collection_id, parent_id, record, id_generator=None,
...
                 from_epoch(:last_modified))
-        RETURNING id, as_epoch(last_modified) AS last_modified;
         """
+
+        if ignore_conflict:
+            # We use DO UPDATE so that the RETURNING clause works
+            # but we don't update anything and keep the previous
+            # last_modified value already stored.
+            query += """ON CONFLICT (id, parent_id, collection_id) DO UPDATE
+        SET last_modified = EXCLUDED.last_modified
```
leplatrem: nit: don't worry too much about query indentation, I would prioritize python identation

kinto/core/storage/postgresql/__init__.py
```diff
@@ -243,8 +243,18 @@ def create(self, collection_id, parent_id, record, id_generator=None,
         VALUES (:object_id, :parent_id,
                 :collection_id, (:data)::JSONB,
                 from_epoch(:last_modified))
-        RETURNING id, as_epoch(last_modified) AS last_modified;
         """
+
+        if ignore_conflict:
+            # We use DO UPDATE so that the RETURNING clause works
+            # but we don't update anything and keep the previous
+            # last_modified value already stored.
```

kinto/core/storage/__init__.py
```diff
@@ -72,7 +72,7 @@ def collection_timestamp(self, collection_id, parent_id, auth=None):
     def create(self, collection_id, parent_id, record, id_generator=None,
                id_field=DEFAULT_ID_FIELD,
                modified_field=DEFAULT_MODIFIED_FIELD,
-               auth=None):
+               auth=None, ignore_conflict=False):
```
leplatrem: don't forget to add it in the docstring

kinto/plugins/default_bucket/__init__.py
```diff
@@ -101,14 +100,11 @@ def resource_create_object(request, resource_cls, uri):
...
     data = {'id': obj_id}
-    try:
-        obj = resource.model.create_record(data)
-        # Since the current request is not a resource (but a straight Service),
-        # we simulate a request on a resource.
-        # This will be used in the resource event payload.
-        resource.postprocess(data, action=ACTIONS.CREATE)
-    except storage_exceptions.UnicityError as e:
-        obj = e.record
+    obj = resource.model.create_record(data, ignore_conflict=True)
```
glasserc:
```python
# We use ignore_conflict here because each user can only ever have one default bucket, and it is created implicitly.
# If we got a conflict, it's from two requests by the same user, both creating a bucket.
# This isn't an error since creating the default bucket is idempotent.
```

PR #1123, opened 2017-03-02
Remove enum backport since we are now Python 3.5+

mozillazg:
It's the same as https://github.com/Kinto/kinto/pull/1111 😂 
r? @Natim


Natim:
Oh we refresh the requirements.txt file on release with the ``make build-requirements``
You can read why here: http://www.servicedenuages.fr/en/handling-python-projects-dependencies

PR #1124, opened 2017-03-02
use with ... open instead of open

mozillazg:
r? @leplatrem


Natim:
That's a good idea yes 👍

leplatrem:
Nice, thanks!

PR #1125, opened 2017-03-02
Return 409 on integrity error (ref #602)

leplatrem:
Ref #602

- [x] Reproduce/fix collection timestamp error
- [ ] Reproduce timestamp/object creation integrity error
- [x] Return 409 on integrity error
- [x] Add migrations
- [x] Add a changelog entry.


kinto/core/storage/postgresql/schema.sql
```diff
@@ -75,18 +75,20 @@ RETURNS TIMESTAMP AS $$
...
-
-    IF ts IS NULL THEN
-      ts := clock_timestamp();
+    WITH create_if_missing AS (
       INSERT INTO timestamps (parent_id, collection_id, last_modified)
-      VALUES (uid, resource, ts);
-    END IF;
+      VALUES (uid, resource, clock_timestamp())
+      ON CONFLICT (parent_id, collection_id) DO NOTHING
+      RETURNING last_modified
```
Natim: You cannot have DO NOTHING and RETURNING last_modified at the same time.

leplatrem: Why not? If it's inserted it will return the inserted value, if there's a conflict it won't return anything. That's why I did an `UNION` after to do like *create or get*

Natim: > If it's inserted it will return the inserted value, if there's a conflict it won't return anything

Not in fact it will never return anything even if it creates it.

leplatrem: Humm, that's not the behaviour I have, look:

```
[testdb] # INSERT INTO timestamps (parent_id, collection_id, last_modified)
      VALUES ('haha', 'hoho', NOW())
      ON CONFLICT (parent_id, collection_id) DO NOTHING
      RETURNING last_modified;
       last_modified        
----------------------------
 2017-03-02 16:45:25.445502
(1 row)

INSERT 0 1
Time: 5,195 ms
[testdb] # INSERT INTO timestamps (parent_id, collection_id, last_modified)
      VALUES ('haha', 'hoho', NOW())
      ON CONFLICT (parent_id, collection_id) DO NOTHING
      RETURNING last_modified;
 last_modified 
---------------
(0 rows)

INSERT 0 0
Time: 0,282 ms
```

tests/core/test_views_transaction.py
```diff
@@ -121,6 +124,28 @@ def test_modifications_are_rolled_back_on_error_accross_backends(self):
         resp = self.app.get('/psilos', headers=self.headers)
         self.assertEqual(len(resp.json['data']), 0)
 
+    def test_integrity_errors_are_returned_as_409(self):
```
leplatrem: XXX: wrong test title

kinto/core/storage/postgresql/schema.sql
Natim: Oh so we could use that to detect a conflict :)

PR #1095, opened 2017-02-20
Fix pserve argument ordering.

Natim:
Refs: Pylons/pyramid#2962

Fixes #1092

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

PR #1129, opened 2017-03-03
Preparing release 6.0.0

leplatrem:
- [x] Wait for #1095


gabisurita:
Should we wait for #1095?

leplatrem:
> Should we wait for #1095?

It was marked as WIP so I didn't bother... What do you think?

gabisurita:
I think we should probably fix the make serve recipe before releasing. I know it's more important for development but I think it's bad to make a release without it working.

Natim:
The thing is even with #1095 fixed, the --reload option won't work until https://github.com/Pylons/pyramid/pull/2962 is fixed.

Natim:
In the meantime we can remove --reload from the ``make serve``

PR #1134, opened 2017-03-07
Default bucket collection data are not kept.

Natim:
Refs https://github.com/Kinto/kinto-http.js/pull/164

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -218,10 +218,12 @@ class Storage(StorageBase):
                auth=None, ignore_conflict=False):
         id_generator = id_generator or self.id_generator
         record = {**record}
-        if id_field in record and not ignore_conflict:
+        if id_field in record:
             # Raise unicity error if record with same id already exists.
             try:
                 existing = self.get(collection_id, parent_id, record[id_field])
+                if ignore_conflict:
+                    return existing
                 raise exceptions.UnicityError(id_field, existing)
             except exceptions.RecordNotFoundError:
                 pass
--- a/tests/plugins/test_default_bucket.py
+++ b/tests/plugins/test_default_bucket.py
@@ -41,6 +41,17 @@ class DefaultBucketViewTest(FormattedErrorMixin, DefaultBucketWebTest):
         result = resp.json
         self.assertIn('system.Everyone', result['permissions']['read'])
 
+    def test_default_bucket_collection_can_still_be_explicitly_created(self):
+        collection = {'data': {'synced': True}}
+        resp = self.app.put_json(self.collection_url, collection, headers=self.headers)
+        result = resp.json
+        self.assertIn('synced', result['data'])
+        self.assertTrue('synced', result['data']['synced'])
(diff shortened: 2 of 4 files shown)


tests/plugins/test_default_bucket.py
```diff
@@ -41,6 +41,17 @@ def test_default_bucket_can_still_be_explicitly_created(self):
         result = resp.json
         self.assertIn('system.Everyone', result['permissions']['read'])
 
+    def test_default_bucket_collection_can_still_be_explicitly_created(self):
+        collection = {'data': {'synced': True}}
+        resp = self.app.put_json(self.collection_url, collection, headers=self.headers)
+        result = resp.json
+        self.assertIn('synced', result['data'])
+        self.assertTrue('synced', result['data']['synced'])
```
leplatrem: Why two parameters?

kinto/core/storage/postgresql/__init__.py
```diff
@@ -218,10 +218,12 @@ def create(self, collection_id, parent_id, record, id_generator=None,
                auth=None, ignore_conflict=False):
         id_generator = id_generator or self.id_generator
         record = {**record}
-        if id_field in record and not ignore_conflict:
+        if id_field in record:
             # Raise unicity error if record with same id already exists.
             try:
                 existing = self.get(collection_id, parent_id, record[id_field])
+                if ignore_conflict:
+                    return existing
```
leplatrem: We should still remove the tombstone in that case.

What about:

```diff
diff --git a/kinto/core/storage/memory.py b/kinto/core/storage/memory.py
index f86f10d..0650955 100644
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -135,22 +135,23 @@ class Storage(MemoryBasedStorage):
                modified_field=DEFAULT_MODIFIED_FIELD, auth=None, ignore_conflict=False):
         id_generator = id_generator or self.id_generator
         record = {**record}
         if id_field in record:
-            if not ignore_conflict:
-                # Raise unicity error if record with same id already exists.
-                try:
-                    existing = self.get(collection_id, parent_id, record[id_field])
-                    raise exceptions.UnicityError(id_field, existing)
-                except exceptions.RecordNotFoundError:
-                    pass
+            # Raise unicity error if record with same id already exists.
+            try:
+                record = self.get(collection_id, parent_id, record[id_field])
+                if not ignore_conflict:
+                    raise exceptions.UnicityError(id_field, record)
+            except exceptions.RecordNotFoundError:
+                pass
```


The storage tests on `ignore_conflict=True` should be more explicit regarding:
* removing the tombstone or not
* bumping the collection timestamp or not

leplatrem:
Are you sure the changes are necessary in the PostgreSQL backend?

If yes, then the `ignore_conflict` placeholder might not be necessary anymore no?

Natim:
> Are you sure the changes are necessary in the PostgreSQL backend?

Yes because tests are failing without it.

> then the ignore_conflict placeholder might not be necessary anymore no?

This one is not related directly.
Actually it is still mandatory to handle RaceCondition and IntegrityErros.

PR #1136, opened 2017-03-07
Provide old objects in history entries (fixes #1135)

### LINKED ISSUE #1135: Provide previous object in history deletion entries

leplatrem:
Fixes #1135

...wasn't that bad :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,13 @@ This document describes changes between each past release.
 6.1.0 (unreleased)
 ------------------
 
+**Protocol**
+
+- The history entries for deletions now provide the deleted object instead of
+  just the tombstone (fixes #1135)
+
+Protocol is now at version **1.15**. See `API changelog`_.
+
 **Bug fixes**
 
 - Fix Memory backend sometimes show empty permissions (#1045)
--- a/kinto/plugins/history/listener.py
+++ b/kinto/plugins/history/listener.py
@@ -11,6 +11,7 @@ def on_resource_changed(event):
     :mod:`kinto.plugins.history.views` module.
     """
     payload = event.payload
+    action = payload['action']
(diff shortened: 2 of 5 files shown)


kinto/plugins/history/listener.py
```diff
@@ -26,7 +27,7 @@ def on_resource_changed(event):
 
     targets = []
     for impacted in event.impacted_records:
-        target = impacted['new']
+        target = impacted['old'] if action == 'delete' else impacted['new']
```
n1k0: So we don't have that previously available `deleted` attribute from the tombstone available... should we keep it ? yeah that's super weird

leplatrem: I guess that your concern comes from this https://github.com/Kinto/kinto-http.js/blob/3d4069b29b675bbe61d2ce5465abb996f883169a/src/collection.js#L355  

But the right way would be to do ``entry.action == delete`` instead of ``entry.target.data.deleted == true`.

I find the breaking change acceptable though, but open to conversation. Adding `deleted: true` doesn't seem a good idea though :/

n1k0: > Adding deleted: true doesn't seem a good idea though :/

Yeah, definitely not a good option. Let's land it that way.

CHANGELOG.rst
```diff
@@ -6,6 +6,13 @@ This document describes changes between each past release.
 6.1.0 (unreleased)
 ------------------
 
+**Protocol**
+
+- The history entries for deletions now provide the deleted object instead of
+  just the tombstone (fixes #1135)
+
+Protocol is now at version **1.15**. See `API changelog`_.
```
gabisurita: The version here doesn't match the version on the __init__ file (should be 1.16 I guess)

PR #1137, opened 2017-03-07
Fix timestamp of deletion entries in history (ref #1135)

leplatrem:
Ref #1135


leplatrem:
Both changes are related and would probably require a test. BRB

tests/plugins/test_history.py
```diff
@@ -182,11 +182,13 @@ def test_tracks_collection_permissions_update(self):
...
 
     def test_tracks_collection_delete(self):
-        self.app.delete(self.collection_uri, headers=self.headers)
+        resp = self.app.delete(self.collection_uri, headers=self.headers)
+        tombstone = resp.json['data']
         resp = self.app.get(self.history_uri, headers=self.headers)
         entry = resp.json['data'][0]
         assert entry['action'] == 'delete'
         assert entry['target']['data']['signed']
+        assert entry['target']['data']['last_modified'] >= tombstone['last_modified']
```
Natim: Why `>=` rather than `==`

PR #1139, opened 2017-03-07
Groups can now be created with a simple PUT (fixes #793)

### LINKED ISSUE #793: Consistency on PUT with mandatory schema fields
While working on #790 I realize that there is something not clear in our specifications.

Currently, if a resource has a mandatory field (eg. groups `members`), then we cannot do a `PUT` with just the `permissions` values. This is because a PUT can lead to a creation, and the `members` fields has to be provided.

On other resources, which have no mandatory field, it is perfectly possible to only provide `permissions`.

But, I believe we should make every resources behave the same way.

For example, when we'll implement the edition of permissions in Kinto-admin, we don't want to have to pass the `data` if it was not changed.

Two solutions:
- Add a default value (`[]`) for the groups members attribute (_my prefered one, trivial and not absurd_)
- Allow `data` to be omitted only when the `PUT` replaces an existing object (_more complex to implement, but would work for any resource with mandatory fields_)

leplatrem:
Fixes #793

### DIFF
--- a/tests/test_views_groups.py
+++ b/tests/test_views_groups.py
@@ -51,6 +51,14 @@ class GroupViewTest(FormattedErrorMixin, BaseWebTest, unittest.TestCase):
                                  headers=self.headers)
         self.assertEqual(resp.json['data']['members'], [])
 
+    def test_groups_can_be_patched_without_members_attribute(self):
+        group = {'data': {'alias': 'admins'}}
+        resp = self.app.patch_json(self.record_url,
+                                   group,
+                                   headers=self.headers)
+        self.assertEqual(resp.json['data']['members'], ['fxa:user'])
+        self.assertEqual(resp.json['data']['alias'], 'admins')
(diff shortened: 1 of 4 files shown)


tests/test_views_groups.py
```diff
@@ -34,6 +34,23 @@ def test_groups_can_be_posted_without_id(self):
...
+                                 headers=self.headers)
+        self.assertNotIn('members', resp.json['data'])
+
+    def test_groups_can_be_create_without_members_attribute(self):
+        group = {'data': {'alias': 'admins'}}
+        resp = self.app.put_json(self.record_url,
+                                 group,
+                                 headers=self.headers)
+        self.assertEqual(resp.json['data']['members'], [])
+
```
gabisurita: Maybe this is stupid, but I think a test to check if we don't overwrite groups on PATCH would be good.

PR #1143, opened 2017-03-08
Prevent storing bytes in cache (fixes #1142)

### LINKED ISSUE #1142: Storing bytes in the memory backend should convert it to str.
- This is an inconsistency with other backends that returns unicode when storing bytes.

Natim:
I wonder if we should do it like that or rather to raise an error when trying to cache bytes.

Fixes #1142 

- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem

### DIFF
--- a/kinto/core/cache/testing.py
+++ b/kinto/core/cache/testing.py
@@ -101,7 +101,7 @@ class CacheTest:
         self.assertEqual(*setget('foobar', {'b': [1, 2]}))
         self.assertEqual(*setget('foobar', 3.14))
 
-    def test_bytes_are_converted_to_unicode(self):
+    def test_bytes_cannot_be_stored_in_the_cache(self):
         with pytest.raises(TypeError):
             self.cache.set('test', b'foo')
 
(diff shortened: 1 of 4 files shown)


leplatrem:
I'd rather raise when trying to cache bytes, so that we have clean signature in the interface

Natim:
Yes that was my opinion too. Let's do that then :) Also it might break a bunch of plugins but it doesn't matter.

leplatrem:
> it might break a bunch of plugins

We can bump a major, no problem. #1144 also introduces breaking changes in `kinto.core.testing`

kinto/core/cache/testing.py
```diff
@@ -100,6 +101,10 @@ def setget(k, v):
         self.assertEqual(*setget('foobar', {'b': [1, 2]}))
         self.assertEqual(*setget('foobar', 3.14))
 
+    def test_bytes_are_converted_to_unicode(self):
```
leplatrem: leftover: test title does not match

PR #1138, opened 2017-03-07
Enforce the permission endpoint when the admin plugin is included (fixes #1059)

leplatrem:
Fixes #1059

PR #1144, opened 2017-03-08
Speedup tests

leplatrem:
I hadn't run the whole tests suite locally for a while. It became incredibly slow.

The Kinto app is initialized in every test method of the views tests (via `__init__`) and `kinto.main()` takes from 0.2 to 0.4 sec (!!). I tried to dig in and could not find obvious bottleneck. So I went for a more radical change: why do we have to initialize the app in every test whereas once per test case would be enough?

I hence moved app initialization to `setUpClass()` and the first attempt was quite successful: 

Before:
```
$ py.test -k tests/test_views_
Results (94.58s):
     200 passed
```

After:

```
$ py.test -k tests/test_views_
Results (21.61s):
     200 passed
```

The current approach introduces a **breaking change** since `get_app_settings()` is now a classmethod. But we could find a hack to introspect the test class and pick the test instance method if available (to avoid migrating every plugin just now).

### DIFF
--- a/tests/swagger/support.py
+++ b/tests/swagger/support.py
@@ -12,21 +12,19 @@ from ..support import (BaseWebTest, MINIMALIST_BUCKET, MINIMALIST_GROUP,
 
 class SwaggerTest(BaseWebTest, unittest.TestCase):
 
-    settings = {
-        'includes': [
-            'kinto.plugins.history',
-            'kinto.plugins.admin',
-        ]
-    }
-
     @classmethod
     def setUpClass(cls):
-        # FIXME: solve memory issues from generating the spec multiple times
-        app = BaseWebTest().make_app(settings=cls.settings)
-
-        cls.spec_dict = app.get('/__api__').json
+        super().setUpClass()
+        cls.spec_dict = cls.app.get('/__api__').json
         cls.spec = Spec.from_dict(cls.spec_dict)
         cls.resources = build_resources(cls.spec)
+
+    @classmethod
+    def get_app_settings(cls, extras=None):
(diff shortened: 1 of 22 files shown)


leplatrem:
```
Results (158.43s):
    1715 passed
```

gabisurita:
Time is not the only issue but also memory. Pytest design doesn't free the memory from a test after it's completed which can make travis go on swap. That's why the tests were erroring in #1138.

I had similar issues and discussed this with @glasserc on #977. This should have been partially fixed on https://github.com/pytest-dev/pytest/pull/2046, but apparently they only freed pytest own objects and left test specific objects.

Some cross referencing:
https://github.com/pytest-dev/pytest/issues/1649
https://github.com/pytest-dev/pytest/pull/2046
http://stackoverflow.com/questions/20900380/py-test-excessive-memory-usage-with-large-number-of-tests

leplatrem:
On Travis, the total time goes from *37 min* to *11 min*.

tests/swagger/support.py
```diff
@@ -12,22 +12,20 @@
...
-    }
-
     @classmethod
     def setUpClass(cls):
-        # FIXME: solve memory issues from generating the spec multiple times
-        app = BaseWebTest().make_app(settings=cls.settings)
-
-        cls.spec_dict = app.get('/__api__').json
+        super().setUpClass()
+        cls.spec_dict = cls.app.get('/__api__').json
```
gabisurita: That's what I've mentioned 👍

glasserc:
I think this is a good change even disregarding pytest behavior. Repeated initialization isn't really valuable for us.

PR #1146, opened 2017-03-08
Removed Structlog binding and bottlenecks (fixes #603)

### LINKED ISSUE #603: [cliquet] Struclog is costly
Original: https://github.com/mozilla-services/cliquet/issues/173

Summary: structlog was found to be a bottleneck when we did some benchmarks.  Having the info local to the request object can improve it.

leplatrem:
Fixes #603


kinto/core/resource/__init__.py
```diff
@@ -167,6 +167,9 @@ def __init__(self, request, context=None):
             parent_id=parent_id,
             auth=auth)
 
+        # Initialize timestamp as soon as possible.
+        self.timestamp
```
Natim: Don't you need = something ?

kinto/core/testing.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, *args, **kwargs):
         self.prefixed_principals = self.effective_principals + [self.prefixed_userid]
         self.json = {}
         self.validated = {}
+        self.log_context = lambda **kw: kw
```
Natim: lambda will probably make flake8 to fail ...

leplatrem: Thx, but it's only when a lambda is assigned to a variable AFAIK

PR #1147, opened 2017-03-08
Move flush to a built-in plugin

gabisurita:
Since usually we usually don't want the flush endpoint enabled, in my opinion it makes sense to have it as a plugin for test purposes instead of as a built-in feature.

- [x] Update documentation.
- [x] Add a changelog entry.

Benefits:

- Modularity (no extra code on the server when disabled)
- Default interface (instead on a custom config variable)
- Simplify the server setup

Some considerations/questions:

- The way it's doesn't allow using flush with the `KINTO_FLUSH_ENDPOINT_ENABLED` environmental variable. Is this ok?
- This is a breaking change that may affect client test environments

r? @Natim @glasserc @n1k0


gabisurita:
Side note: if we are moving `__flush__` maybe it also makes sense to move the permission endpoint.

docs/configuration/settings.rst
```diff
@@ -480,6 +480,10 @@ list of Python modules:
 | ``kinto.plugins.admin``               | It is a Web admin UI to manage data from a Kinto server.                 |
 |                                       | (:ref:`more details <api-history>`).                                     |
 +---------------------------------------+--------------------------------------------------------------------------+
+| ``kinto.plugins.flush``               | Completely remove all data from the database backend for                 |
```
leplatrem: nit: Here I would mention `Adds an endpoint to completely...` no ?

CHANGELOG.rst
```diff
@@ -12,6 +12,8 @@ This document describes changes between each past release.
   class method (#1144)
 - ``kinto.core.api`` was renamed to ``kinto.core.openapi`` (#1145)
 - Forbid storing bytes in the cache backend. (#1143)
+- The flush endpoint is now a built-in plugin at ``kinto.plugins.flush`` and should
+  be enabled using the ``includes`` section of the configuration file. (#1147)
```
leplatrem: We should mention the dropped env variable here

leplatrem:
> if we are moving __flush__ maybe it also makes sense to move the permission endpoint.

+1

leplatrem:
Thanks!

docs/api/1.x/flush.rst
```diff
@@ -0,0 +1,47 @@
+.. _api-flush:
+
+
+Flush Endpoint
+##############
+
+
+The Flush endpoint is used to flush (completely remove) all data from the
+database backend. While this can be useful during development, but can be
```
Natim: database backends

PR #1150, opened 2017-03-09
Get rid of structlog (fixes #1110)

### LINKED ISSUE #1110: Consider moving to mozlog instead of the build-in Heka Renderer
Refs [Bug 1340674](https://bugzilla.mozilla.org/show_bug.cgi?id=1340674)
Refs http://mozbase.readthedocs.io/en/latest/mozlog.html

leplatrem:
Fixes #1110 

* Switched from structlog renderer to standard logging formatter
* Dropped Mozilla heka renderer
* Got rid of structlog
* Use `logging.getLogger(__name__)` instead of `from kinto.logs import logger`


leplatrem:
The default configuration with `make serve` does not output request summaries. We should fix that before merging.

leplatrem:
The fact that (sub)request logs are swallowed is to due to https://github.com/Pylons/pyramid/issues/2975 

I will see if I can fix it by limiting import side effects in `kinto/core/__init__.py` and `kinto/__init__.py`

leplatrem:
The only way I could achieve what I desired was to move the formatter out of the package.

Otherwise Python logging module will import the parents modules of the formatter, and `pserve` will get rid of them if not explicit defined in `.ini`. 

It is now fixed using an external package :)

leplatrem:
you're welcome :)

glasserc:
Nice explanation! If the Pyramid issue gets fixed, will we move the code back into Kinto, or do you think having it as an external package is a good long-term plan?

Natim:
> an external package is a good long-term plan

It is a good plan for me.

leplatrem:
It is a good plan mainly because we could have the same problem in `app.wsgi` too. 

The piece of code that recursively imports the parent modules of the specified formatter [is in Python logging stdlib](https://github.com/python/cpython/blob/v3.5.3/Lib/logging/config.py#L90-L102)

PR #1164, opened 2017-03-17
Raise 400 on batch unknown attributes

### LINKED ISSUE #1163: Batch requests accept additional attributes
We should probably forbid them.

```
> echo '{"requests": [], "foo": {}}' | http post http://localhost:8888/v1/batch 

HTTP/1.1 200 OK
Access-Control-Expose-Headers: Retry-After, Alert, Content-Length, Backoff
Content-Length: 16
Date: Thu, 16 Mar 2017 23:01:01 GMT
Server: waitress

{
    "responses": []
}
```

Edit: I just notice we currently test for batch ignoring additional attributes. Is this a feature?

gabisurita:
Fixes #1163

- [x] Add tests.
- [x] Add a changelog entry.
- [x] Add an API changelog entry [in the docs](https://github.com/Kinto/kinto/blob/master/docs/api/index.rst)

### DIFF
--- a/tests/core/test_views_batch.py
+++ b/tests/core/test_views_batch.py
@@ -175,9 +175,8 @@ class BatchSchemaTest(unittest.TestCase):
     def test_requests_is_mandatory(self):
         self.assertInvalid({})
 
-    def test_unknown_attributes_are_dropped(self):
-        deserialized = self.schema.deserialize({'requests': [], 'unknown': 42})
-        self.assertNotIn('unknown', deserialized)
+    def test_raise_invalid_on_unknown_attributes(self):
+        self.assertInvalid({'requests': [], 'unknown': 42})
 
     def test_list_of_requests_can_be_empty(self):
         self.schema.deserialize({'requests': []})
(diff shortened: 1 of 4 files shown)


Natim:
Thanks

tests/core/test_views_batch.py
```diff
@@ -247,12 +250,10 @@ def test_defaults_must_be_a_request_schema_if_specified(self):
         batch_payload = {'requests': [request], 'defaults': defaults}
         self.assertInvalid(batch_payload)
 
-    def test_unknown_defaults_are_ignored_silently(self):
+    def test_raise_invalid_on_default_unkown_attributes(self):
```
glasserc: `unknown` :)

Natim:
This seems to break kinto-signer: https://travis-ci.org/Kinto/kinto-signer/jobs/212165621

gabisurita:
This breaks the current release of kinto-http.py, but it has been fixed on https://github.com/Kinto/kinto-http.py/pull/122

Natim:
So we should release kinto-http.py :)

gabisurita:
> So we should release kinto-http.py :)

We definitely should.

PR #1167, opened 2017-03-22
Add a note about accounts API in the concepts doc

kulshekhar:
Fixes #

- [x] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.

r? @

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -70,7 +70,9 @@ to several objects.
 Authentication
 ==============
 
-Kinto doesn't provide users management, like user sign-up, password modification, etc.
+Kinto comes with an accounts API that allows user sign-up, password modification and user deletion. It allows you to configure an administrator account that has read/write access to all other accounts.
+
+By default, Kinto accounts are authenticated using basic authentication.
 
 Kinto can rely on a third-party called "`Identity provider <https://en.wikipedia.org/wiki/Identity_provider>`_"
 like *Auth0*, *Github* or *Google* in order to authenticate the incoming request and assign a :term:`user id`.
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -70,7 +70,9 @@ to several objects.
 Authentication
 ==============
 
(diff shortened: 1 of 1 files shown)


docs/concepts.rst
```diff
@@ -70,7 +70,9 @@ to several objects.
 Authentication
 ==============
 
-Kinto doesn't provide users management, like user sign-up, password modification, etc.
+Kinto comes with an accounts API that allows user sign-up, password modification and user deletion. It allows you to configure an administrator account that has read/write access to all other accounts.
+
+By default, Kinto accounts are authenticated using basic authentication.
```
leplatrem: Here I would mention that Kinto is not coupled to any authentication mecanism or provider. And then in another paragraph mention there's an `account` built-in plugin that provide user sign-up, ... etc with a link to the docs ``` :ref:`link <api-accounts>` ```

kulshekhar: > I don't think we should put the built-in account plugin too visible until it becomes enabled by default.

It was my understanding that by the time this branch is merged with master, this plugin would become the default and all related documentation needs to be updated to highlight this.

If this is not the case, it'd be helpful to have some pointers to which docs need to be updated.

My initial list consisted of the following:

- docs/core/quickstart.rst
- docs/concepts.rst
- docs/faq.rst
- docs/configuration/settings.rst
- docs/tutorials/first-steps.rst
- docs/api/1.x/authentication.rst

kulshekhar:
> I don't think we should put the built-in account plugin too visible until it becomes enabled by default.

It was my understanding that by the time this branch is merged with master, this plugin would become the default and all related documentation needs to be updated to highlight this.

If this is not the case, it'd be helpful to have some pointers to which docs need to be updated.

My initial list consisted of the following:

- docs/core/quickstart.rst
- docs/concepts.rst
- docs/faq.rst
- docs/configuration/settings.rst
- docs/tutorials/first-steps.rst
- docs/api/1.x/authentication.rst

PR #1181, opened 2017-04-04
A better comment to describe PERMISSIONS_INHERITANCE_TREE

tiberiuichim:
The already existing comment was confusing and hard to understand when describing a vital variable. This PR improves it.


glasserc:
Thanks!

PR #1184, opened 2017-04-11
Update cornice swagger to 0.5.1 (fixes #1180)

### LINKED ISSUE #1180: Swagger spec still invalid
The Swagger spec is still invalid it seems

Extract:

```json
/{prefix}/{api_ver:\\d+}/{application_guid}/{application_ver}/{metrics:.*}": {
      "parameters": [
        {
          "name": "prefix",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "api_ver:\\d+",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "application_guid",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "application_ver",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "metrics:.*",
          "type": "string",
          "required": true,
          "in": "path"
        }
      ],
```

In this API definitions, smwogger will choke on api_ver and metrics because their definitions in the parameter list should not include the regexp and just be the name of the parameter

The right definition should be

```json
/{prefix}/{api_ver:\\d+}/{application_guid}/{application_ver}/{metrics:.*}": {
      "parameters": [
        {
          "name": "prefix",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "api_ver",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "application_guid",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "application_ver",
          "type": "string",
          "required": true,
          "in": "path"
        },
        {
          "name": "metrics",
          "type": "string",
          "required": true,
          "in": "path"
        }
      ],
```

To validate that it works, you can try this small program with Smowgger to print all operations:

```python
import asyncio
from smwogger import API


async def print_operations():
    async with API('http://path.to.kinto/v1/__api__') as api:
        print(api.operations)

loop = asyncio.get_event_loop()
try:
    loop.run_until_complete(print_operations())
finally:
    loop.close()
```

cc @gabisurita @chartjes

gabisurita:
Fixes #1180

cornice_swagger 0.5.1 supports subpath and regex on pyramid routes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ colander==1.3.2
 colorama==0.3.7
 contextlib2==0.5.4
 cornice==2.4.0
-cornice-swagger==0.5.0
+cornice-swagger==0.5.1
 hupper==0.4.2
 iso8601==0.1.11
 jsonpatch==1.15
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ REQUIREMENTS = [
     'bcrypt',
     'colander >= 1.3.2',
     'cornice >= 2.4',
-    'cornice_swagger >= 0.5',
+    'cornice_swagger >= 0.5.1',
     'jsonschema',
     'jsonpatch',
     'logging-color-formatter >= 1.0.1',  # Message interpolations.


Natim:
Thanks.

Natim:
Maybe it worth updating the CHANGELOG with a note about which bug got fixed by it.

leplatrem:
I updated the changelog, thanks Gabi!

PR #1186, opened 2017-04-11
Prepare 6.0.4

leplatrem:


kinto/plugins/admin/npm-debug.log.1455376938
```diff
@@ -0,0 +1,22 @@
+0 info it worked if it ends with ok
```
Natim: I don't think we want to keep that log.

PR #1204, opened 2017-05-01
Fix installation instructions.

almet:
The virtualenv needs to be created with python3, otherwise errors will occur.
Also update the OS dependencies accordingly.


leplatrem:
👍 thanks!

PR #1206, opened 2017-05-02
Removed useless deletions in quota plugin

leplatrem:


leplatrem:
Some failing tests because of #1208

glasserc:
In addition to removing deletions which aren't necessary, @leplatrem responded to my confusion about what the heck was going on in the plugin by renaming some variables. Remember, in Cliquet, `collection_id` meant something more like `resource_name` (see #710). Also, the IDs are not IDs of buckets and collections, but rather of "records" that track bucket and collection quota info respectively.

leplatrem:
Now that you found the root cause of our misery (#1211), I'll merge this one :)

PR #1208, opened 2017-05-02
Fix tests with Pyramid 1.9a

leplatrem:


glasserc:
The actual failure is `plaster.exceptions.InvalidURI: Could not determine the loader scheme for the supplied config_uri.`

PR #1227, opened 2017-05-22
fix #1224: exception in accounts plugin authentication

### LINKED ISSUE #1224: exception in accounts plugin authentication: Unicode-objects must be encoded before hashing
Running a fresh docker image built from master (36dbcc0) with postgresql storage

After:
```
echo '{"data": {"password": "PASSWD"}}' | http PUT http://localhost:8889/v1/accounts/eric
```
Running
```
http GET http://localhost:8889/v1/ --auth eric:coucou3333
```
Triggers following exception :
```
Unicode-objects must be encoded before hashing 
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/pyramid/tweens.py", line 22, in excview_tween
    response = handler(request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid_tm/__init__.py", line 119, in tm_tween
    reraise(*exc_info)
  File "/usr/local/lib/python3.5/dist-packages/pyramid_tm/compat.py", line 15, in reraise
    raise value
  File "/usr/local/lib/python3.5/dist-packages/pyramid_tm/__init__.py", line 98, in tm_tween
    response = handler(request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/router.py", line 155, in handle_request
    view_name
  File "/usr/local/lib/python3.5/dist-packages/pyramid/view.py", line 612, in _call_view
    response = view_callable(context, request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/config/views.py", line 181, in __call__
    return view(context, request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/viewderivers.py", line 389, in attr_view
    return view(context, request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/viewderivers.py", line 367, in predicate_wrapper
    return view(context, request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/viewderivers.py", line 438, in rendered_view
    result = view(context, request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/viewderivers.py", line 147, in _requestonly_view
    response = view(request)
  File "/usr/local/lib/python3.5/dist-packages/cornice/service.py", line 493, in wrapper
    response = view_(request)
  File "/code/kinto/core/views/hello.py", line 45, in get_hello
    if Authenticated in request.effective_principals:
  File "/usr/local/lib/python3.5/dist-packages/pyramid/security.py", line 375, in effective_principals
    return policy.effective_principals(self)
  File "/usr/local/lib/python3.5/dist-packages/pyramid_multiauth/__init__.py", line 119, in effective_principals
    userid = policy.authenticated_userid(request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/authentication.py", line 92, in authenticated_userid
    callback_ok = self.callback(userid, request)
  File "/usr/local/lib/python3.5/dist-packages/pyramid/authentication.py", line 1123, in callback
    return self.check(username, password, request)
  File "/code/kinto/plugins/accounts/authentication.py", line 18, in account_check
    if hashed == bcrypt.hashpw(pwd_str, hashed):
  File "/usr/local/lib/python3.5/dist-packages/bcrypt/__init__.py", line 62, in hashpw
    raise TypeError("Unicode-objects must be encoded before hashing")
```

It is fixed by encoding the value received from the session at [this line in authentication.py](https://github.com/Kinto/kinto/blob/master/kinto/plugins/accounts/authentication.py#L16):
```
 hashed = existing['password'].encode(encoding='utf8')
```

elelay:
Unicode-objects must be encoded before hashing.

Fixes #1224 

- [X] Add a changelog entry.

r? @Natim

### DIFF
--- a/kinto/plugins/accounts/views.py
+++ b/kinto/plugins/accounts/views.py
@@ -89,9 +89,11 @@ class Account(resource.ShareableResource):
     def process_record(self, new, old=None):
         new = super(Account, self).process_record(new, old)
 
-        # Store password safely in database.
+        # Store password safely in database as str
+        # (bcrypt.hashpw returns base64 bytes).
         pwd_str = new["password"].encode(encoding='utf-8')
-        new["password"] = bcrypt.hashpw(pwd_str, bcrypt.gensalt())
+        hashed = bcrypt.hashpw(pwd_str, bcrypt.gensalt())
+        new["password"] = hashed.decode(encoding='utf-8')
 
         # Administrators can reach other accounts and anonymous have no
         # selected_userid. So do not try to enforce.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ This document describes changes between each past release.
   the ``quotas`` plugin to no longer leak "quota" when used with the
   ``default_bucket`` plugin. (#1226)
 - Do not allow to reuse deletion tokens (fixes #1171)
+- ``accounts`` plugin: fix exception on authentication. (#1224)
 
 
 7.0.1 (2017-05-17)
--- a/kinto/plugins/accounts/authentication.py
+++ b/kinto/plugins/accounts/authentication.py
@@ -13,7 +13,7 @@ def account_check(username, password, request):
     except storage_exceptions.RecordNotFoundError:
         return None
 
-    hashed = existing['password']
+    hashed = existing['password'].encode(encoding='utf-8')
     pwd_str = password.encode(encoding='utf-8')
     if hashed == bcrypt.hashpw(pwd_str, hashed):
         return True  # Match! Return anything but None.


elelay:
I can't find where in the mocks is the non-unicode password stored.
Because [this existing test](https://github.com/Kinto/kinto/blob/master/tests/plugins/test_accounts.py#L52)  should fail.

elelay:
Well, now tests do fail...

glasserc:
Thanks for your contribution! The tests you mention don't seem to actually use a mock. Instead I think they just use the `memory` backend. It's possible that storing a `bytes` to Postgres makes it come back as `str` but come back as `bytes` in the memory backend. If that's true, I wonder how we're storing the "bytes" of the hashed password in Postgres? We might be accidentally storing it as a JSON string full of `\uXXXX`. If that's true, I think the best fix would be to take the hashed password and encode it as base64 in the `Account` view (which reduces storage use as well as giving us a consistent type to work with).

Natim:
The storage API is supposed to always store unicode and return unicode (`str`) reason why you need to encode the password before passing it to the hashlib API. (which only supports bytes hashing)

elelay:
@glasserc, @Natim indeed, tests [do use](https://github.com/Kinto/kinto/blob/master/kinto/core/testing.py#L158) the memory backend, which [doesn't touch the given record](https://github.com/Kinto/kinto/blob/master/kinto/core/storage/memory.py#L153)

On the contrary, the postgres backend [sneakily](https://github.com/Kinto/kinto/blob/master/kinto/core/storage/postgresql/__init__.py#L10) uses [ujson.dumps](https://github.com/Kinto/kinto/blob/master/kinto/core/utils.py#L14) to pass the record [as json string](https://github.com/Kinto/kinto/blob/master/kinto/core/storage/postgresql/__init__.py#L271) to PostgreSQL.
`json.dumps({'hello': 'haché'.encode(encoding='utf-8')})` explodes, while `ujson.dumps({'hello': 'haché'.encode(encoding='utf-8')})` happily returns `'{"hello":"hach\\u00e9"}'`. 

Following the `PUT` to create account:
 - the password is [hashed and put in the record](https://github.com/Kinto/kinto/blob/master/kinto/plugins/accounts/views.py#L94) as bytes
 - with memory backend, the record is stored as-is in memory  => we get bytes in `authentication.py`
 - with postgres backend, the record is converted to utf-8 JSON => we get str in `authentication.py`

Now I think the right way to fix the issue is not in the authentication as I did, but storing the hashed password as string, explicitly calling decode on bcrypt's result, which is [base64](https://github.com/pyca/bcrypt/blob/master/src/_csrc/bcrypt.c#L163) anyway. This way, tests and production work.

- Do you concur? 
- Is it ok if I force-push the new solution in this PR or do I open a new one?


 - maybe the memory backend should call `record =  ujson.loads(ujson.dumps(record))` before storing for consistency with the postgres backend (yes it adds the conversion cost),
- as a side note, performance may be gained by setting ujson.ensure_ascii to false in postgres backend, if it's supported by psycopg2,

Natim:
>  I think the right way to fix the issue is not in the authentication as I did, but storing the hashed password as string, explicitly calling decode on bcrypt's result, which is base64 anyway. This way, tests and production work.

Yes please 👍 you can force push on your branch if you want.

> ujson.ensure_ascii to false in postgres backend

This 👍

glasserc:
When you say `decode on bcrypt's result`, do you mean base64 => bytes, or do you mean base64 bytes => base64 str? If I understand correctly, the second sounds OK to me.

glasserc:
I feel like there's another larger issue here which is that we should only support storing true JSON records -- meaning, we should have barfed when the hashed password (as `bytes`) tried to get stored to Postgres. Is there an option for ujson to let us *not* serialize bytes, but only the usual booleans, integers, floats, strings, lists, dicts, and None?

elelay:
@glasserc 
> Is there an option for ujson to let us not serialize bytes, but only the usual booleans, integers, floats, strings, lists, dicts, and None?

No, there isn't : str and bytes are handled [the same way](https://github.com/esnme/ultrajson/blob/master/python/objToJSON.c#L543) - lines 543,550.
But ujson will raise on non utf8 bytes, though, so it's safe-ish.

To my mind, what should be enforced is consistency between memory and postgresql storage, by calling `record = ujson.loads(ujson.dumps(record))` before storing in the memory backed (in `memory.Storage.create` and `memory.Storage.update`.
It would have caught this issue and would also catch non-dict objects, etc.

glasserc:
Great work, solid investigation! Thank you so much for your effort on this!

kinto/plugins/accounts/views.py
```diff
@@ -89,9 +89,11 @@ def collection_post(self):
     def process_record(self, new, old=None):
         new = super(Account, self).process_record(new, old)
 
-        # Store password safely in database.
+        # Store password safely in database as str
+        # (bcrypt.hashpw returns base64 bytes).
         pwd_str = new["password"].encode(encoding='utf-8')
-        new["password"] = bcrypt.hashpw(pwd_str, bcrypt.gensalt())
+        hashed = bcrypt.hashpw(pwd_str, bcrypt.gensalt())
```
Natim: 👍 thanks

PR #1228, opened 2017-05-22
Demonstrate problem with default_bucket and quotas plugins

### LINKED ISSUE #1226: GET on default bucket increases quota usage
While investigating #1225, I discovered that *every* GET request to e.g. `/buckets/default/collections/some-collection` causes quota usage to increase. For example,

```
select * from records where collection_id='quota';
       id        |                                    parent_id                 
                   | collection_id |       last_modified        |               
                data                                
-----------------+--------------------------------------------------------------
-------------------+---------------+----------------------------+---------------
----------------------------------------------------
 bucket_info     | /buckets/74f86a68-a03c-a5f5-fb29-e421b362f5b5                
                   | quota         | 2017-05-22 18:07:14.715782 | {"record_count
": 0, "storage_size": 1745, "collection_count": 13}
 collection_info | /buckets/74f86a68-a03c-a5f5-fb29-e421b362f5b5/collections/a-n
ewest-collection-3 | quota         | 2017-05-22 18:07:14.716923 | {"record_count
": 0, "storage_size": 540}
```

```
> http --auth 'user:pass6' 'localhost:8888/v1/buckets/default/collections/a-newest-collection-3'
HTTP/1.1 200 OK
..... more junk .....
```

```
select * from records where collection_id='quota';
       id        |                                    parent_id                 
                   | collection_id |       last_modified        |               
                data                                
-----------------+--------------------------------------------------------------
-------------------+---------------+----------------------------+---------------
----------------------------------------------------
 bucket_info     | /buckets/74f86a68-a03c-a5f5-fb29-e421b362f5b5                
                   | quota         | 2017-05-22 18:56:52.448115 | {"record_count
": 0, "storage_size": 1880, "collection_count": 14}
 collection_info | /buckets/74f86a68-a03c-a5f5-fb29-e421b362f5b5/collections/a-n
ewest-collection-3 | quota         | 2017-05-22 18:56:52.4493   | {"record_count
": 0, "storage_size": 600}
```

This appears to be partly because every request that goes to the default bucket "creates" a bucket and collection.

glasserc:
Fixes #1226.

- (n/a) Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @Natim

### DIFF
--- a/tests/plugins/test_default_bucket.py
+++ b/tests/plugins/test_default_bucket.py
@@ -338,6 +338,20 @@ class EventsTest(DefaultBucketWebTest):
         assert _events[2].payload['uri'] == records_uri.replace('default',
                                                                 bucket_id)
 
+    def test_second_call_on_default_bucket_doesnt_send_create_events(self):
+        bucket_url = '/buckets/default'
+        self.app.get(bucket_url, headers=self.headers)
+        assert len(_events) == 1
+        self.app.get(bucket_url, headers=self.headers)
+        assert len(_events) == 1
+
+    def test_second_call_on_default_bucket_collection_doesnt_send_create_events(self):
+        collection_url = '/buckets/default/collections/articles'
+        self.app.get(collection_url, headers=self.headers)
+        assert len(_events) == 2
+        self.app.get(collection_url, headers=self.headers)
+        assert len(_events) == 2
+
 
 class ReadonlyDefaultBucket(DefaultBucketWebTest):
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,13 @@ This document describes changes between each past release.
 
 - ``delete()`` method from cache backend now returns the deleted value (fixes #1231)
 
+**Bug fixes**
+
+- The ``default_bucket`` plugin no longer sends spurious "created"
+  events for buckets and collections that already exist. This causes
+  the ``quotas`` plugin to no longer leak "quota" when used with the
+  ``default_bucket`` plugin. (#1226)
+
 
 7.0.1 (2017-05-17)
 ------------------
--- a/kinto/plugins/default_bucket/__init__.py
+++ b/kinto/plugins/default_bucket/__init__.py
@@ -6,6 +6,7 @@ from pyramid.security import NO_PERMISSION_REQUIRED, Authenticated
 
 from kinto.core.errors import raise_invalid
 from kinto.core.events import ACTIONS
+from kinto.core.storage.exceptions import UnicityError
 from kinto.core.utils import (
     build_request, reapply_cors, hmac_digest, instance_uri, view_lookup)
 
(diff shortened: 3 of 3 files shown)


glasserc:
OK, I guess this was easier to fix than I was worried it would be.

tests/plugins/test_quotas.py
```diff
@@ -544,6 +544,50 @@ def test_bulk_delete(self):
                              object_id=COLLECTION_QUOTA_OBJECT_ID)
 
 
+class QuotaDefaultBucketTest(QuotaWebTest):
+    headers = {"Content-Type": "application/json",
+               "Authorization": "Basic dXNlcjpwYXNzNg=="}
```
leplatrem: nit: we have a `get_user_headers()` helper

tests/plugins/test_quotas.py
```diff
@@ -544,6 +544,50 @@ def test_bulk_delete(self):
...
+                                        object_id=BUCKET_QUOTA_OBJECT_ID)
+
+        self.app.get("/buckets/default/collections/some-random-collection",
+                     headers=self.headers)
+        later_data = self.storage.get(collection_id=QUOTA_RESOURCE_NAME,
+                                      parent_id=bucket_uri,
+                                      object_id=BUCKET_QUOTA_OBJECT_ID)
+        self.assertStatsEqual(initial_data, later_data)
+
+    def test_default_bucket_doesnt_increase_quota_for_collection(self):
```
Natim: Does this means that we can store as many data as we want in the collection to get infinite storage quota? Same question for bucket metadata.

Natim: After thinking about this a bit more, I think it is fixing the fact that once it exists, it will not keep increasing it. Is that correct?

glasserc:
I decided that the tests really belonged on `default_bucket`, since that's the component that changed. I rewrote the branch and threw away the tests on the quotas plugin.

tests/plugins/test_default_bucket.py
```diff
@@ -338,6 +338,20 @@ def test_events_sent_on_bucket_and_collection_creation(self):
...
+        assert len(_events) == 1
+        self.app.get(bucket_url, headers=self.headers)
+        assert len(_events) == 1
+
+    def test_second_call_on_default_bucket_collection_doesnt_send_create_events(self):
+        collection_url = '/buckets/default/collections/articles'
+        self.app.get(collection_url, headers=self.headers)
+        assert len(_events) == 2
+        self.app.get(collection_url, headers=self.headers)
+        assert len(_events) == 2
```
Natim: A comment telling which events are sent there would be great. I guess it is the read event.

glasserc: They're the same create events from `test_an_event_is_sent_on_implicit_bucket_creation` and `test_an_event_is_sent_on_implicit_collection_creation`. Since they're already there, I didn't think it was worth rehashing them. But a comment is a good idea.

PR #1233, opened 2017-05-23
Clear timestamps table when deleting parents (fixes #1223)

### LINKED ISSUE #1223: Rows from the timestamps table are never deleted
When we delete a bucket, all its descendants are deleted. But the `timestamps` table will still retain the row for all its collections and groups.

For clients, it means that when you recreate a collection, it will reuse the previous timestamp instead of a new one.

leplatrem:
Fixes #1223 

- [x] Fail to reproduce in tests
- [x] Succeed to reproduce in tests
- [x] Fix bug
- [ ] Add a changelog entry.

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -503,11 +503,23 @@ class Storage(StorageBase):
                       modified_field=DEFAULT_MODIFIED_FIELD,
                       auth=None):
         delete_tombstones = """
-        DELETE
-        FROM deleted
-        WHERE {parent_id_filter}
-              {collection_id_filter}
-              {conditions_filter}
+        WITH delete_tombstones AS (
+            DELETE
+            FROM deleted
+            WHERE {parent_id_filter_tombstones}
+                  {collection_id_filter}
+                  {conditions_filter}
+            RETURNING parent_id
+        ),
+        delete_timestamps AS (
+            DELETE
+            FROM timestamps
+            WHERE {parent_id_filter_timestamps}
+            RETURNING parent_id
+        )
+        SELECT * FROM delete_tombstones
+        UNION
+        SELECT * FROM delete_timestamps
         """
         id_field = id_field or self.id_field
         modified_field = modified_field or self.modified_field
@@ -517,10 +529,18 @@ class Storage(StorageBase):
         safeholders = defaultdict(str)
         # Handle parent_id as a regex only if it contains *
         if '*' in parent_id:
-            safeholders['parent_id_filter'] = 'parent_id LIKE :parent_id'
+            safeholders['parent_id_filter_tombstones'] = 'parent_id LIKE :parent_id'
             placeholders['parent_id'] = parent_id.replace('*', '%')
         else:
-            safeholders['parent_id_filter'] = 'parent_id = :parent_id'
+            safeholders['parent_id_filter_tombstones'] = 'parent_id = :parent_id'
+
+        # If purging everything from a parent_id, then it means we won't have
+        # anything left for this resource. We can wipe out timestamps.
+        if collection_id is None and before is None:
+            safeholders['parent_id_filter_timestamps'] = safeholders['parent_id_filter_tombstones']
+        else:
+            safeholders['parent_id_filter_timestamps'] = '0 = 1'  # never true.
+
         # If collection is None, remove it from query.
         if collection_id is None:
             safeholders['collection_id_filter'] = ''
@@ -534,18 +554,7 @@ class Storage(StorageBase):
 
         with self.client.connect() as conn:
             result = conn.execute(delete_tombstones.format_map(safeholders), placeholders)
-            deleted = result.rowcount
-
-            # If purging everything from a parent_id, then clear timestamps.
-            if collection_id is None and before is None:
-                delete_timestamps = """
-                DELETE
-                FROM timestamps
-                WHERE {parent_id_filter}
-                """
-                conn.execute(delete_timestamps.format_map(safeholders), placeholders)
-
-        return deleted
+            return result.rowcount
 
     def get_all(self, collection_id, parent_id, filters=None, sorting=None,
                 pagination_rules=None, limit=None, include_deleted=False,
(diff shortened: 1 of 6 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -533,9 +533,19 @@ def purge_deleted(self, collection_id, parent_id, before=None,
             placeholders['before'] = before
 
         with self.client.connect() as conn:
-            result = conn.execute(query.format_map(safeholders), placeholders)
+            result = conn.execute(delete_tombstones.format_map(safeholders), placeholders)
+            deleted = result.rowcount
+
+            # If purging everything from a parent_id, then clear timestamps.
+            if collection_id is None and before is None:
```
glasserc: Do we need to check that `with_deleted` is `False` too?

kinto/core/storage/postgresql/__init__.py
```diff
@@ -533,9 +533,19 @@ def purge_deleted(self, collection_id, parent_id, before=None,
...
+                delete_timestamps = """
+                DELETE
+                FROM timestamps
+                WHERE {parent_id_filter}
+                """
+                conn.execute(delete_timestamps.format_map(safeholders), placeholders)
```
glasserc: Should we try to combine these queries into one? Or does `READ COMMITTED` somehow protect us from race conditions?

kinto/core/storage/testing.py
```diff
@@ -981,6 +981,28 @@ def test_purge_deleted_remove_all_tombstones_by_parent_id(self):
...
+        before2 = self.storage.collection_timestamp(parent_id='efg', collection_id='c')
+        # Different parent_id without record.
+        before3 = self.storage.collection_timestamp(parent_id='ijk', collection_id='c')
+
+        self.storage.delete_all(parent_id='/abc/*', collection_id=None, with_deleted=False)
+        self.storage.purge_deleted(parent_id='/abc/*', collection_id=None)
+
+        after1 = self.storage.collection_timestamp(parent_id='/abc/a', collection_id='c')
+        after2 = self.storage.collection_timestamp(parent_id='efg', collection_id='c')
+        after3 = self.storage.collection_timestamp(parent_id='ijk', collection_id='c')
```
glasserc: Are the slashes in `/abc/a` part of why the behavior is different? Can we make them all have slashes?

kinto/core/storage/postgresql/__init__.py
leplatrem: `with_deleted` is not part of `purge_deleted()` (?)

glasserc: Oh, oops.. I thought this was being added to `delete_all`. Never mind..

kinto/core/storage/postgresql/__init__.py
leplatrem: Unless I am completely confused at this time of the day, the second query does not depend on the result of the first one, so we can safely execute those two queries separately. They will be executed in the same transaction anyway.

glasserc: What I was wondering is:

1. T1 deletes a collection "record".
2. T2 inserts a collection "record" of the same name and commits. (Because of READ COMMITTED, it doesn't see the collection in T1 anyhow.)
3. T3 looks for the records that are children of that collection, causing a `timestamps` record to be written.
4. T1 then does `purge_deleted` and wipes the timestamps created in T3 (which it sees because of `READ COMMITTED`).

But I'm not really sure I understand SQL isolation well enough to know if this is a problem, and even if it is, we may not really be able to do anything about it. By the time we get to `purge_deleted`, we've already executed the delete query for the collection and deleted all of its records..

kinto/core/storage/postgresql/__init__.py
glasserc: Thinking about this more, it's a little strange that purging tombstones is the thing that triggers removal from the `timestamps` collection. If I delete a bucket, surely its "collection" timestamps should be deleted before I delete any records that are "under it" or purge their tombstones? Maybe the problem is just that `purge_deleted` is a bad name for this method, and we should change the concept to something like `recursive_delete` (which is outside the scope of this PR).

leplatrem: Indeed, the tombstones are the last bit of info that we keep about a resource. Once we delete the tombstones, we can wipe out timestamps.

I agree that changing the name would be the best option.

kinto/core/storage/postgresql/__init__.py
leplatrem: I'll write it as a single op, it does not harm :)

leplatrem: Oh, and thanks for the detailed explanation 😊

leplatrem:
@glasserc in you have some bandwidth, I'd gladly ask for help here please...

 I don't explain why the returned value of `purge_deleted()` has changed with my last changes... It is supposed the number of tombstones that were deleted, and I don't explain why it's not behaving as expected :/ 

Thanks!

glasserc:
Isn't it because your query now returns the deleted tombstones `UNION` the deleted timestamps and returns the rowcount as the number of deleted things?

To be honest, I don't think this fixes the problem and I am no longer sure it's worth fixing. I'm not as experienced with SQL stuff as you are but the [Postgres documentation](https://www.postgresql.org/docs/current/static/transaction-iso.html) says "UPDATE, DELETE, SELECT FOR UPDATE, and SELECT FOR SHARE commands behave the same as SELECT in terms of searching for target rows: they will only find target rows that were committed as of the command start time." My interpretation of this is that if you have two DELETE queries as subqueries of a single SELECT, things can be committed after the first one and before the second one, which is the thing we wanted to avoid.

I think we should probably just go back to the way the code was before this last commit and (as a follow-up) make the isolation level a Kinto-wide configuration option. Otherwise we're just trying to force transactional semantics by wedging our queries together.

leplatrem:
> Isn't it because your query now returns the deleted tombstones UNION the deleted timestamps and returns the rowcount as the number of deleted things?

Oops, sorry I had an uncommitted local change that was reading the value of `SELECT COUNT(*) FROM delete_tombstones`... But still it didn't work as expected.

I reverted the hack to have two delete in a single query. Thanks a lot for helping :]

PR #1235, opened 2017-05-23
rebuild-quotas script and tests

### LINKED ISSUE #1230: Offer a "reset quotas" endpoint
To pick up after the mess of #1226, we'll need to wipe all of the quota information for current users and rebuild it from the data in the database. This might be an expensive operation so be sure to limit access to the superuser or only to people behind the VPN or something (@jasonthomas might have an opinion?).

glasserc:
Super WIP. Feedback earnestly solicited on the approach, and especially whether it's right to use `kinto.storage` operations rather than a SQL query, and whether stuffing tests full of scripted method calls is remotely acceptable.

Fixes #1230.

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem, @Natim


kinto/__main__.py
```diff
@@ -23,7 +23,8 @@ def main(args=None):
 
     parser = argparse.ArgumentParser(description="Kinto Command-Line "
                                                  "Interface")
-    commands = ('init', 'start', 'migrate', 'delete-collection', 'version')
+    commands = ('init', 'start', 'migrate', 'delete-collection', 'version',
+                'rebuild-quotas')
```
leplatrem: nitpick: we could add `rebuild-quotas` only if plugin is included?

kinto/core/scripts.py
```diff
@@ -88,3 +91,110 @@ def delete_collection(env, bucket_id, collection_id):
...
+
+        bucket_pagination = [
+            [Filter('last_modified', bucket['last_modified'], COMPARISON.GT)]
+        ]
+
+    current_transaction.commit()
+    return 0
+
+
+def rebuild_quotas_collection(env, bucket_id, collection, dry_run=False):
```
leplatrem: I would pass storage backend instead of `env`

kinto/core/scripts.py
```diff
@@ -88,3 +91,110 @@ def delete_collection(env, bucket_id, collection_id):
...
+    """Helper method for rebuild_quotas that updates a single collection."""
+    registry = env['registry']
+    record_pagination = None
```
leplatrem: nit: put `record_pagination` closer to the loop where it gets updated

kinto/core/scripts.py
```diff
@@ -88,3 +91,110 @@ def delete_collection(env, bucket_id, collection_id):
...
+        record_pagination = [
+            [Filter('last_modified', record['last_modified'], COMPARISON.GT)]
+        ]
+
+    logger.info("Bucket {}, collection {}. Final size: {} records, {} bytes.".format(
+        bucket_id, collection_id, collection_record_count, collection_storage_size))
+    return {
+        "record_count": collection_record_count,
+        "storage_size": collection_storage_size,
+    }
```
leplatrem: why not return a tuple instead ?

tests/core/test_scripts.py
```diff
@@ -119,3 +122,183 @@ def test_delete_collection_raise_if_the_collection_does_not_exist(self):
...
+        self.registry.storage.update.assert_any_call(
+            collection_id='quota',
+            parent_id='/buckets/bucket-1/collections/collection-2',
+            object_id='collection_info',
+            record={'record_count': 1, 'storage_size': 160})
+
+        mocked.info.assert_any_call('Bucket bucket-1, collection collection-1. Final size: 1 records, 78 bytes.')
+        mocked.info.assert_any_call('Bucket bucket-1, collection collection-2. Final size: 1 records, 79 bytes.')
+        mocked.info.assert_any_call('Bucket bucket-1. Final size: 2 records, 193 bytes.')
+
```
leplatrem: LGTM.

I'm not super convinced about the utilty of the assertions on the series of calls to storage, but they may help a lot when fixing potential bugs.

tests/core/test_scripts.py
```diff
@@ -119,3 +122,183 @@ def test_delete_collection_raise_if_the_collection_does_not_exist(self):
...
+            pagination_rules=[[Filter('last_modified', 100, COMPARISON.GT)]],
+            sorting=[self.OLDEST_FIRST],
+            )
+        self.registry.storage.get_all.assert_any_call(
+            collection_id='bucket',
+            limit=self.BATCH_SIZE,
+            parent_id='',
+            pagination_rules=[[Filter('last_modified', 10, COMPARISON.GT)]],
+            sorting=[self.OLDEST_FIRST],
+            )
```
leplatrem: Maybe in this one we can just assert about the output and not individual calls to `get_all()` ?

kinto/__main__.py
Natim: I don't think this would be possible, because to know that you would need to read the configuration file which is already a parameter of the CLI tool...

kinto/__main__.py
```diff
@@ -143,6 +153,11 @@ def main(args=None):
                                          parsed_args['bucket'],
                                          parsed_args['collection'])
 
+    elif which_command == 'rebuild-quotas':
+        dry_run = parsed_args['dry_run']
+        env = bootstrap(config_file)
+        return scripts.rebuild_quotas(env, dry_run=dry_run)
```
Natim: I would be ok to move that function to a scripts module in the quota plugin.

kinto/core/scripts.py
```diff
@@ -88,3 +91,110 @@ def delete_collection(env, bucket_id, collection_id):
...
+    https://github.com/Kinto/kinto/issues/1226.
+    """
+    registry = env['registry']
+    settings = registry.settings
+    readonly_mode = asbool(settings.get('readonly', False))
+
+    if readonly_mode:
+        # FIXME: there is the possibility of race conditions if under
+        # normal operation while the rebuild operation
+        # happens. Shouldn't we *enforce* readonly mode?
```
Natim: If you are in readonly mode you cannot write on the database, even from a script.
It is not the responsibility of this script to turn users in maintenance mode, if need is, it is the OPS that will do it.

kinto/core/scripts.py
```diff
@@ -88,3 +91,110 @@ def delete_collection(env, bucket_id, collection_id):
...
+        message = ('Cannot rebuild quotas while in readonly mode.')
+        logger.error(message)
+        return 31
+
+    bucket_pagination = None
+    while True:
```
Natim: I am a bit puzzle by a while True here, can we turn thoose into `for bucket in buckets` instead?

kinto/core/scripts.py
glasserc: I just followed the example of the `delete-collection` script, although I feel like these scripts don't need to obey the same rules about readonly mode.

kinto/core/scripts.py
glasserc: This is the record that will be inserted into the DB so I figured I would just return that. But it's really just because I was being lazy.

kinto/core/scripts.py
glasserc: I have a two-level structure because I'm trying to handle pagination in the correct/natural way. But it's really clumsy. Maybe the right thing to do would be to write a "paginated" generator, but maybe that's overkill for this script?

leplatrem: generator would be nicer indeed. I don't think it's overkill in put into small helpers

kinto/__main__.py
glasserc: I can add a check to the script to see if it's in `includes`, how does that sound?

kinto/core/scripts.py
glasserc: I decided that the FIXME wasn't necessary and instead filed https://github.com/Kinto/kinto/issues/1236.

kinto/__main__.py
glasserc: I decided to keep the wrapper (that checks config parameters and returns exit status codes) in scripts, but move the body of the script into the quota plugin.

glasserc:
This should be ready for legit review now.

docs/commandline.rst
```diff
@@ -110,3 +110,30 @@ For example:
...
     This command does not go through the HTTP API and won't trigger
     :class:`kinto.core.events.ResourceChanged` events.
+
+Rebuild quotas
+--------------
+
+Recalculate the amount of storage taken up by buckets and collections
+and update quota records to match reality. This can be useful if
+you've been bitten by
+https://github.com/Kinto/kinto/issues/1226. However, this isn't
```
leplatrem: nit: `bug #1226 <https://github.com/Kinto/kinto/issues/1226>`_

kinto/__main__.py
leplatrem: If that's possible it would be cool, but not crucial

PR #1237, opened 2017-05-25
make memory storage consistent with pg wrt bytes

elelay:
Following #1227, make memory storage consistent with PostgreSQL storage:
 - bytes in records are interpreted as utf-8 encoded.
 - one always gets str back, not bytes.

Fixes #

- [ ] Add documentation.
- [X] Add tests.
- [ ] Add a changelog entry.
- [X] no HTTP API change

r? @glasserc @Natim

### DIFF
--- a/kinto/core/storage/testing.py
+++ b/kinto/core/storage/testing.py
@@ -1250,7 +1250,7 @@ class ParentRecordAccessTest:
         self.assertEqual(back_record['steak'], 'haché')
 
     def test_create_bytes_value_bad_encoding_raises(self):
-        self.assertRaises(Exception,
+        self.assertRaises(OverflowError,
                           self.create_record,
                           {'steak': 'haché'.encode(encoding='iso-8859-1')}
                           )
@@ -1274,7 +1274,7 @@ class ParentRecordAccessTest:
         record = self.create_record()
 
         new_record = {'steak': 'haché'.encode(encoding='iso-8859-1')}
-        self.assertRaises(Exception,
+        self.assertRaises(OverflowError,
                           self.storage.update,
                           object_id=record['id'],
                           record=new_record,
--- a/kinto/core/storage/testing.py
+++ b/kinto/core/storage/testing.py
@@ -1238,7 +1238,7 @@ class ParentRecordAccessTest:
                                        **self.storage_kw)
         self.assertNotIn("another", not_updated)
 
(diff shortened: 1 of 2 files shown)


kinto/core/storage/testing.py
```diff
@@ -1238,6 +1238,22 @@ def test_parent_cannot_update_other_parent_record(self):
...
 
+    def test_bytes_value_gets_back_str(self):
+        data = {'steak': 'haché'.encode(encoding='utf-8')}
+        self.assertIsInstance(data['steak'], bytes)
+
+        record = self.create_record(data)
+
+        back_record = self.storage.get(object_id=record['id'],
+                                       **self.storage_kw)
+        self.assertIsInstance(back_record['steak'], str)
```
glasserc: Maybe assert that it's still equal to `'haché'`, and not e.g. `'hachÃ©'`?

kinto/core/storage/testing.py
```diff
@@ -1238,6 +1238,22 @@ def test_parent_cannot_update_other_parent_record(self):
...
+
+    def test_bytes_value_bad_encoding_raises(self):
+        self.assertRaises(Exception,
```
glasserc: Can we specify a type of exception?

Natim: UnicodeDecodeError maybe?

kinto/core/storage/testing.py
elelay: yes!

kinto/core/storage/testing.py
elelay: UnicodeDecodeError would make sense, but it's [OverflowError](https://github.com/esnme/ultrajson/blob/master/python/objToJSON.c#L834) for any python to json error, actually. It seems rather specific to ujson. Do you want that in the tests?

Natim: I guess yes.

glasserc: Can we `self.assertRaises((OverflowError, UnicodeDecodeError), ...)`? Per https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaises, it seems like we can.

glasserc:
This is clearly an improvement, I'm just going to merge as-is. Thanks again for your hard work!

elelay:
Great, thanks!

PR #1246, opened 2017-05-31
Prepare 7.1.0

leplatrem:


Natim:
Can we rebase with #1249 ?

leplatrem:
Tests fail because of https://github.com/pypa/setuptools/issues/1042

PR #1249, opened 2017-05-31
Fix bucket cascade deletion

leplatrem:
This one is pretty heavy (deleting `/buckets/a` would also wipe out `/buckets/a-staging`).


Natim:
> how did you find it?

I did, when I removed the fennec bucket in stage and that fennec-preview and fennec-staging disappeared.

Issue #1216, opened 2017-05-05
Filter objects where a particular field contains an empty list


leplatrem:
`/records?field.name=[]` ?

leplatrem:
Related #343 #344

glasserc:
Fixed by #1252 with tests added in #1258.

PR #1220, opened 2017-05-15
WIP experiments around safe casts in postgresql backend (ref #1217)

leplatrem:
Fixes #1217

Issue #1215, opened 2017-05-05
Filter objects where a particular field is null 


leplatrem:
`/records?field.name=null` ?

leplatrem:
Related #343 #344

glasserc:
Isn't that syntax ambiguous? It's the same to search for any object whose value is the string `"null"`. I'd rather be more rigid about what our query syntax is for this case, maybe introducing a new syntax for JSON-y values like `is_field.name=null` or `json_field.name=null`.

Natim:
``https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records?build.id=null``

Returns an empty list, so I guess it is not sufficient.

glasserc:
Should `field.name=null` match records where the field is missing, or only where it is explicitly present and null? (I would vote for the second option)

glasserc:
Fixed by #1252 and #1258.

PR #1252, opened 2017-06-05
Handle querystring parameters as JSON encoded values. Fixes #1217

### LINKED ISSUE #1217: Crash on dev ota server when requesting /records?min_target.version=53
```
$ http GET https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/archives/records\?min_target.version\=53`

HTTP/1.1 503 Service Unavailable
Access-Control-Expose-Headers: Total-Records, Next-Page, Expires, Last-Modified, ETag, Content-Length, Pragma, Alert, Cache-Control, Retry-After, Backoff
Connection: keep-alive
Content-Length: 151
Date: Fri, 05 May 2017 15:13:28 GMT
Retry-After: 3
Server: nginx

{
    "code": 503, 
    "errno": 201, 
    "error": "Service Unavailable", 
    "message": "Service temporary unavailable due to overloading or maintenance, please retry later."
}
```

Natim:
Fixes #1217 

- [X] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,10 @@ This document describes changes between each past release.
 7.2.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
+**Bug fixes**
+
+- Handle querystring parameters as JSON encoded values
+  to avoid treating number as number where they should be strings. (#1217)
 
 
 7.1.0 (2017-05-31)
--- a/tests/core/test_utils.py
+++ b/tests/core/test_utils.py
@@ -33,8 +33,23 @@ class NativeValueTest(unittest.TestCase):
     def test_simple_string(self):
         self.assertEqual(native_value('value'), 'value')
 
+    def test_defined_string(self):
+        self.assertEqual(native_value('"value"'), 'value')
+
+    def test_null_value(self):
+        self.assertEqual(native_value('null'), None)
+
+    def test_defined_null_as_text_value(self):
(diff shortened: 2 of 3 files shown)


tests/core/test_utils.py
```diff
@@ -44,12 +59,12 @@ def test_float(self):
         self.assertEqual(native_value('3.14'), 3.14)
 
     def test_true_values(self):
-        true_strings = ['True', 'on', 'true', 'yes']
```
magopian: Why change this?

Natim: Because they are not valid JSON values and we don't use it, so I'd like to change the behavior to stick with the JSON spec there.

Natim:
It doesn't fix #1217 but it is a first step toward. I think the Postgresql issue will be handled by #1220

glasserc:
I believe this fixes #1216 and #1215.

Natim:
I am not sure we should add specific tests to make sure it works I guess.

CHANGELOG.rst
```diff
@@ -6,7 +6,10 @@ This document describes changes between each past release.
 7.2.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
+**Bug fixes**
+
+- Handle querystring parameters as JSON encoded values
+  to avoid treating number as number where they should be strings. (#1217)
```
leplatrem: This could be an API breaking change (although `off` and `no` were not documented...)

PR #1253, opened 2017-06-05
Add tests for the Next-Page with optional fields feature.

### LINKED ISSUE #816: Next-Page with filtering on a not mandatory field raises a KeyError.
Reproductible easily with the Kinto-Admin. Let me write a test for it.

Natim:
Fixes #816
Superseed #816

- [X] Add tests.
- [x] Add a changelog entry.


Natim:
It was for the memory backend but I don't know if my test are running for the postgresql one.

Natim:
No because we are then making sure last_modified value are different. Also the tests are validating that we don't see twice the same record.

PR #1254, opened 2017-06-05
Add startswith and endswith filters.

Natim:
Adding startswith and endswith filters

- [x] Add documentation.
- [X] Add tests.
- [x] Add a changelog entry.

r? @n1k0

### DIFF
--- a/docs/api/1.x/filtering.rst
+++ b/docs/api/1.x/filtering.rst
@@ -60,6 +60,18 @@ Prefix field name with ``like_``:
 
 * ``/collection?like_field=foo``
 
+**Search string fields with starting string**
+
+Prefix field name with ``startswith_``:
+
+* ``/collection?startswith_field=foo``
+
+**Search string fields with ending string**
+
+Prefix field name with ``endswith_``:
+
+* ``/collection?endswith_field=foo``
+
 .. note::
 
     Will return an error if a field is unknown.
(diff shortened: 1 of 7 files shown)


docs/api/1.x/filtering.rst
```diff
@@ -60,6 +60,18 @@ Prefix field name with ``like_``:
...
+Prefix field name with ``startswith_``:
+
+* ``/collection?startswith_field=foo``
+
+**Search string fields with ending string**
+
+Prefix field name with ``endswith_``:
+
+* ``/collection?endswith_field=foo``
+
```
leplatrem: Why not `/like_field=*foo*` ?

For example, if like operator contains `*` then respect it, otherwise do `*<keyword>*` like currently.

kinto/core/storage/testing.py
```diff
@@ -525,6 +525,24 @@ def test_get_all_handle_all_pagination_rules(self):
...
+
+    def test_records_filtered_when_searched_for_endswith_on_string_field(self):
+        self.create_record({'name': 'barfoo'})
+        self.create_record({'name': 'foobar'})
+        self.create_record({'name': 'FOOBAR'})
+
+        filters = [Filter('name', 'BaR', utils.COMPARISON.ENDSWITH)]
+        results, count = self.storage.get_all(filters=filters, **self.storage_kw)
+        self.assertEqual(len(results), 2)
+
```
leplatrem: As @n1k0  pointed out, we should add tests when stored objects contain integers, lists etc. We've seen numerous bugs (in PG) when relying implicit casts...

docs/api/1.x/filtering.rst
Natim: Yes that's what @n1k0 suggested, so it would allow to do `*a*b*c` requests for instance.

PR #1255, opened 2017-06-05
use json instead of ujson in tests

### LINKED ISSUE #1238: specify record contents accepted by storage
An inconsistency between memory and PG storage led to shipping broken `accounts` plugin (#1224).

It was caused by memory storage not touching the contents of given record (contained a bytes value), whereas PG storage converted it to str. See discussion in #1227.

Memory can be made like PG storage with a moderate cost (serialiaze/deserialize records with ujson), like in PR #1237. This leaves potential errors in production when somebody sends non utf-8 encoded input where we only tested with utf-8 input.

But, as @glasserc pointed out, storage could also be made stricter on datatypes it accepts in records (i.e. REJECT bytes). Then errors will happen during development, which is better.

To do this, I think one would have to scan records beforehand or add an option to ujson to do it or retuirn to standard package json.

elelay:
Fixes #1238

- [X] Add documentation.
- [X] Add tests.
- [x] Add a changelog entry.
- [X] not changed the HTTP API

r? @glasserc @Natim


tests/plugins/test_history.py
```diff
@@ -17,7 +17,8 @@ class PluginSetup(unittest.TestCase):
     def test_a_statsd_timer_is_used_for_history_if_configured(self):
         settings = {
             "statsd_url": "udp://127.0.0.1:8125",
-            "includes": "kinto.plugins.history"
+            "includes": "kinto.plugins.history",
+            "storage_strict_json": True
```
leplatrem: I don't think it's necessary to change this for the history plugin tests?

docs/configuration/settings.rst
```diff
@@ -122,6 +122,8 @@ Storage
 +------------------------------+-------------------------------+--------------------------------------------------------------------------+
 | kinto.storage_max_backlog    | ``-1``                        | Number of threads that can be in the queue waiting for a connection.     |
 +------------------------------+-------------------------------+--------------------------------------------------------------------------+
+| kinto.storage_strict_json    | ``False``                     | (test only) validate that records are serializable using the json module.|
++------------------------------+-------------------------------+--------------------------------------------------------------------------+
```
leplatrem: If it is meant to be used in tests only I don't think it's necessary to document it here. No?

Natim:
> Are we still sure that ujson brings us enough benefit compared to python 3.6 json native module? Maybe we could drop it completely.

Kinto speed is due to the use of ujson. We should definitely keep it.

tests/plugins/test_history.py
Natim: Why not?

Natim:
> I haven't followed the whole conversation about this

The rationale is that PostgreSQL backend is using ujson while the memory backend was not. We discovered that sometime ujson doesn't behave exactly as the json module.

@elelay upgraded the memory backend to use ujson so that both backend behave the same way with regards to JSON handling.

To be honest I would just use ujson everywhere without adding a new setting.

leplatrem:
> Kinto speed is due to the use of ujson. We should definitely keep it.

Do we have numbers? We did benchmarks 2 years ago with Python 2.7. I continuously read about python 3.6 super performances. We should probably confirm that with updated figures.

> To be honest I would just use ujson everywhere without adding a new setting.

+1

Natim:
> We should probably confirm that with updated figures.

Sure if you feel like doing so. 

> I continuously read about python 3.6 super performances.

We are not using Python 3.6 yet in production as far as I know. But I really doubt JSON handling in Python could be faster than in libc.

glasserc:
The problem is that the behavior of ujson is arguably broken and lets a developer accidentally serialize things that shouldn't be serializable. Any arbitrary object will be serialized by using its `__dict__`. But what I'm most worried about is that we would serialize a bytestring that seems OK until by chance it contains some non-UTF-8 encoded text and then failing.

Personally I don't think this is too much complexity to avoid those kinds of bugs. (The PR only changes ~75ish lines of code.) The only thing that I might consider is trying to localize the complexity to the `utils` module.

elelay:
@leplatrem 
> We should probably confirm that with updated figures.

I see an up to 5x speed improvement (decode 256 UTF8 strings) for ujson/json, with an average of 2x for encode and decode.


|                                                                               | ujson      | yajl       | simplejson | json       |
|-------------------------------------------------------------------------------|------------|------------|------------|------------|
| **Array with 256 doubles**                                                        |            |            |            |            |
| encode                                                                        |   22485.14 |    8507.37 |    5752.49 |    6314.80 |
| decode                                                                        |   35739.87 |   15104.25 |   14935.88 |   15758.40 |
| **Array with 256 UTF-8 strings**                                                  |            |            |            |            |
| encode                                                                        |    3941.41 |    3788.01 |    2562.49 |    2573.70 |
| decode                                                                        |    2675.51 |     918.95 |     536.55 |     522.48 |
| **Array with 256 strings**                                                        |            |            |            |            |
| encode                                                                        |   51644.13 |   21904.67 |   23306.93 |   25884.61 |
| decode                                                                        |   33265.51 |   25833.36 |   40359.65 |   29960.73 |
| **Medium complex object**                                                         |            |            |            |            |
| encode                                                                        |   15259.40 |    8426.39 |    5482.92 |    7910.24 |
| decode                                                                        |   14702.96 |    8594.63 |    8479.10 |   10029.45 |
| **Array with 256 True values**                                                    |            |            |            |            |
| encode                                                                        |  159535.24 |  172716.43 |   78707.41 |   96372.96 |
| decode                                                                        |  234235.84 |  109203.62 |  142965.52 |  168322.68 |
| **Array with 256 dict{string, int} pairs**                                        |            |            |            |            |
| encode                                                                        |   20273.73 |   15388.58 |    4463.77 |    9060.82 |
| decode                                                                        |   17670.78 |   11974.63 |    9352.05 |   12167.52 |
| **Dict with 256 arrays with 256 dict{string, int} pairs**                         |            |            |            |            |
| encode                                                                        |      76.93 |      56.22 |      15.78 |      33.29 |
| decode                                                                        |      44.10 |      34.39 |      27.66 |      33.44 |
| **Complex object**                                                                |            |            |            |            |
| encode                                                                        |     644.05 |            |     554.48 |     586.59 |
| decode                                                                        |     591.66 |     307.53 |     226.03 |     219.96 |

elelay:
> @glasserc  But what I'm most worried about is that we would serialize a bytestring that seems OK until by chance it contains some non-UTF-8 encoded text and then failing.

+1 this is why I took the pain to implement this PR.

Now, choosing between ujson and json could maybe be implemented more elegantly. Is it possible to mock an attribute in any future instance of a class or sthing? Then we could switch the json attribute of StorageBase before tests and not have to explicitly configure it.

elelay:
Meanwhile, I'm modifying ujson to add a 'reject_bytes' parameter.
If it's merged, it would allow us to always use ujson, with this parameter always True.

glasserc:
Refs https://github.com/esnme/ultrajson/pull/266.

I tried "simplifying" this by centralizing the logic to dispatch between JSON flavors in `utils.json`, but that didn't work because there's nothing to dispatch according to. (This isn't `flask`; I don't have global access to the request or the config.) I think it would be possible to implement @elelay's suggestion of mocking/overriding the `utils.json` flavor using a monkeypatch, but that seems much uglier to me than just configuring it explicitly (as has been done here).

@leplatrem @Natim Would you veto if I wanted to land it? I see a +1 for using ujson on everything, but no strong objection against the patch as it stands now. Do you have an alternative implementation in mind that would be cleaner?

This PR needs a rebase in the meantime, plus there are a few remaining nits from the review.

Natim:
I am not confident handling the conflicts myself.

elelay:
Thanks @glasserc. I think all nits are addressed.

glasserc:
Thanks for your work and for bearing with us while we sorted this out!

Issue #1257, opened 2017-06-06
Handling missing fields


glasserc:
If you try to filter for fields whose value is the empty string, it also matches records without those fields:

```
> http 'https://kinto.dev.mozaws.net/v1/buckets/46524be8-0ad7-3ac6-e260-71f8993feffa/collections/tasks/records?any_arbitrary_field=""' -a user:pass
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Next-Page, Pragma, Retry-After, Backoff, Alert, Cache-Control, ETag, Last-Modified, Expires, Total-Records, Content-Length
Cache-Control: no-cache, no-store
Connection: keep-alive
Content-Length: 429
Content-Type: application/json
Date: Tue, 06 Jun 2017 20:50:47 GMT
ETag: "1496685126658"
Last-Modified: Mon, 05 Jun 2017 17:52:06 GMT
Server: nginx
Total-Records: 4
X-Content-Type-Options: nosniff

{
    "data": [
        {
            "done": false,
            "id": "3dd095cb-f4f7-4c87-8112-6cd3e7660f41",
            "last_modified": 1496685088616,
            "title": "test"
        },
        {
            "done": false,
            "id": "0717608f-1cf2-472b-8a49-3136d2811946",
            "last_modified": 1496685088608,
            "title": "blbl"
        },
        {
            "done": false,
            "id": "65e5b714-dcb2-4136-836c-996e83514297",
            "last_modified": 1496524477816,
            "title": "to a poo"
        },
        {
            "done": false,
            "id": "5a1fc9a7-6c6a-428e-b932-930a69d6495a",
            "last_modified": 1496422987261,
            "title": "444"
        }
    ]
}
```

This is because of https://github.com/Kinto/kinto/blob/master/kinto/core/storage/postgresql/__init__.py#L709. Postgres appears to be the reference implementation of Kinto's storage backend (see #815), so you might expect this behavior to leak into other backends too.

The [documentation](https://github.com/Kinto/kinto/blob/master/docs/api/1.x/filtering.rst) is pretty scanty on the subject of missing fields, saying only `Will return an error if a field is unknown`, and I think that's meant to only apply to `like_`.

[...]

Issue #1004, opened 2016-12-23
Should we make querystrings reserved words?


gabisurita:
I think maybe we shouldn't to keep a simple schema, but we should do a big warning about this in the docs. We can do some crazy things with them:

```json
$ echo '{"data": {"_limit": 5}}' | http put localhost:8888/v1
{
    "data": {
        "_limit": 5,
        "id": "b1",
        "last_modified": 1482506056818
    },
    "permissions": {
        "write": [
            "basicauth:902db65276f24f6f084cd57fddb1bb3835054d27858a606f59410e8c227465e2"
        ]
    }
}

$ http get localhost:8888/v1/buckets?_limit=1 -a a:a
{
    "data": [
        {
            "_limit": 5,
            "id": "b1",
            "last_modified": 1482506056818
        }
    ]
}
```

```json
$ echo '{"data": {"max_field": "10", "field": 5}}' | http put localhost:8888/v1/buckets/b1 -a a:a
{
    "data": {
        "field": 5,
        "id": "b1",
        "last_modified": 1482505784675,
        "max_field": "10"
    },
    "permissions": {
        "write": [
            "basicauth:902db65276f24f6f084cd57fddb1bb3835054d27858a606f59410e8c227465e2"
        ]
    }
}

$ http get localhost:8888/v1/buckets?max_field=10 -a a:a
{
    "data": [
        {
            "field": 5,
            "id": "b1",
            "last_modified": 1482505784675,
            "max_field": "10"
        }
    ]
}
```

leplatrem:
:) 

Indeed... I wouldn't be against raising a 400 when a record uses a reserved word, but that may be a bit radical I don't know!

gabisurita:
Maybe we could send just a warning. Can we use the Alert header for this?

Natim:
I like your suggestion about the Alert header because we don't break any API.
However I am not against raising a 400 because the API starts to be broken with an undefined behavior as soon as people start to use _sort and _limit in their records.

Issue #344, opened 2015-12-16
How to search for records having or not having a property?


Natim:
Let say I have [some records](https://kinto-ota.dev.mozaws.net/v1/buckets/mozlando/collections/wall/records).

**How can I select only records having a text attribute?**

https://kinto-ota.dev.mozaws.net/v1/buckets/mozlando/collections/wall/records?having_text=true ?
https://kinto-ota.dev.mozaws.net/v1/buckets/mozlando/collections/wall/records?having_text=false ?

almet:
contains?

Natim:
Contains c'est pas top surtout si on l'utilise pour #343

almet:
defined, exists?

Natim:
I like defined :+1: 

- ``text_defined=true``
- ``text_defined=false``

leplatrem:
* `?has_attr=true`
* `?has_attr=false`

gabisurita:
It's a bit unfriendly, but this is possible via:

- `?attr`  for records not having the attribute
- `?not_attr` for records having the attribute defined

For example:

```
$ http GET localhost:8888/v1/buckets/default/collections/c1/records --auth aaa:aaa
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
Cache-Control: no-cache, no-store
Content-Length: 166
Content-Type: application/json; charset=UTF-8
Date: Mon, 10 Oct 2016 22:20:16 GMT
Etag: "1476137810045"
Last-Modified: Mon, 10 Oct 2016 22:16:50 GMT
Server: waitress
Total-Records: 3

{
    "data": [
        {
            "id": "r3",
            "last_modified": 1476137810045,
            "param": "nope"
        },
        {
            "id": "r2",
            "last_modified": 1476137797249
        },
        {
            "id": "r1",
            "last_modified": 1476137764176,
            "param": "yeap"
        }
    ]
}

$ http GET localhost:8888/v1/buckets/default/collections/c1/records?not_param --auth aaa:aaa
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
Cache-Control: no-cache, no-store
Content-Length: 124
[...]

PR #1258, opened 2017-06-07
Allow JSON filtering and clarify sort order

### LINKED ISSUE #1215: Filter objects where a particular field is null

glasserc:
Fixes #1215, #1216, #1257, and possibly others.

Instead of coercing all values to strings when doing comparisons in Postgres, let Postgres use its own JSONB type comparison. This gets us 90% of the way to supporting JSON comparisons, including comparisons against null. The remaining part is about handling "missing" fields. I've made the executive decision that a "missing" value is not equal to any value of any type, including the JSON null. (Fortunately, this is easy in Postgres.) We handle these explicitly by adding `field IS NOT NULL AND` or `field IS NULL OR` as needed.

Because we already support sorting on a field with mixed-type values and/or missing values, we already implicitly have an ordering defined for comparisons like LT, GT, MAX, MIN, etc. This ordering puts missing values as "bigger" than other values. While we're messing around with comparisons involving NULL, fix this too.

We don't appear to have any clear recommendation about our ordering in the memory backend. To some extent, the tests assume that the ordering is the same. However, I am skeptical that we can always harmonize on the Postgres ordering. I think a more sensible approach is to allow each backend to define its own sort order, as long as it's consistent. For this reason, I didn't struggle with making the memory backend comparison function exactly the same as the Postgres one does.

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem, @Natim 

I'm pretty confident in the approach, but I'm not sure whether this qualifies as a major version and/or API version.

### DIFF
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -374,16 +374,19 @@ def schwartz_transform(value):
         return (0, value)
     if isinstance(value, str):
         return (1, value)
+    if isinstance(value, bool):
+        # This has to be before Number, because bools are a subclass
+        # of int :(
+        return (3, value)
     if isinstance(value, numbers.Number):
         return (2, value)
-    if isinstance(value, bool):
-        return (3, value)
     if isinstance(value, abc.Sequence):
         return (4, value)
     if isinstance(value, abc.Mapping):
         return (5, value)
-    if value == MISSING:
+    if value is MISSING:
         return (6, value)
+    raise ValueError("Unknown value: {}".format(value))   # pragma: no cover
 
 
 def apply_sorting(records, sorting):
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -346,8 +346,8 @@ def apply_filters(records, filters):
             if f.operator in (COMPARISON.IN, COMPARISON.EXCLUDE):
                 right, left = left, right
             elif f.operator not in (COMPARISON.LIKE, COMPARISON.HAS):
-                left = schwartz_transform(left)
-                right = schwartz_transform(right)
+                left = schwartzian_transform(left)
+                right = schwartzian_transform(right)
 
             if f.operator == COMPARISON.HAS:
                 matches = left != MISSING if f.value else left == MISSING
@@ -357,7 +357,7 @@ def apply_filters(records, filters):
             yield record
 
 
-def schwartz_transform(value):
+def schwartzian_transform(value):
     """Decorate a value with a tag that enforces the Postgres sort order.
 
     The sort order, per https://www.postgresql.org/docs/9.6/static/datatype-json.html, is:
@@ -398,7 +398,7 @@ def apply_sorting(records, sorting):
         return result
 
     def column(record, name):
-        return schwartz_transform(find_nested_value(record, name, default=MISSING))
+        return schwartzian_transform(find_nested_value(record, name, default=MISSING))
 
     for sort in reversed(sorting):
         result = sorted(result,
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -1,6 +1,8 @@
 import re
 import operator
 from collections import defaultdict
+from collections import abc
+import numbers
 
 from kinto.core import utils
 from kinto.core.decorators import synchronized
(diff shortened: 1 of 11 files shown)


glasserc:
I've added some commits to implement a `has_` operator, addressing #344 directly, since the change in how we handle missing fields breaks a test each in `test_views_groups` and `test_views_collections`. I know this is kind of outside the scope of this PR, but it touches a lot of the same code and benefits from the new cleaner implementation. If you'd rather, I can separate out the commits for that feature.

kinto/core/storage/memory.py
```diff
@@ -332,26 +341,54 @@ def apply_filters(records, filters):
...
-                        continue
-                    else:
-                        left = ''  # To mimic what we do for postgresql.
-            matches = matches and operators[f.operator](left, right)
+                matches = matches and operators[f.operator](left, right)
         if matches:
             yield record
 
 
+def schwartz_transform(value):
```
leplatrem: cool, did not know about this name :) 

btw, on wikipedia it's called schwartzian transform

kinto/core/storage/postgresql/__init__.py
```diff
@@ -686,6 +686,7 @@ def _format_conditions(self, filters, id_field, modified_field,
         holders = {}
         for i, filtr in enumerate(filters):
             value = filtr.value
+            query_is_like = filtr.operator == COMPARISON.LIKE
```
leplatrem: **µnit**:  boolean values usually start with `is_` or `has_`... what about `is_like_operator` ?

kinto/core/storage/postgresql/__init__.py
```diff
@@ -701,43 +702,72 @@ def _format_conditions(self, filters, id_field, modified_field,
...
+                # everything, including JSON null. Do this on a
+                # per-operator basis.
+                if filtr.operator in (
+                        # NULLs aren't EQ to anything (definitionally).
+                        COMPARISON.EQ,
+                        # So they can't match anything in an INCLUDE.
+                        COMPARISON.IN,
+                        # Nor can they be LIKE anything.
+                        COMPARISON.LIKE,
+                ):
```
leplatrem: **µnit**: in python multi-lines `if` are not recommended. Would it be better to have two intermediary variables like `null_comparable_ops` and `null_incomparable_ops` ?

kinto/core/storage/testing.py
```diff
@@ -350,6 +350,50 @@ def test_get_all_handle_sorting_on_subobject(self):
...
+    def test_get_all_sorting_is_consistent_with_filtering(self):
+        self.create_record({'flavor': 'strawberry'})
+        self.create_record({'flavor': 'blueberry', 'author': None})
+        self.create_record({'flavor': 'raspberry', 'author': 1})
+        self.create_record({'flavor': 'orange', 'author': True})
+        self.create_record({'flavor': 'watermelon', 'author': 'Ethan'})
+        sorting = [Sort('author', 1)]
+        records, _ = self.storage.get_all(sorting=sorting, **self.storage_kw)
+        # Some interesting values to compare against
+        VALUES = ['A', 'Z', '', 0, 4]
```
leplatrem: µnit: I don't think it's necessary to have uppercase here

leplatrem:
> I'm not sure whether this qualifies as a major version and/or API version.

The way I see it:

* You fixed inconsistencies and bugs
* Those fixes deserve a mention in the API changelog, so we should bump its version
* I don't think it requires a major dump for the API version ( #1252 didn't :) )
* It doesn't break anything on the Python side, so major dump in the package version either

glasserc:
I only have two concerns about HTTP API compatibility:

- We break the technique shown in https://github.com/Kinto/kinto/issues/344 (but that was never documented, so I'm not super worried about it)
- We implicitly discourage the use of fields with the name `has_foo`

leplatrem:
> We break the technique shown in #344 

With all due respect, it was very hacky.

> We implicitly discourage the use of fields with the name has_foo

Well, I agree, but acceptable IMO since that's also true for other fields as shown in #1004

Natim:
This is awesome thank you for tackling it!

PR #1259, opened 2017-06-08
Fix schema validation with internal fields (fixes #1244)

### LINKED ISSUE #1244: If id, schema or last_modified are marked as required, record can't be validated
Because we pop those fields before validating them with the collection schema.

We can either:
* prevent those fields to be mentioned as `required` when defining the collection schema
* or not pop them if they are present in the schema before validating

leplatrem:
Fixes #1244

### DIFF
--- a/kinto/views/records.py
+++ b/kinto/views/records.py
@@ -1,5 +1,3 @@
-import copy
-
 import jsonschema
 from kinto.core import resource, utils
 from kinto.core.errors import raise_invalid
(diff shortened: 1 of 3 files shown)


kinto/views/records.py
```diff
@@ -55,23 +53,32 @@ def process_record(self, new, old=None):
...
+        # Remove internal and auto-assigned fields from schema and record.
+        internal_fields = (self.model.id_field,
+                           self.model.modified_field,
+                           self.schema_field,
+                           self.model.permissions_field)
+        required_fields = [f for f in schema.get('required', []) if f not in internal_fields]
+        if required_fields:
+            schema = {**schema, 'required': required_fields}
+        else:
+            schema = {f: v for f, v in new.items() if f != 'required'}
```
glasserc: I don't think we need to go to this lengths to avoid putting an empty `required` field. The [JSON Schema Spec](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.17) says that omitting it is the same as an empty array. Unless you're trying to do this so that we don't confuse users by secretly changing their schema, but I don't think that's valuable, since we're already changing the schema.

Should we be doing these manipulations to the schema at schema save time?

Should we be removing definitions of internal fields from `schema['properties']`?

kinto/views/records.py
```diff
@@ -55,23 +53,32 @@ def process_record(self, new, old=None):
...
+        data = {f: v for f, v in new.items() if f not in internal_fields}
```
glasserc: :+1:

kinto/views/records.py
Natim: I agree with @glasserc that we should do all this at save time, rejecting the schema or changing it silently if it has internal fields in it.

leplatrem: This approach has some advantages:
* it avoids breaking with existing collection metadata
* we strip record and schema at the same place (stripping record fields will still be necessary if the users don't allow additional props on their schemas)
* we are in a resource class, where `self.model.id_field` etc. are available. It's not the case within the colander validator for collection schema
* it is fairly similar to what there were before this PR, so I'm not chocked by those changes

Please expose your whys :)

leplatrem: > I don't think we need to go to this lengths to avoid putting an empty required field

The empty array gives me an error in jsonschema library that we use: 

```
Failed validating 'minItems' in schema['properties']['required']:
    {'items': {'type': 'string'},
     'minItems': 1,
     'type': 'array',
     'uniqueItems': True}

On instance['required']:
    []
```

leplatrem:
Please open an issue if you think that we should refactor this.

PR #1260, opened 2017-06-08
Fix requests output when running with make serve (fixes #1242)

### LINKED ISSUE #1242: Bad default logging configuration?
This is the output I've got on batch requests with the default logging configuration (generated by `kinto init`).

screenshot from 2017-05-29 12-52-19

We are supposed to see subrequests and probably not the connection pool information.

leplatrem:
Fixes #1242


leplatrem:
It's already the case (`pyramid > 1.8`) :)

PR #1261, opened 2017-06-09
Add support of startswith/endswith to like operator

leplatrem:
Supersedes #1254

### DIFF
--- a/docs/api/1.x/filtering.rst
+++ b/docs/api/1.x/filtering.rst
@@ -17,7 +17,8 @@ name. Values can be any JSON encoded value (e.g. ``24``, ``"hello"``,
 ``null``). Anything not recognized as a JSON value is interpreted as a
 string.
 
-**Single value**
+Single value
+------------
 
 * ``/collection?field=value``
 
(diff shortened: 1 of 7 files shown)


docs/api/1.x/filtering.rst
```diff
@@ -76,48 +83,59 @@ types is not defined. For example, if you have a record like
...
 Prefix field name with ``like_``:
 
 * ``/collection?like_field=foo``
 
-**Field existence**
+The specified value can also contain wildchars:
+
+* ``/collection?like_field=foo*`` (starts with ``foo``)
+* ``/collection?like_field=*foo`` (ends with ``foo``)
+* ``/collection?like_field=*foo*`` (equivalent to ``like_field=foo``)
```
n1k0: Except if I've miread the patch, I'd probably add a warning about the fact that now searching for a field string actually containing the `*` is now impossible.

docs/api/index.rst
```diff
@@ -11,7 +11,16 @@ API
...
 
-1.16 (unreleased)
+1.17 (unreleased)
+'''''''''''''''''
+
+- Filtering with like can now contain wild chars (eg. ``?like_nobody=*you*``)
+  It is thus now impossible to search for the ``*`` character with this operator.
+- Handle querystring parameters as JSON encoded values
+  to avoid treating number as number where they should be strings. (#1217)
+- Introduce ``has_`` filter operator (fixes #344).
```
glasserc: Oops, I didn't realize we had released a 1.16. You should probably remove the `has` and JSON-encoded notes from 1.16 (below).

Issue #292, opened 2015-11-20
Run multiple commands : kinto init migrate start


leplatrem:
Would that make sense ?

Or would it make sense to run `init` and `migrate` on `start` only when necessary ?

conlini:
It would make sense to tie in dependency chain on commands.

So running _migrate_ would ensure _init_ is run if not done previously
and running _start_ would run _init_ > _migrate_ and then _start_

That would reduce the need for multi step setup/startup

Natim:
We try to keep up with the *"Explicit is better than implicit"* [Python Zen philosophy](https://www.python.org/dev/peps/pep-0020/#content).

Doing the chain automatically will add complexity to detect cases.

For instance we might generate a new configuration file and run migration whereas the user actually forget to set its configuration file path or didn't run the kinto command from the proper repository.

Also letting the user run: ``kinto --ini config/kinto.ini init migrate start`` seems to make a lot of sense for me.

conlini:
hmm yes that does add a level of complexity and branches when there are multiple configurations.
Trying to work that out would require asking user to chose one of the configs, which would defeat the purpose of a single command starting up the server.

leplatrem:
Unlikely to happen (been here for too long)

PR #1262, opened 2017-06-09
Follow Dockerflow recommendations (fixes #998)

### LINKED ISSUE #998: Check compliance with Dockerflow spec
See https://github.com/mozilla-services/Dockerflow

* `$PORT` env variable
* Copy to `/app`
* `app` group:user
* `USER` directive
* `ENTRYPOINT` directive
* ...

leplatrem:
Fixes #998

Apparently we can't run two commands in one `CMD` instruction when using `ENTRYPOINT`.

Because #292 was never implemented, we can't have the `kinto` command as the entrypoint and `migrate` + `start` in the `CMD`.


leplatrem:
Ideally, if we could have multiple commands in `kinto` CLI, we should do:

```
ENTRYPOINT kinto
CMD migrate start --ini=$INI --port=$PORT
```

glasserc:
I'm not crazy about the multiple-commands thing. Parsing such a command line sounds kind of crazy (`--ini` has to apply to both `migrate` and `start`, but `--port` should only apply to `start`?). Why don't we define a special command migrate-and-start for this use case? Or maybe define a [setup.py entrypoint](https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/) in Python and call it `kinto-run-in-docker`?

mostlygeek:
Multiple commands is ok. The recommended way to do this is with a `run.sh`. You could use balrog's [run.sh](https://github.com/mozilla/balrog/blob/master/scripts/run.sh) as an example. 

Your `ENTRYPOINT` will be the run.sh. You don't need a `CMD` in your dockerfile. 

What's the rational behind doing `migrate start ...` with every command? Does that modify the schema of the underlying database?

glasserc:
Yes, `kinto migrate` updates the database schema to bring it up to date with the code that's running against it. This is especially important for the first run, which initializes the schema :)

Natim:
> This is especially important for the first run, which initializes the schema :)

Note that if no migration are necessary this will be really fast, it will also detect error while trying to access the DB.

mostlygeek:
Schema migration should be explicit, separated and not default. It is better to have migration and server startup as two discrete functions. Using `migrate-and-start` as a default is convenient during development but is too risky for production. 

For ops, code deployments and schema changes are usually discrete steps. Deployment are either: 

1. Update the schema, old code continues working. Deploy new code.
2. Deploy new code, continues working with old schema. Update the schema.

leplatrem:
Thanks for your feedback on this!

I'd like to recall that the `Dockerfile` of this repo is not the one that is used by ops at Mozilla (we use [a distribution of various plugins: kinto-dist](https://github.com/mozilla-services/kinto-dist/blob/master/Dockerfile)).

I personally find it super convenient to be able to do `docker run kinto` and get an instance running. But since we use the memory backend by default, we don't need `migrate` by default.

So we can put `kinto` as the entrypoint, and let the commands be `start`, `migrate` etc. as for the CLI.
That means if someone wants to run the container with a postgres database, it will be necessary to run migrate and start as separate commands. After giving it some thoughts, it makes sense.

We'll still probably need a `run.sh` file, since we'll want to set `--ini $KINTO_INI` as a default argument for every command.

PR #1263, opened 2017-06-09
Upgrade to Pyramid 2.X (fixes #1187)

### LINKED ISSUE #1187: Upgrade to Pyramid-tm 2.X
> 2.0
> The pyramid_tm tween has been moved over the EXCVIEW tween. This means the transaction is open during exception view execution. See https://github.com/Pylons/pyramid_tm/pull/55

http://docs.pylonsproject.org/projects/pyramid_tm/en/latest/#id2

Related #877 and #878

leplatrem:
Fixes #1187

**Do not merge** until we clearly understand (and comment) what happens :)


glasserc:
You found a good one. Some additional information:

- It changed between pyramid_tm 1.1.1 and pyramid_tm 2.0.
- The `transaction.abort()` call that you added does *not* fix anything in pyramid_tm 2.0 -- it only "works" in pyramid_tm 2.1.
- It seems like in all versions of pyramid_tm, the transaction is aborted once an exception is raised.

Is it maybe time that we stop registering `ResourceChanged` as a pre-commit hook and used it instead as a [commit veto hook](http://docs.pylonsproject.org/projects/pyramid_tm/en/latest/#adding-a-commit-veto-hook)?

Still digging..

glasserc:
Worked on this a bit more today. The behavior seems to be a bit different in pyramid 2.0 and pyramid 2.1. In pyramid 2.0, I see (from the logs) that the transaction is aborted, but then another transaction is created and that one commits. However, this isn't what fails the test. The test fails because the exception isn't turned into a failing 507 response. That bug gets fixed in 2.1 (I think it's https://github.com/Pylons/pyramid_tm/pull/61). However, now, the transaction doesn't get aborted; it doesn't commit as part of the response, but it does commit later (when it goes out of scope?), which causes subsequent requests to fail. (Actually, we got lucky; the DELETE to clean up is what fails, even though it probably shouldn't.)

Some of the error handling paths in pyramid_tm cause the transaction to be aborted. It might be a bug that prevents it from being aborted in this case. I'm going to file a bug upstream and see what they think.

leplatrem:
Thanks for your efforts on this :)

glasserc:
Based on my discussions with upstream, using a `beforeCommit` hook here is not the right choice. `beforeCommit` hooks are only called once the decision to commit has been made. Instead, we can use a tween to fire ResourceChanged events at the end of the request. If this raises an exception, `pyramid_tm` will abort the transaction. I've pushed a change demonstrating the technique. What do you think, @leplatrem?

leplatrem:
Excellent! If it's the right way, let's do this! I have always found it hard to pick the appropriate pattern with Pyramid :)

PR #1267, opened 2017-06-12
Fix pagination when max fetch storage is reached (fixes #1266)

### LINKED ISSUE #1266: Cannot retrieve more than `storage_max_fetch_size` records through pagination
It seems that the 10000 server limit is enforced on the number of records that have been queried, and not the number of records returned.

Eg: using `_limit=5000` will allow two requests to be made and "succeed" (with proper content and `Next-Page` headers). Using `_limit=100` will allow 100 requests. But using `_limit=10000` will only allow one.

https://github.com/mozilla-services/buildhub/issues/121#issuecomment-307729540

leplatrem:
Fixes #1266

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -480,9 +480,9 @@ class Storage(StorageBase):
             safeholders['pagination_rules'] = 'AND {}'.format(sql)
             placeholders.update(**holders)
 
-        if limit:
-            # We validate the limit value in the resource class as integer.
-            safeholders['pagination_limit'] = 'LIMIT {}'.format(limit)
+        # We validate the limit value in the resource class as integer.
+        limit = min(self._max_fetch_size, limit) if limit else self._max_fetch_size
+        safeholders['pagination_limit'] = 'LIMIT {}'.format(limit)
 
         with self.client.connect() as conn:
             result = conn.execute(query.format_map(safeholders), placeholders)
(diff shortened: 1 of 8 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -637,15 +636,15 @@ def get_all(self, collection_id, parent_id, filters=None, sorting=None,
             safeholders['pagination_rules'] = 'WHERE {}'.format(sql)
             placeholders.update(**holders)
 
-        if limit:
-            # We validate the limit value in the resource class as integer.
-            safeholders['pagination_limit'] = 'LIMIT {}'.format(limit)
+        limit = min(self._max_fetch_size, limit) if limit else self._max_fetch_size
+        # We validate the limit value in the resource class as integer.
+        safeholders['pagination_limit'] = 'LIMIT {}'.format(limit)
```
glasserc: We don't actually need this to be a `safeholder` any more, we can put `LIMIT` in the query.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -567,7 +567,6 @@ def get_all(self, collection_id, parent_id, filters=None, sorting=None,
              WHERE {parent_id_filter}
                AND collection_id = :collection_id
                {conditions_filter}
-             LIMIT {max_fetch_size}
```
glasserc: `max_fetch_size` can get removed from `safeholders` too.

glasserc:
You beat me to this! Good work!

PR #1268, opened 2017-06-13
Fix jsonschema errors

glasserc:
https://github.com/Kinto/kinto/pull/1245 changed the format of the jsonschema error messages. It isn't clear exactly why; the commit message just says "Simplify and fix coverage". https://github.com/Kinto/kinto-attachment has a test that started failing because it relies on the format of the error message. Restore it.

- (n/a) Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

r? @leplatrem


leplatrem:
>  It isn't clear exactly why; the commit message just says "Simplify and fix coverage"

There was a piece of code that was never executed. Since no test was failing without it, I just removed it :)

leplatrem:
Since #1245 was released, we'll need a changelog entry

Issue #1281, opened 2017-06-22
Accounts: Unable to create a user using POST with an admin authentication


Natim:
```'message': 'data.id in body: Username and account ID do not match.'```

**Comment by leplatrem** [comment]
Can you send config and request please?

This is explictly tested so I'm surprised...

**Comment by Natim** [comment]
```ini
#
# Auth configuration.
#
kinto.userid_hmac_secret = fcc3...d6b2dfd3f9970b

# Put your user id there instead to only allow the admin to create buckets.
kinto.bucket_create_principals = account:foo

multiauth.policies = account
multiauth.policy.account.use = kinto.plugins.accounts.authentication.AccountsAuthenticationPolicy
kinto.account_create_principals = account:foo
kinto.account_write_principals = account:foo


kinto.includes = kinto.plugins.default_bucket
                 kinto.plugins.admin
                 kinto.plugins.accounts
                 kinto.plugins.history
                 kinto_attachment
```

```echo '{"data": {"idi": "new_user", "password": "azerty"}}' | http POST http://localhost:8888/v1/accounts --auth foo:bar```

glasserc:
The tests use `account_write_principals` and `account_read_principals`, but don't have `account_create_principals` (which is the setting mentioned in the documentation). Adding `account_create_principals` breaks everything. (I'll open a PR soon.) Probably we should rerun the tests with a few different valid combinations of settings (create, read+write, create+read+write...).

glasserc:
OK, so setting `account_create_principals` gets you into an interesting bootstrapping problem where you can't create an account without being an admin, but you can't be an admin without creating the account. So that's why all the tests failed. However, what actually happened for me when I reproduced Rémy's problem is that the tests don't consider you an admin unless you also have account_write permission. This makes some sense -- probably anyone should be able to create their own account, so having that permission doesn't make you an admin -- but it's definitely missing from the documentation.

PR #1284, opened 2017-06-22
Account bugs

glasserc:
Fix a few minor bugs in the accounts plugin to try to make it more production-ready.

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

Of specific interest is #1281, where a user is in `account_create_principals` but not `account_write_principals`. The current interpretation of this configuration is that users cannot create accounts, except for the specified user, who is allowed to only create their own account. This seems nonsensical, since there's no way to authenticate the given user, so whichever user is first to create that account ID gets it. However, I was hesitant to outlaw this configuration, so I just logged a warning. The alternative would be to throw a ConfigurationError. What do you think?

r? @leplatrem @Natim

### DIFF
--- a/kinto/plugins/accounts/views.py
+++ b/kinto/plugins/accounts/views.py
@@ -27,6 +27,12 @@ def _extract_posted_body_id(request):
         raise http_error(httpexceptions.HTTPUnauthorized(), error=error_msg)
 
 
+class AccountIdGenerator(NameGenerator):
+    """Allow @ signs in account IDs."""
+
+    regexp = r'^[a-zA-Z0-9][.@a-zA-Z0-9_-]*$'
+
+
 class AccountSchema(resource.ResourceSchema):
     password = colander.SchemaNode(colander.String())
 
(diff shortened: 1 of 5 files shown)


glasserc:
I changed my mind, and decided that it makes more sense to raise a ConfigurationError if someone is allowed to create accounts but isn't themselves an admin, because there doesn't appear to be a valid use case for it yet. We can always undo this restriction later if it turns out to be useful to someone.

Natim:
Yeah it makes sense somehow. I added account_create because I didn't want some people to be able to change other people password or delete other people account while they were able to create new people.

kinto/plugins/accounts/views.py
```diff
@@ -27,6 +27,12 @@ def _extract_posted_body_id(request):
         raise http_error(httpexceptions.HTTPUnauthorized(), error=error_msg)
 
 
+class AccountIdGenerator(NameGenerator):
+    """Allow @ signs in account IDs."""
+
+    regexp = r'^[a-zA-Z0-9][.@a-zA-Z0-9_-]*$'
```
Natim: Nice 👍

mariustresor1:
Thanks @glasserc 👌

PR #1300, opened 2017-07-24
Add cluster property

rkleine:
The app will crash if `pusher.cluster` is not set.

Fixes #

- [X] Add documentation.

### DIFF
--- a/docs/tutorials/notifications-websockets.rst
+++ b/docs/tutorials/notifications-websockets.rst
@@ -47,6 +47,7 @@ add these lines to setup a listener, and be notified of record updates per colle
     pusher.app_id = <pusher-app-id>
     pusher.key = <pusher-key>
     pusher.secret = <pusher-secret>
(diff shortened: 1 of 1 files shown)


docs/tutorials/notifications-websockets.rst
```diff
@@ -47,6 +47,7 @@ add these lines to setup a listener, and be notified of record updates per colle
     pusher.app_id = <pusher-app-id>
     pusher.key = <pusher-key>
     pusher.secret = <pusher-secret>
+    pusher.cluster = <pusher-cluster>
```
leplatrem: Should we add an example? Or put `pusher.cluster = eu` ?

Natim:
Thanks @rkleine

docs/tutorials/notifications-websockets.rst
Natim: 6f57dfd5

PR #1316, opened 2017-08-17
Add a kinto create user command.

### LINKED ISSUE #1315: Add a kinto createsuperuser command
With the [accounts plugin](https://github.com/Kinto/kinto/tree/master/kinto/plugins/accounts) it is now possible to handle user directly in Kinto.

However currently the steps are to activate the user creation for everyone, create an administrator account and then enable the creation of account only for that superuser. 

Each configuration change being a server restart.

For deployment it is something tricky to do. It would be convenient to have a `kinto createsuperuser` command that works a bit like the [`django-admin` command](https://docs.djangoproject.com/fr/1.11/ref/django-admin/#createsuperuser) and allow the creation of the super user from the server.

Natim:
Fixes #1315

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/docs/api/1.x/accounts.rst
+++ b/docs/api/1.x/accounts.rst
@@ -34,6 +34,21 @@ Add the following settings to the ``.ini`` file:
     kinto.account_create_principals = system.Everyone
 
 
+You can use the ``create-user`` command to create an admin:
+
+.. code-block:: bash
+
+    kinto create-user --ini /etc/kinto.ini --username admin --password ThisIsN0tASecurePassword
+
+You can then use it in your config:
+
+.. code-block:: ini
+
+    # Allow anyone to create accounts.
+    kinto.account_create_principals = account:admin
+    kinto.account_write_principals = account:admin
+
+	
 .. _accounts-auth:
 
 Authentication
--- a/kinto/plugins/accounts/scripts.py
+++ b/kinto/plugins/accounts/scripts.py
@@ -34,7 +34,7 @@ def create_user(env, username=None, password=None):
             username = input('Username: ')
 
         if password is None:
-            while "The user didn't entered twice the same password":
+            while True:  # The user didn't entered twice the same password
                 password = getpass.getpass('Please enter a password for {}: '.format(username))
                 confirm = getpass.getpass('Please confirm the password: '.format(username))
 
--- a/kinto/plugins/accounts/scripts.py
+++ b/kinto/plugins/accounts/scripts.py
@@ -1,0 +1,62 @@
+import logging
+import getpass
+from pyramid.settings import asbool
+
+from .utils import hash_password
+from .views import AccountIdGenerator
+
+
+logger = logging.getLogger(__name__)
+
+
+def create_user(env, username=None, password=None):
+    """Administrative command to create a new user."""
+    registry = env['registry']
+    settings = registry.settings
+    readonly_mode = asbool(settings.get('readonly', False))
+
+    if readonly_mode:
+        message = ('Cannot create a user with a readonly server.')
+        logger.error(message)
+        return 51
+
+    if 'kinto.plugins.accounts' not in settings['includes']:
+        message = ('Cannot create a user when the accounts plugin is not installed.')
+        logger.error(message)
+        return 52
+
+    try:
+        validator = AccountIdGenerator()
+        if username is None:
+            username = input('Username: ')
+        while True:
+            if not validator.match(username):
+                print('{} is not a valid username.')
+                print('Username should match {0!r}, please try again.'.format(validator.regexp))
+                username = input('Username: ')
+            else:
+                break
+
+        if password is None:
+            while True:
+                password = getpass.getpass('Please enter a password for {}: '.format(username))
+                confirm = getpass.getpass('Please confirm the password: '.format(username))
+
+                if password != confirm:
+                    print('Sorry, passwords do not match, please try again.')
+                else:
+                    break
+    except EOFError:
+        print('User creation aborted')
+        return 53
+
+    print("Creating user '{}'".format(username))
+    record = {'id': username, 'password': hash_password(password)}
+    registry.storage.create(collection_id='account',
+                            parent_id=username,
+                            record=record,
+                            ignore_conflict=True)
+    registry.permission.add_principal_to_ace(
+        '/accounts/{}'.format(username), 'write', 'account:{}'.format(username))
+
+    return 0
--- a/kinto/plugins/accounts/scripts.py
+++ b/kinto/plugins/accounts/scripts.py
@@ -29,13 +29,10 @@ def create_user(env, username=None, password=None):
         validator = AccountIdGenerator()
         if username is None:
             username = input('Username: ')
-        while True:
-            if not validator.match(username):
-                print('{} is not a valid username.')
-                print('Username should match {0!r}, please try again.'.format(validator.regexp))
-                username = input('Username: ')
-            else:
-                break
+        while not validator.match(username):
+            print('{} is not a valid username.')
+            print('Username should match {0!r}, please try again.'.format(validator.regexp))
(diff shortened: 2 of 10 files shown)


Natim:
> I'm not super crazy about this because it reinforces the idea of the accounts API as Basic Auth v2.

Today that's what it is. What you describe in #1317 achieve another purpose than our current accounts plugin.

kinto/plugins/accounts/scripts.py
```diff
@@ -0,0 +1,59 @@
...
+
+    print("Creating user '{}'".format(username))
+    record = {'id': username, 'password': hash_password(password)}
+    registry.storage.create(collection_id='account',
+                            parent_id=username,
+                            record=record,
+                            ignore_conflict=True)
+    registry.permission.add_principal_to_ace(
+        '/accounts/{}'.format(username), 'write', 'account:{}'.format(username))
+
```
leplatrem: You could also call the POST view with a fake request instead no? (just wondering...)

Natim: You don't want to have to worry about permission handling I guess.

glasserc:
> Today that's what it is.

Maybe we should rename it from "accounts API" to "basic auth v2".

Natim:
> Maybe we should rename it from "accounts API" to "basic auth v2".

Let's discuss that in another issue.

tests/plugins/test_accounts.py
```diff
@@ -319,3 +321,59 @@ def test_fallsback_on_basicauth(self):
...
+
+
+class CreateUserTest(unittest.TestCase):
+    def setUp(self):
+        self.registry = mock.MagicMock()
+        self.registry.settings = {
+            "includes": "kinto.plugins.accounts",
+        }
+
+    def test_create_user_in_read_only_display_an_error(self):
```
leplatrem: nit: displays

tests/plugins/test_accounts.py
```diff
@@ -319,3 +321,59 @@ def test_fallsback_on_basicauth(self):
...
+        with mock.patch('kinto.plugins.accounts.scripts.logger') as mocked:
+            self.registry.settings['readonly'] = 'true'
+            code = scripts.create_user({'registry': self.registry})
+            assert code == 51
+            mocked.error.assert_called_once_with('Cannot create a user with '
+                                                 'a readonly server.')
+
+    def test_create_user_when_not_included_display_an_error(self):
```
leplatrem: nitto

tests/plugins/test_accounts.py
```diff
@@ -319,3 +321,59 @@ def test_fallsback_on_basicauth(self):
...
+        with mock.patch('kinto.plugins.accounts.scripts.input', side_effect=["&zert", "username"]):
+            with mock.patch('kinto.plugins.accounts.scripts.getpass.getpass', side_effect=[
+                    "password", "p4ssw0rd", "password", "password"]):
+                code = scripts.create_user({'registry': self.registry}, None, None)
+                self.registry.storage.create.assert_called_once()
+                self.registry.permission.add_principal_to_ace.assert_called_with(
+                    "/accounts/username", "write", "account:username")
+                assert code == 0
+
+    def test_create_user_with_an_valid_username_and_password_confirmation(self):
```
leplatrem: nit: a valid

tests/plugins/test_accounts.py
```diff
@@ -319,3 +321,59 @@ def test_fallsback_on_basicauth(self):
...
+
+    def test_create_user_with_an_invalid_username_and_password_confirmation(self):
+        with mock.patch('kinto.plugins.accounts.scripts.input', side_effect=["&zert", "username"]):
+            with mock.patch('kinto.plugins.accounts.scripts.getpass.getpass', side_effect=[
+                    "password", "p4ssw0rd", "password", "password"]):
+                code = scripts.create_user({'registry': self.registry}, None, None)
+                self.registry.storage.create.assert_called_once()
+                self.registry.permission.add_principal_to_ace.assert_called_with(
+                    "/accounts/username", "write", "account:username")
+                assert code == 0
```
leplatrem: In this test, the titlle suggests that it will fail to create it (compared to the next test below)

tests/plugins/test_accounts.py
```diff
@@ -319,3 +321,59 @@ def test_fallsback_on_basicauth(self):
...
+        with mock.patch('kinto.plugins.accounts.scripts.input', return_value="username"):
+            with mock.patch('kinto.plugins.accounts.scripts.getpass.getpass',
+                            return_value="password"):
+                code = scripts.create_user({'registry': self.registry})
+                self.registry.storage.create.assert_called_once()
+                self.registry.permission.add_principal_to_ace.assert_called_with(
+                    "/accounts/username", "write", "account:username")
+                assert code == 0
+
+    def test_create_user_with_an_valid_username_and_password_parameters(self):
```
leplatrem: nit: with valid

docs/api/1.x/accounts.rst
```diff
@@ -34,6 +34,21 @@ Add the following settings to the ``.ini`` file:
     kinto.account_create_principals = system.Everyone
 
 
+You can use the ``create-user`` command to create an admin:
+
+.. code-block:: bash
+
+    kinto create-user --ini /etc/kinto.ini --username admin --password ThisIsN0tASecurePassword
```
leplatrem: I think the most common way is to rely on prompt for password no?

kinto/plugins/accounts/scripts.py
```diff
@@ -0,0 +1,58 @@
...
+logger = logging.getLogger(__name__)
+
+
+def create_user(env, username=None, password=None):
+    """Administrative command to create a new user."""
+    registry = env['registry']
+    settings = registry.settings
+    readonly_mode = asbool(settings.get('readonly', False))
+    if readonly_mode:
+        message = ('Cannot create a user with a readonly server.')
```
leplatrem: nit: superflous parenthesis

kinto/plugins/accounts/scripts.py
```diff
@@ -0,0 +1,58 @@
...
+        logger.error(message)
+        return 51
+
+    if 'kinto.plugins.accounts' not in settings['includes']:
+        message = ('Cannot create a user when the accounts plugin is not installed.')
```
leplatrem: nitto

kinto/plugins/accounts/scripts.py
```diff
@@ -0,0 +1,58 @@
...
+        validator = AccountIdGenerator()
+        if username is None:
+            username = input('Username: ')
+        while not validator.match(username):
+            print('{} is not a valid username.')
+            print('Username should match {0!r}, please try again.'.format(validator.regexp))
+            username = input('Username: ')
+
+        if password is None:
+            while True:
```
leplatrem: nit: http://sametmax.com/alternative-au-do-while-en-python/

kinto/plugins/accounts/scripts.py
```diff
@@ -0,0 +1,58 @@
...
+        return 53
+
+    print("Creating user '{}'".format(username))
+    record = {'id': username, 'password': hash_password(password)}
+    registry.storage.create(collection_id='account',
+                            parent_id=username,
+                            record=record,
+                            ignore_conflict=True)
+    registry.permission.add_principal_to_ace(
+        '/accounts/{}'.format(username), 'write', 'account:{}'.format(username))
```
leplatrem: nit: this identation does not match the instruction above at line 51

docs/api/1.x/accounts.rst
Natim: That's the case if you don't provide any arguments, however OPS will use SOPS and ENV variables so they want to be able to do: `into create-user --ini /etc/kinto.ini --username $USERNAME --password $PASSWORD`

PR #1319, opened 2017-08-18
Reword description of ETags

Grahack:
Enhances documentation.

### DIFF
--- a/docs/tutorials/synchronisation.rst
+++ b/docs/tutorials/synchronisation.rst
@@ -165,7 +165,7 @@ ETags are provided in response headers, for the collection as well as individual
 records.
 
 Even though it is recommended to consider them as opaque and abstract, it can still
-be useful to notice that ETags are a string with the quoted record last modified value
+be useful to notice that ETags are the ``last_modified`` value of the record, quoted
 (``"<record.last_modified>"``)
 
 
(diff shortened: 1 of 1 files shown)


Natim:
Sounds really french to me like that :/ I guess the previous version is better english.
I will let our expert @glasserc decide :)

docs/tutorials/synchronisation.rst
```diff
@@ -165,7 +165,7 @@ ETags are provided in response headers, for the collection as well as individual
 records.
 
 Even though it is recommended to consider them as opaque and abstract, it can still
-be useful to notice that ETags are a string with the quoted record last modified value
+be useful to notice that ETags are the ``last_modified`` value of the record, quoted
```
glasserc: I think this is fine, but having `quoted` at the end of the noun phrase feels a little clumsy to me. Normally I'd probably write `ETags are the quoted ``last_modified`` value of the record`, but I see that the example `(``"<record.last_modified>"``)` being next to the word "quoted" is kind of nice too.

I'd probably also use "observe" rather than "notice" here, but I think notice is OK.

Grahack:
About writing English, I'd recommend this book, which also improved my French writing!!!
https://en.wikipedia.org/wiki/The_Elements_of_Style

PR #1332, opened 2017-09-22
Add memcached cache backend.

### LINKED ISSUE #1311: Write a memcache cache backend
Memcache is a great candidate for a cache backend, because it's a key-value cache with a TTL. One of our ops guys mentioned that memcache is more robust than redis because it's not master-slave replicated.

Natim:
Fixes #1311 

- [x] Add documentation.
- [X] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -91,7 +91,8 @@ Backends
 Kinto relies on three types of backends: storage, cache and permission. The
 settings names have a different prefix for each.
 
-For each of them, the supported services are currently PostgreSQL, Redis, Memcached and Memory.
+For each of them, the supported services are currently PostgreSQL, Redis, and Memory.
+Memcached is also available as a cache backend.
 
 Storage
 :::::::
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -91,7 +91,7 @@ Backends
 Kinto relies on three types of backends: storage, cache and permission. The
 settings names have a different prefix for each.
 
-For each of them, the supported services are currently PostgreSQL, Redis, and Memory.
+For each of them, the supported services are currently PostgreSQL, Redis, Memcached and Memory.
 
 Storage
 :::::::
@@ -163,6 +163,10 @@ Cache
 +----------------------------+-----------------------------+------------------------------------------------------------------------------+
 | kinto.cache_max_backlog    | ``-1``                      | Number of threads that can be in the queue waiting for a connection.         |
 +----------------------------+-----------------------------+------------------------------------------------------------------------------+
+| kinto.cache_hosts          | ``''``                      | The space separated list of memcached hosts.                                 |
++----------------------------+-----------------------------+------------------------------------------------------------------------------+
+
+**For PostgreSQL**
 
 .. code-block:: ini
 
@@ -171,6 +175,14 @@ Cache
 
     # Control number of pooled connections
     # kinto.cache_pool_size = 50
+
+
+For **memcached**
+
+.. code-block:: ini
+
+    kinto.cache_backend = kinto.core.cache.memcached
+    kinto.cache_hosts = 127.0.0.1:11211 127.0.0.2:11211
 
 Permissions
 :::::::::::
--- a/kinto/core/cache/memcached.py
+++ b/kinto/core/cache/memcached.py
@@ -19,7 +19,8 @@ def wrap_memcached_error(func):
             return func(*args, **kwargs)
         except TypeError:
             raise
-        except Exception as e:
+        except (memcache.Client.MemcachedKeyError,
+                memcache.Client.MemcachedStringEncodingError) as e:
             logger.exception(e)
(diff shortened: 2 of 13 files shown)


docs/configuration/settings.rst
```diff
@@ -91,7 +91,7 @@ Backends
 Kinto relies on three types of backends: storage, cache and permission. The
 settings names have a different prefix for each.
 
-For each of them, the supported services are currently PostgreSQL, Redis, and Memory.
+For each of them, the supported services are currently PostgreSQL, Redis, Memcached and Memory.
```
glasserc: This isn't true. This sentence should stay the same, along with `Memcached is also available as a cache backend.`

kinto/core/cache/memcached.py
```diff
@@ -0,0 +1,106 @@
...
+def wrap_memcached_error(func):
+    @wraps(func)
+    def wrapped(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except TypeError:
+            raise
+        except Exception as e:
+            logger.exception(e)
+            raise exceptions.BackendError(original=e)
```
glasserc: What is this about? Is there no Memcache-specific exception parent class we can use here?

kinto/core/cache/memcached.py
```diff
@@ -0,0 +1,106 @@
...
+    return wrapped
+
+
+def create_from_config(config, prefix=''):
+    """Redis client instantiation from settings.
+    """
+    settings = config.get_settings()
+    hosts = aslist(settings[prefix + 'hosts'])
+    return memcache.Client(hosts)
```
glasserc: This might be a silly question, but do we have to worry about reentrancy here (sharing the Memcache client across threads)? Or does uwsgi just spawn multiple threads with their own interpreters?

kinto/core/cache/memcached.py
```diff
@@ -0,0 +1,106 @@
...
+        else:
+            value = self.get(key)
+            self.set(key, value, ttl)
+
+    @wrap_memcached_error
+    def set(self, key, value, ttl):
+        if isinstance(value, bytes):
+            raise TypeError("a string-like object is required, not 'bytes'")
+        value = json.dumps({"value": value, "ttl": ceil(time() + ttl)})
+        self._client.set(self.prefix + key, value, int(ttl))
```
glasserc: I guess we use this JSON format so that we can retrieve the TTL in the `ttl()` method? It's kind of a bummer that there's no `ttl` command (as in Redis).

kinto/core/cache/memcached.py
```diff
@@ -0,0 +1,106 @@
...
+        value, _ = self._get(key)
+        return value
+
+    @wrap_memcached_error
+    def expire(self, key, ttl):
+        if ttl == 0:
+            self.delete(key)
+        else:
+            value = self.get(key)
+            self.set(key, value, ttl)
```
glasserc: A comment explaining why we can't use the `touch` command would be a good idea.

kinto/core/cache/memcached.py
Natim: I didn't find one unfortunately :(

kinto/core/cache/memcached.py
Natim: The latter.

kinto/core/cache/memcached.py
Natim: Maybe we can actually. I will have a look.

kinto/core/cache/memcached.py
Natim: Yes I know, apparently there is a way using stats but it doesn't looks as robust as my workaround.

kinto/core/cache/memcached.py
Natim: Actually not with the TTL...

docs/configuration/settings.rst
```diff
@@ -163,6 +164,10 @@ Cache
 +----------------------------+-----------------------------+------------------------------------------------------------------------------+
 | kinto.cache_max_backlog    | ``-1``                      | Number of threads that can be in the queue waiting for a connection.         |
 +----------------------------+-----------------------------+------------------------------------------------------------------------------+
+| kinto.cache_hosts          | ``''``                      | The space separated list of memcached hosts.                                 |
```
leplatrem: `Memcached`

docs/configuration/settings.rst
```diff
@@ -172,6 +177,14 @@ Cache
     # Control number of pooled connections
     # kinto.cache_pool_size = 50
 
+
+For **memcached**
```
leplatrem: Memcached

setup.py
```diff
@@ -104,6 +108,7 @@ def read_file(filename):
       install_requires=REQUIREMENTS,
       extras_require={
           'redis': REDIS_REQUIRES,
+          'memcached': MEMCACHED_REQUIRES,
```
leplatrem: This should probably be documented in the install/settings section

PR #1333, opened 2017-09-27
Documentation about Memcached

Natim:


leplatrem:
I think we should also mention the `pip install kinto[memcached]` somewhere

Natim:
Isn't it https://github.com/Kinto/kinto/pull/1333/files#diff-84c3e7bada317747370dff2b0c64cfb8R183 ?

Natim:
I created #1334 and #1335 about that thanks.

PR #1338, opened 2017-09-28
Use Python:3.6-slim as the base image.

Natim:


glasserc:
I'm OK with this, but what's the purpose of it? Is it smaller or something?

Natim:
>  Is it smaller or something?

Yes a bit smaller.

Issue #1339, opened 2017-09-28
`kinto create-user` command should fallback to KINTO_INI env variable for the config file;


Natim:

bqbn:
Sounds good.

When this gets implemented, do we must put the user password in the env file?

Natim:
I guess we can already do: `kinto create-user --username "$USERNAME" --password "$PASSWORD" --ini $KINTO_INI`

What variable name would you like to default to for username and password?

PR #1340, opened 2017-09-29
Commit the transaction at the end of the create-user script.

### LINKED ISSUE #1336: `kinto create-user` doesn't override the password if the user already exists.

Natim:
Fixes #1336

- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,10 +3,12 @@ Changelog
 
 This document describes changes between each past release.
 
-7.6.0 (unreleased)
+7.5.1 (unreleased)
 -------------------
 
-- Nothing changed yet.
+**Bug fixes**
+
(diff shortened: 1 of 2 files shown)


kinto/__main__.py
```diff
@@ -11,7 +11,7 @@
 from kinto import __version__
 from kinto.config import init
 
-DEFAULT_CONFIG_FILE = 'config/kinto.ini'
+DEFAULT_CONFIG_FILE = os.getenv('KINTO_INI', 'config/kinto.ini')
```
leplatrem: Looks like you are also fixing #1339

Natim: You are right I should do that in different PR's

leplatrem: Not necessarily, those are tiny fixes, but don't forget to update CHANGELOG

PR #1344, opened 2017-10-01
faq.rst: Spelling correction

stellargo:
'overriden' was corrected to 'overridden'

Fixes #

- [x] Add documentation.
- [ ] Add a changelog entry.


Natim:
I know that both words exists but I will let @glasserc review it.

glasserc:
As far as I know, there is no word `overriden`, only `overridden` is correct. Nice catch!

PR #1345, opened 2017-10-01
Requirements

haroon-sheikh:
Fixes #

- [ ] Added dependencies status badge in README
- [ ] Tags dev and docs requirements (Best Practice)
- [ ] Updates pyramid dependency to latest.

Do we know why `setuptools` is tagged at `<36`? Is there a reason for this? Can we update to latest?


Natim:
Thanks, but that's not how we handle requirements in that project you can read [this article](http://www.servicedenuages.fr/en/handling-python-projects-dependencies) for more information

haroon-sheikh:
That makes sense @Natim. Will close the PR.

PR #1353, opened 2017-10-21
Fix JSON Merge on removing permissions (fixes #1322)

### LINKED ISSUE #1322: Merge PATCH doesn't work on permissions
```
$ http -a 'user:pass' http://localhost:8888/v1/buckets/b4/collections/cc
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Expires, Content-Length, Pragma, Backoff, Cache-Control, Alert, Retry-After, ETag, Last-Modified
Cache-Control: no-cache, no-store
Content-Length: 182
Date: Fri, 18 Aug 2017 16:52:04 GMT
Etag: "1503075082664"
Last-Modified: Fri, 18 Aug 2017 16:51:22 GMT
Server: waitress

{
    "data": {
        "a": "whoo",
        "id": "cc",
        "last_modified": 1503075082664
    },
    "permissions": {
        "read": [
            "hi"
        ],
        "write": [
            "basicauth:08eee2542a2328390dfaa7390fa7df0d64a7081b44d629bfacd9b4e02bd2676d"
        ]
    }
```

Let's get rid of the `read` permissions.

```
$ http -a 'user:pass'  PATCH http://localhost:8888/v1/buckets/b4/collections/cc "Content-Type:application/merge-patch+json" data:='{"a": null}' permissions:='{"read": null}'
HTTP/1.1 400 Bad Request
Access-Control-Expose-Headers: Alert, Backoff, Content-Length, Retry-After
Content-Length: 215
Date: Fri, 18 Aug 2017 16:52:47 GMT
Server: waitress

{
    "code": 400,
    "details": [
        {
            "description": "\"None\" is not iterable",
            "location": "body",
            "name": "permissions.read"
        }
    ],
    "errno": 107,
    "message": "permissions.read in body: \"None\" is not iterable"
}
```

Note that JSON Patch does work by applying changes uniformly to `data` and `permissions`.

gabisurita:
Fixes  #1322

Allow null on permissions to support removing lists of principals on record permissions. As a side effect this allows passing null to regular patch operations with no effect, but we can change this if needed.

- [x] Add tests.
- [x] Add a changelog entry.


leplatrem:
Gabi \o/ Welcome back :)

Thanks for your «patch» (huhu)

To be honest, I could not take time to review it thoroughly yet. But the tests look good!

I think we might have to be a bit more talkative about the changes in the changelog (and API changelog?) with regards to null values. (as well as in the docs on patch endpoints)

gabisurita:
Hey @leplatrem ! Good to see you! 

I can update the documentation as soon as I have some time or keep the current behavior by not allowing null values using custom validation. What do you think makes more sense?

kinto/core/resource/__init__.py
```diff
@@ -1184,18 +1184,29 @@ def process_record(self, new, old=None):
...
+                                  ignores=(None,))
+
         else:
-            permissions = self.request.validated['body'].get('permissions', {})
+            permissions = {k: v for k, v in payload.get('permissions', {}).items()
+                           if v is not None}
 
         annotated = {**new}
 
         if permissions:
```
glasserc: This is outside the scope of this PR, but I guess this means we can never set the permissions to the empty object.

glasserc:
Also in the documentation is this example, which I think is incorrect: `{"a": {"b":"c"}} + {"a":{"d":"e"}} → {"a":{"b:c", "d":"e"}}` (because "b:c" isn' valid JSON). That might be outside the scope of this PR, but if you're in there anyhow... 😇

kinto/core/resource/__init__.py
gabisurita: Actually, this change is meaningful to the fix. Without it a null value on a regular patch would remove permissions as well.

gabisurita:
@glasserc Thanks for the review! I've tried to sketch some minor upgrades on the documentation for this section. What do you think?

gabisurita:
Flake8 seems to be failing because of this new error "E722 do not use bare except". May I just ignore this error and open an issue for it?
 https://travis-ci.org/Kinto/kinto/jobs/297283472

docs/api/1.x/_details-patch-object.rst
```diff
@@ -82,8 +65,20 @@ For more information about each operation, please refer to
...
+the current value before adding some principal to the list.
+
+Permissions
+-----------
+
+.. In the JSON request payloads, at least one of ``data`` and ``permissions``
+.. must be provided.
+
+When permissions are modified, the same behaviour used for the body is expected.
+But notice thought that there are some permissions specific behaviours, such as:
```
glasserc: I think `but` and `though` (N.B. typo) are redundant -- either one is fine. I'd probably write `note` instead of `notice`. `permissions-specific`.

kinto/core/resource/__init__.py
glasserc: Yes, but IMHO passing an empty object should be different from not passing anything.

docs/api/1.x/_details-patch-object.rst
```diff
@@ -92,3 +87,17 @@ on permission operations and can be omitted.
         { "op": "add", "path": "/permissions/read/system.Everyone" },
         { "op": "remove", "path": "/permissions/read/fxa:bob" }
     ]
+
+3. On JSON Merge operations, ``null`` values can be used to remove all principals
+   from a specific permission level. As a consequence, using ``null`` for permissions
+   on other types of patch operations is considered a no-op.
```
glasserc: Wait, maybe I misunderstood, I didn't think that was a consequence. I would have written something like `However, using null instead of a set of principals when using other patch types is not valid and will result in an error`.

gabisurita: From my understanding what I've described is the implemented behavior. This is what I've tried to explain in the PR by saying "As a side effect this allows passing null to regular patch operations with no effect".

glasserc: Ah, yes, then I misunderstood. I thought we had logic to only do the preprocess/postprocess null step if we were in a merge patch. In that case, why do we have this preprocess/postprocess step, versus just allowing permissions to be nullable? I'm guessing it's because we want to only allow nulls in certain kinds of operations, but it seems like we added it to the PermissionSchema, so it seems like it's global anyhow?

Natim:
Shall I merge and release this?

PR #1361, opened 2017-11-07
Remove timestamps denormalization (fixes #1357)

### LINKED ISSUE #1357: PostgreSQL timestamps bottleneck
A [long time ago](https://github.com/Kinto/kinto/commit/c6d81f7f4794a9ec2c0734a5111f84cc8d7d393a) we added the `timestamps` table, that stores the highest timestamps for each resource.

It has mainly one purpose: it allows the `ETag` obtained on an empty collection of records to be always the same until some records are created.

But the current implementation makes it a serious bottleneck. Every operation on the same resource (eg. same `(parent_id, collection_id)`) leads to an update of the same row in the `timestamps` table.

This applies to:
* creation of buckets (`(parent_id='', collection_id='bucket')`)
* creation of collections in the same bucket (`(parent_id=/buckets/{bid}, collection_id='collection')`)
* creation of records in the same collection (`(parent_id=/buckets/{bid}/collections/{cid}, collection_id='record')`)
* ...and some others like groups, accounts, ...

Since the timestamp column has a sorted index, I believe we can get rid of this denormalization and use the `timestamps` table just for the corner case of empty collections. For the nominal situation, the highest timestamp would be the max between the first row of the records index and the first row of the tombstones index.

leplatrem:
Fixes #1357


leplatrem:
I'm blocked with two failing tests:
* `test_create_ignores_specified_last_modified_if_in_the_past`
* `test_update_ignores_specified_last_modified_if_in_the_past`

There is no way I can *bump* the collection timestamp without a `timestamps`  table if we allow specifying last_modified field in the past.

To be honest, now that I completely forgot the background story behind this specification, I can't figure out why and how it makes sense to bump the collection timestamp when you specify last_modified in the past.  The polling for changes with (`?_since=`) won't catch the changes anyway :|

leplatrem:
The original implementation was done here:
https://github.com/Kinto/kinto/commit/805cf67c2e470579c223664a7e698f040333f612
Apparently to align PostgreSQL with other backends implementations (ie. memory and redis)

Can someone tell me why it makes sense to support last_modified in the past when the collection is not empty ?

Natim:
I don't think it does to be honest.

glasserc:
The changes look fine to me. The only wrinkle I can think of is emptying an existing collection, but you already thought of that by handling tombstones.

I'm not 100% sure how to handle updating timestamps that are in the past. If we set last_modified back to the "new" old record, then we will re-sync certain records. If we don't update last_modified, there's no way for the client to know about the records in the past. I think we have to accept that putting records in the past is going to break synchronization. Based on that, I would say we should try to keep last_modified monotonic.

My opinion, therefore, about the right behavior of the two tests:

- `test_create_ignores_specified_last_modified_if_in_the_past`: `timestamp` should be the same as `timestamp_before` (change `assertGreater` to `assertEqual`).
- `test_update_ignores_specified_last_modified_if_in_the_past`: `timestamp` goes backwards here because the `last_modified` of the only record goes backwards. We could handle this by insert into `timestamps` on the first record save, which will ensure that the collection's timestamp doesn't go backwards in this case, but still leaves us open to slightly more complicated cases (record A is inserted, followed by record B; timestamp_before is taken; B is put 10 seconds in the past; timestamp is now record A). So I think the right behavior here is to do nothing (let `timestamp` go backwards). Anyone explicitly setting last_modified had better know what they're doing and might break synchronization.

CHANGELOG.rst
```diff
@@ -6,6 +6,10 @@ This document describes changes between each past release.
 7.6.0 (unreleased)
 ------------------
 
+**Protocol**
+
+- XXX last_modified in the past...
```
Natim: Please improve the message here.

leplatrem:
Before merging this, we should it does not break:
- [ ] kinto-signer
- [ ] kinto-changes + kinto-signer (= kinto-dist)
- [ ] kinto-wizard

Natim:
I ran kinto-dist e2e tests with success.

Natim:
Let's merge it and see how it goes in kinto-signer / kinto-dist / kinto-changes respectives CI

PR #1365, opened 2017-11-09
Improve logging (fixes #1363)

### LINKED ISSUE #1363: Improvements to logging
There are two goals we want to achieve here: 

1. Information to aid debugging / troubleshooting
2. Telemetry / measuring of service

**General things we should do:**

1. Log all http responses we generate
    - 2xx, 3xx, 4xx, 5xx ... log it. 
    - if we send a non 5xx, must have `Fields.errno`, `Fields.msg`
    - use `Fields.code` for the HTTP response code
    - always have `Fields.path` and `Fields.method`
2. `Fields.msg` is for humans. Not for computers. If we must dump a python exception put it somewhere else, ie: `Fields.error` 
3. Omit empty fields. At scale all those extra bits start costing storage/io/cpu to skip
4. when logging level is DEBUG, we should log http request headers/params/body (if printable) and response headers/params/body(if printable)
    - this will dump a ton of data but super helpful to flip a flag and see issues

**Specific changes**:

This is what's logged for a successful POST to /v1/batch:

```
{
  "Severity": 6,
  "Logger": "%",                       [1]
  "Type": "subrequest.summary",        
  "Fields": {                          
    "uid": null,                       [2]
    "querystring": {},                 [3]
    "authn_type": null,                [4]
    "lang": "en-US,en;q=0.5",          
    "batch_size": 1,                   
    "path": "/v1/batch",               
    "errno": null,                     [5]
    "message": "subrequest.summary",   [6]
    "method": "POST",
    "agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64..."
  },
  "Pid": 10,
  "Timestamp": 1510145585496232400,
  "Hostname": "ip-<redacted>",
  "EnvVersion": "2.0"
}
```

1. Should be `kintowe`
2. omit if `null`
3. omit if empty
4. omit if `null`
5. should be `0` for success
6. key should be `msg`, not `message`

Here is an error log:

```
{
  "Severity": 6,
  "Logger": "%",
  "Type": "kinto.core.storage.postgresql.pool.QueuePoolWithMaxBacklog",
  "Fields": {
    "message": "Connection <ConnectionWrapper at 0x7f1a95356d48 for psycopg2.extensions.connection at 0x7f1a9d4dd508> exceeded timeout; recycling"
  },
  "Pid": 18,
  "Timestamp": 1510174390025757700,
  "Hostname": "ip-<redacted>",
  "EnvVersion": "2.0"
}
```

1. `Logger` - should be `kintowe`
2. `Fields.message` should be `Fields.msg`
    - ideal would be something human readable with another field `Fields.error` with debugging information
3. `Fields.errno` is missing. Maybe use a code from [errors.py](https://github.com/Kinto/kinto/blob/master/kinto/core/errors.py)
4. `Fields.path` and `Fields.method` are missing. we should log info about the request that was made that caused the error.  Also helps to match up with nginx logs if available

Here's a slightly longer one: 

```
{
  "Severity": 3,
  "Logger": "%",
  "Type": "kinto.core.storage.postgresql.client",
  "Fields": {
    "message": "(psycopg2.IntegrityError) duplicate key value violates unique constraint \"deleted_pkey\"\nDETAIL:  Key (id, parent_id, collection_id)=(<redacted>, , bucket) already exists.\n [SQL: '\\n            WITH deleted_record AS (\\n                DELETE\\n                FROM records\\n                WHERE id = %(object_id)s\\n                  AND parent_id = %(parent_id)s\\n                  AND collection_id = %(collection_id)s\\n                RETURNING id\\n            )\\n            INSERT INTO deleted (id, parent_id, collection_id, last_modified)\\n            SELECT id, %(parent_id)s, %(collection_id)s, from_epoch(%(last_modified)s)\\n              FROM deleted_record\\n            RETURNING as_epoch(last_modified) AS last_modified;\\n            '] [parameters: {'object_id': '<redacted>', 'collection_id': 'bucket', 'last_modified': None, 'parent_id': ''}]"
  },
  "Pid": 8,
  "Timestamp": 1510174685609011700,
  "Hostname": "ip-<redacted>",
  "EnvVersion": "2.0"
}
```

1. Same changes as other error message comments above

leplatrem:
Fixes #1363

- [x] Do not log empty values
- [x] Fix upstream `msg` field name https://github.com/mozilla/mozilla-cloud-services-logger/pull/4
- [x] Find out a way to set `Logger` without using  ugly `style = ` in config
- [x] Add header/body logger when level is DEBUG
- [x] Add extra request info in error logs
- [x] Find out why `exc_info` does not output `error` and `traceback` fields
- [x] Find out why we end up with `"path": "/v1/batch",` in a `subrequest.summary`


kinto/core/initialization.py
```diff
@@ -335,14 +335,15 @@ def on_new_request(event):
                 errno=errors.ERRORS.INVALID_PARAMETERS,
                 message='Invalid URL path.')
 
+        qs = dict(errors.request_GET(request))
         request.log_context(agent=request.headers.get('User-Agent'),
                             path=request_path,
                             method=request.method,
-                            querystring=dict(errors.request_GET(request)),
+                            querystring=(qs if len(qs) else None),
```
glasserc: You could shorten this to `qs or None`. But why not just build a dict of parameters we want to include and only add `querystring` if there's something here, similar to the way you do in the error view?

kinto/core/__init__.py
```diff
@@ -154,6 +171,9 @@ def add_api_capability(config, identifier, description='', url='', **kw):
     # Add CORS settings to the base kinto.core Service class.
     Service.init_from_settings(settings)
 
+    # Use the project name as the main logger name (Logger field in MozLog).
+    JsonLogFormatter.init_from_settings(settings)
```
Natim: If it is the default, why do we need to add it in the config?

leplatrem: I didn't understand this question

PR #1366, opened 2017-11-09
Fix pyramid requirement in setup.py to match #1355

glasserc:
I believe this is why kinto.js builds are failing against current master.


glasserc:
Could I get a rereview now that I had to commit to >1.9?

PR #1378, opened 2017-11-16
Preparing release 7.6.0

Natim:


CHANGELOG.rst
```diff
@@ -3,7 +3,7 @@ Changelog
 
 This document describes changes between each past release.
 
-7.6.0 (unreleased)
+7.6.0 (2017-11-16)
```
leplatrem: The protocol part is not detailed enough, we have to make explicit (and in api/docs also) that something changes:
* if a record is pushed with an older timestamp, the collection timestamps is not dumped anymore

requirements.txt
```diff
@@ -6,6 +6,7 @@ colander==1.4
 colorama==0.3.9
 cornice==3.0.0
 cornice-swagger==0.5.2
+dockerflow==2017.5.0
```
leplatrem: this won't work I think, since we need https://github.com/mozilla-services/python-dockerflow/pull/21

CHANGELOG.rst
Natim: Is this good enough? https://github.com/Kinto/kinto/blob/master/docs/api/index.rst#118-2017-11-16

leplatrem: > Fix raise conditions regarding the collection_timestamp. As a side-effect when adding a record in the past, the collection_timestamp is not bumped anymore.

It was not about race conditions, but bottleneck. Just keep what you wrote in the README , it's enough/great

PR #1386, opened 2017-11-21
Postgresql: merge deleted and records tables

leplatrem:
- [x] fix tests
- [x] lock during migration
- [x] thorough tests (load, migration...)

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -357,10 +357,10 @@ class Storage(StorageBase):
                 from_epoch(:last_modified),
                 FALSE)
         ON CONFLICT (id, parent_id, collection_id) DO UPDATE
-            SET data=(:data)::JSONB,
-                deleted=FALSE,
-                last_modified = GREATEST(from_epoch(:last_modified),
-                                         EXCLUDED.last_modified)
+        SET data = (:data)::JSONB,
+            deleted = FALSE,
+            last_modified = GREATEST(from_epoch(:last_modified),
+                                     EXCLUDED.last_modified)
         RETURNING as_epoch(last_modified) AS last_modified;
         """
         placeholders = dict(object_id=object_id,
@@ -385,10 +385,13 @@ class Storage(StorageBase):
         if with_deleted:
             query = """
             UPDATE records
-               SET deleted=TRUE, data=(:deleted_data)::JSONB, last_modified =from_epoch(:last_modified)
+               SET deleted=TRUE,
+                   data=(:deleted_data)::JSONB,
+                   last_modified=from_epoch(:last_modified)
              WHERE id = :object_id
                AND parent_id = :parent_id
                AND collection_id = :collection_id
+               AND deleted = FALSE
             RETURNING as_epoch(last_modified) AS last_modified;
             """
         else:
@@ -397,6 +400,7 @@ class Storage(StorageBase):
             WHERE id = :object_id
                AND parent_id = :parent_id
                AND collection_id = :collection_id
+            RETURNING as_epoch(last_modified) AS last_modified;
             """
         deleted_data = self.json.dumps(dict([(deleted_field, True)]))
         placeholders = dict(object_id=object_id,
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -285,15 +285,17 @@ class Storage(StorageBase):
         safe_holders = {}
 
         if ignore_conflict:
-            # We use DO UPDATE so that the RETURNING clause works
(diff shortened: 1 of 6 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -395,32 +384,26 @@ def delete(self, collection_id, parent_id, object_id,
...
-                WHERE id = :object_id
-                  AND parent_id = :parent_id
-                  AND collection_id = :collection_id
-                RETURNING id
-            )
-            INSERT INTO deleted (id, parent_id, collection_id, last_modified)
-            SELECT id, :parent_id, :collection_id, from_epoch(:last_modified)
-              FROM deleted_record
+            UPDATE records
+               SET deleted=TRUE, data=(:deleted_data)::JSONB, last_modified =from_epoch(:last_modified)
```
glasserc: Weird spacing around `last_modified`.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -395,32 +384,26 @@ def delete(self, collection_id, parent_id, object_id,
...
-                DELETE
-                FROM records
-                WHERE id = :object_id
-                  AND parent_id = :parent_id
-                  AND collection_id = :collection_id
-                RETURNING as_epoch(last_modified) AS last_modified;
+            DELETE FROM records
+            WHERE id = :object_id
+               AND parent_id = :parent_id
+               AND collection_id = :collection_id
```
glasserc: Missing semicolon?

glasserc:
Should we maybe try to add a statement to the migration to handle cases where the record was both in `deleted` and `records`?

Natim:
> Should we maybe try to add a statement to the migration to handle cases where the record was both in `deleted` and `records`?

Yes I think we should

leplatrem:
> Should we maybe try to add a statement to the migration to handle cases where the record was both in deleted and records?

It is listed as a task as `Thorough migration tests`

leplatrem:
@glasserc The tests are green :) If you can, please add a dedicated storage migration test 🙏

kinto/core/storage/postgresql/__init__.py
```diff
@@ -443,29 +431,24 @@ def delete_all(self, collection_id, parent_id, filters=None,
...
+            WITH matching_records AS (
+                SELECT id, parent_id, collection_id
+                    FROM records
+                    WHERE {parent_id_filter}
+                          {collection_id_filter}
+                          AND deleted = FALSE
+                          {conditions_filter}
+                          {pagination_rules}
+                    {sorting}
+                    LIMIT :pagination_limit
```
glasserc: I think we should consider putting a `FOR UPDATE` here, although I don't really have a strong justification for why. It seems like two different queries could select the same not-yet-deleted rows and delete them simultaneously, although they might want to each delete 10. (But this might be a bigger topic than we want to address in this PR; see [this SO discussion](https://dba.stackexchange.com/questions/69471/postgres-update-limit-1)).

kinto/core/storage/postgresql/migrations/migration_017_018.sql
```diff
@@ -0,0 +1,28 @@
...
+
+INSERT INTO records (id, parent_id, collection_id, data, last_modified, deleted)
+  SELECT id, parent_id, collection_id, '{"deleted": true}'::JSONB, last_modified, TRUE
+  FROM deleted
+-- Because of Bug Kinto/kinto#1375, some tombstones may exist.
+  ON CONFLICT (id, parent_id, collection_id) DO UPDATE
+  SET data = '{"deleted": true}'::JSONB,
+      deleted = TRUE,
+      last_modified = EXCLUDED.last_modified;
+COMMIT WORK;
```
glasserc: This prefers to keep the tombstone. Maybe we should prefer the existing record by running a `DELETE` query first?

tests/core/test_storage_migrations.py
```diff
@@ -52,6 +52,13 @@ def _delete_everything(self):
         with self.storage.client.connect() as conn:
             conn.execute(q)
 
+    def _load_schema(self, filepath):
+        with self.storage.client.connect() as conn:
+            here = os.path.abspath(os.path.dirname(__file__))
+            with open(os.path.join(here, filepath)) as f:
+                old_schema = f.read()
+            conn.execute(old_schema)
```
glasserc: Nice cleanup 👍

glasserc:
Speaking about this more, this still leaves the TOCTOU whole in `create` described by #1388, but we think that's essentially solveable and will go ahead merging/starting the deploy process moving for this PR. I'll take over to write a PR for #1388.

leplatrem:
Tests are green, #1402 makes `py35-raw` fail.

Merging 🥇

PR #1404, opened 2017-11-28
Prepare 7.6.2

glasserc:


leplatrem:
I would have emphasized the schema change, with a special note about running the `migrate` command no?

glasserc:
No problem, I was just trying to model it off of other existing changelog entries and didn't see anything that looked relevant.

Issue #1376, opened 2017-11-15
duplicate key value violates unique constraint "records_pkey"


glasserc:
https://sentry.prod.mozaws.net/operations/kinto-webextensions-prod/issues/689286/

N.B. Not exactly the same as #1375, although perhaps it's related if we're replaying requests or something.

Natim:
This is because create is called with `ignore_conflict=False` somewhere which is the default.

Natim:
`resource_create_object` for the default_bucket should handle this kind of errors.

glasserc:
This should be fixed in 8.0.0.

PR #1406, opened 2017-11-29
Fix toctou in create

### LINKED ISSUE #1401: ignore_conflict option for create() is not used anywhere
In Kinto or in any of the plugins I checked.

I propose we get rid of it

Ref #1122

glasserc:
Fixes #1401, #1376.

- (n/a) Add documentation.
- (n/a) Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -285,7 +285,8 @@ class Storage(StorageBase):
             WHERE records.deleted = TRUE
             RETURNING id, data, last_modified
         )
-        SELECT id, data, as_epoch(last_modified) AS last_modified, TRUE AS inserted FROM create_record
+        SELECT id, data, as_epoch(last_modified) AS last_modified, TRUE AS inserted
+            FROM create_record
         UNION ALL
         SELECT id, data, as_epoch(last_modified) AS last_modified, FALSE AS inserted FROM records
         WHERE id = :object_id AND parent_id = :parent_id AND collection_id = :collection_id
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -269,10 +269,10 @@ class Storage(StorageBase):
         # https://stackoverflow.com/questions/15939902/is-select-or-insert-in-a-function-prone-to-race-conditions/15950324#15950324
         # a WHERE clause in the DO UPDATE will lock the conflicting
         # row whether it is true or not, so the subsequent SELECT is
-        # safe. We add a dummy "inserted" field to know whether we
+        # safe. We add a constant "inserted" field to know whether we
         # need to throw or not.
         query = """
-        WITH ins AS (
+        WITH create_record AS (
             INSERT INTO records (id, parent_id, collection_id, data, last_modified, deleted)
             VALUES (:object_id, :parent_id,
                     :collection_id, (:data)::JSONB,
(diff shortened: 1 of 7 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -252,57 +252,47 @@ def collection_timestamp(self, collection_id, parent_id, auth=None):
...
-        RETURNING id, as_epoch(last_modified) AS last_modified;
-        """
-
-        safe_holders = {}
-
-        if ignore_conflict:
-            # If we ignore conflict, then we do not touch the existing data.
-            # Unless if the conflict comes from a tombstone.
-            safe_holders['on_conflict'] = """
+        WITH ins AS (
```
leplatrem: nit: `ins` → `create_record`

kinto/core/storage/postgresql/__init__.py
```diff
@@ -252,57 +252,47 @@ def collection_timestamp(self, collection_id, parent_id, auth=None):
...
         query_record.pop(modified_field, None)
 
+        # If there is a record in the table and it is deleted = TRUE,
+        # we want to replace it. Otherwise, we want to do nothing and
+        # throw a UnicityError. Per
+        # https://stackoverflow.com/questions/15939902/is-select-or-insert-in-a-function-prone-to-race-conditions/15950324#15950324
+        # a WHERE clause in the DO UPDATE will lock the conflicting
+        # row whether it is true or not, so the subsequent SELECT is
+        # safe. We add a dummy "inserted" field to know whether we
+        # need to throw or not.
```
leplatrem: Why `dummy`?

kinto/core/storage/postgresql/__init__.py
```diff
@@ -252,57 +252,47 @@ def collection_timestamp(self, collection_id, parent_id, auth=None):
...
-            """
+            WHERE records.deleted = TRUE
+            RETURNING id, data, last_modified
+        )
+        SELECT id, data, as_epoch(last_modified) AS last_modified, TRUE AS inserted FROM ins
+        UNION ALL
+        SELECT id, data, as_epoch(last_modified) AS last_modified, FALSE AS inserted FROM records
+        WHERE id = :object_id AND parent_id = :parent_id AND collection_id = :collection_id
+        LIMIT 1;
+        """
```

kinto/core/storage/postgresql/__init__.py
```diff
@@ -312,6 +302,12 @@ def create(self, collection_id, parent_id, record, id_generator=None,
             result = conn.execute(query % safe_holders, placeholders)
             inserted = result.fetchone()
 
+        if not inserted['inserted']:
+            record = inserted['data']
+            record[id_field] = inserted['id']
+            record[modified_field] = inserted['last_modified']
+            raise exceptions.UnicityError(id_field, record)
```

leplatrem:
I don't know if we want to tackle this here (or if we track it in another issue), but in the `resource/__init__` there is also a very ugly `storage.get() + if + storage.create()` code, whereas we could probably catch `UnicityError`

glasserc:
I'm not actually sure how to fix the TOCTOU at the resource level. The reason we do a get first is to enforce 412 concurrency headers. Would we have to do a SELECT FOR UPDATE? If so, that would probably require another break in compatibility with the storage backend, in which case yes maybe we should do it in this PR.

glasserc:
N.B. the TOCTOU bug at the resource level affects `put`, `patch`, `delete`, and `collection_post`.

Another approach might be to have the `INSERT`/`UPDATE`/`DELETE` queries return any replaced rows, and do the 412 afterwards, rolling back if necessary??

glasserc:
Or another option is to push the 412 conditions into the SQL layer somehow. This would also be a breaking change...

glasserc:
I'm going to merge this for now and punt on the TOCTOU at the resource level, which is now #1407. I'll cut a release and at least we can cut down on the Sentry errors..

PR #1413, opened 2017-12-06
[Account plugin] add cache to authentication method

Stanley:
When debugging my application, I've noticed very poor performance when logged in using account plugin. It turns out that the biggest bottle neck is bcrypt and its `hashpw` method (https://github.com/Kinto/kinto/blob/master/kinto/plugins/accounts/authentication.py#L18). Making sure it is executed no more than once per request drastically reduces response times.

### Before:
```
> ab -n 100 -A admin:admin 127.0.0.1:8888/v1/
This is ApacheBench, Version 2.3 <$Revision: 1807734 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient).....done


Server Software:        waitress
Server Hostname:        127.0.0.1
Server Port:            8888

Document Path:          /v1/
Document Length:        697 bytes

Concurrency Level:      1
Time taken for tests:   39.127 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      95200 bytes
HTML transferred:       69700 bytes
Requests per second:    2.56 [#/sec] (mean)
Time per request:       391.272 [ms] (mean)
Time per request:       391.272 [ms] (mean, across all concurrent requests)
Transfer rate:          2.38 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:   387  391   4.8    390     427
Waiting:      386  391   4.8    390     427
Total:        387  391   4.8    390     427

Percentage of the requests served within a certain time (ms)
  50%    390
  66%    391
  75%    392
  80%    393
  90%    395
  95%    399
  98%    402
  99%    427
 100%    427 (longest request)
```
image

### After:
```
> ab -n 100 -A admin:admin 127.0.0.1:8888/v1/
This is ApacheBench, Version 2.3 <$Revision: 1807734 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 127.0.0.1 (be patient).....done


Server Software:        waitress
Server Hostname:        127.0.0.1
Server Port:            8888

Document Path:          /v1/
Document Length:        697 bytes

Concurrency Level:      1
Time taken for tests:   0.392 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      95200 bytes
HTML transferred:       69700 bytes
Requests per second:    255.05 [#/sec] (mean)
Time per request:       3.921 [ms] (mean)
Time per request:       3.921 [ms] (mean, across all concurrent requests)
Transfer rate:          237.11 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:     2    4   4.1      3      43
Waiting:        2    4   4.1      3      43
Total:          2    4   4.1      3      43

Percentage of the requests served within a certain time (ms)
  50%      3
  66%      4
  75%      4
  80%      4
  90%      5
  95%      5
  98%      6
  99%     43
 100%     43 (longest request)
```

screenshot_20171206_084711

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.


Natim:
@Stanley thank you for investigating this. Are you willing to take over and fix the tests (basically remove the cache key when changing the password or deleting the account) or should we do it?

Stanley:
Sure, I'll do it :)

Natim:
Thanks, let me know if you need help or if you want to pair on it. I am really exited by the feature to land 👍

Stanley:
Done. All tests are passing :) Anything else?

Natim:
Can you add a line in the CHANGELOG?

Stanley:
Done :)

kinto/plugins/accounts/authentication.py
```diff
@@ -15,8 +30,12 @@ def account_check(username, password, request):
 
     hashed = existing['password'].encode(encoding='utf-8')
     pwd_str = password.encode(encoding='utf-8')
-    if hashed == bcrypt.hashpw(pwd_str, hashed):
-        return True  # Match! Return anything but None.
+    # Check if password is valid (it is a very expensive computation)
+    if bcrypt.checkpw(pwd_str, hashed):  # Match!
+        # Remember result so we don't have to calculate bcrypt hash next time.
+        cache.set(cache_key, "1", ttl=30)  # Time to live: 30 seconds
```
gabisurita: I think it would be great to expose the TTL in settings.

Natim:
> if an attacker gain access the cache layer and the HMAC secret, than we can have a password leakage.

No because it is a hmac and it is not reversible, however we can rotate the hmac key with this plugin since we can wipe the cache whenever we want.

kinto/plugins/accounts/authentication.py
Natim: Can you think of a use case for that?

Natim: Can you think of a use case for that?

gabisurita:
@Natim Sorry, I wasn't clear at this point. We're not leaking raw passwords but since we have a constant key/salt, one can brute force it with a dictionary attack over the hash. The problem isn't about being reversible, but being deterministic. I agree having a rotating seed and wiping out the cache periodically is a nice way to go.

kinto/plugins/accounts/authentication.py
gabisurita: One can increase this value to have more cache hits (e.g across periodic requests). Refreshing the TTL at each request would also be a great feature IMO, but this can be done in another PR.

Stanley: Thought about it too, but couldn't figure out how to implement cache that would expire after one request. Any ideas?

Natim:
> The problem isn't about being reversible, but being deterministic.

Oh right, we should use the username hash of the username and then have a random salt and a saltedhashed password as the value.

Thanks for detecting this!

almet:
Probably this hasn't survived the end of year crazyness :-) Here's a little bump -- what's missing on this to go ahead?

Natim:
I guess we should merge it !!

leplatrem:
We can merge once the following is done:

- [ ] unit tests that assert a cache key is set (or bcrypt is not called on second attempt)
- [ ] unit tests that assert what happens when user/pass is changed/deleted
- [ ] the TTL value in settings
- [ ] a cache refresh on each request (basically call `cache.set()`)

@Stanley are you still interested in working on this and go through those last steps? Let us know, maybe somebody can help and take over!

Thanks!

kinto/plugins/accounts/authentication.py
```diff
@@ -1,10 +1,29 @@
...
 from kinto.core.storage import exceptions as storage_exceptions
 
+from . import ACCOUNT_CACHE_KEY
+
 
 def account_check(username, password, request):
+    settings = request.registry.settings
+    hmac_secret = settings['userid_hmac_secret']
+    cache_key = utils.hmac_digest(hmac_secret, ACCOUNT_CACHE_KEY.format(username))
+    cache_ttl = settings.get('account_cache_ttl_seconds', 30)
```
leplatrem: This setting should be documented and could be added to settings template IMO

kinto/plugins/accounts/authentication.py
```diff
@@ -1,10 +1,29 @@
...
+
+    # Check cache to see whether somebody has recently logged in with the same
+    # username and password.
+    cache = request.registry.cache
+    cache_result = cache.get(cache_key)
+
+    # Username and password is correct. No need to compare hashes
```
leplatrem: I don't think `is correct` applies here. It would probably be `was already verified`

kinto/plugins/accounts/authentication.py
```diff
@@ -1,10 +1,29 @@
...
+    if cache_result:
+        cache.expire(cache_key, cache_ttl)
```
leplatrem: Add a comment to say we refresh the cache ttl on each request?

kinto/plugins/accounts/views.py
```diff
@@ -127,3 +132,19 @@ def process_record(self, new, old=None):
...
+# Clear cache on account change
+@subscriber(ResourceChanged,
+            for_resources=('account',),
+            for_actions=(ACTIONS.UPDATE, ACTIONS.DELETE))
+def on_account_changed(event):
+    request = event.request
+    cache = request.registry.cache
+    settings = request.registry.settings
+    # Extract username and password from current user
+    username, password = extract_http_basic_credentials(request)
```
leplatrem: Shouldn't it be `request.matchdict['id']` instead ? Since we make sure they match in the resource code?

tests/plugins/test_accounts.py
```diff
@@ -99,6 +101,48 @@ def test_cannot_create_other_account_if_authenticated(self):
...
+        with mock.patch('kinto.plugins.accounts.authentication.bcrypt') as mocked_bcrypt:
+            resp = self.app.get('/', headers=get_user_headers('me', 'bouh'))
+            assert resp.json['user']['id'] == 'account:me'
+
+            resp = self.app.get('/', headers=get_user_headers('me', 'bouh'))
+            assert resp.json['user']['id'] == 'account:me'
+
+            mocked_bcrypt.checkpw.assert_called_once()
+
+    def test_authentication_check_bcrypt_again_if_password_changes(self):
```
leplatrem: nit: checks

tests/plugins/test_accounts.py
```diff
@@ -99,6 +101,48 @@ def test_cannot_create_other_account_if_authenticated(self):
...
+
+            assert mocked_bcrypt.checkpw.call_count == 2
+
+    def test_authentication_refresh_the_cache_each_time_we_authenticate(self):
+        self.app.post_json('/accounts', {'data': {'id': 'me', 'password': 'bouh'}},
+                           status=201)
+        resp = self.app.get('/', headers=get_user_headers('me', 'bouh'))
+        assert resp.json['user']['id'] == 'account:me'
+
+        time.sleep(2)
```
leplatrem: We should not slow the tests by two seconds here, we can do better like call `.cache.expire(cache_key, 1)` first and then make sure it's `>1` or whatever

CHANGELOG.rst
```diff
@@ -9,6 +9,7 @@ This document describes changes between each past release.
 **New features**
 
 - Add Openid connect support (#939, #1425). See `demo <https://github.com/leplatrem/kinto-oidc-demo>`_
+- Account plugin now caches authentication method (#1413)
```
leplatrem: Method? Hmm, I would say `verification`

kinto/plugins/accounts/authentication.py
Natim: I guess you mean that https://github.com/Kinto/kinto/pull/1413/files#diff-2193261fa1533ff793bf90b447434bc8R32 is not sufficient?

tests/plugins/test_accounts.py
Natim: I don't understand how setting cache.expire helps here can you elaborate?

Natim: Ok got it

docs/api/1.x/accounts.rst
```diff
@@ -28,6 +28,9 @@ Add the following settings to the ``.ini`` file:
     # Allow anyone to create accounts.
     kinto.account_create_principals = system.Everyone
 
+    # Set the session time to live in seconds for the session
```
leplatrem: nit: `session time` and `for the session` does not sound super clear in the same sentence :)

Natim:
Thanks a lot @Stanley for starting this !!!

PR #1419, opened 2017-12-08
Don't purge_deleted

glasserc:
Perhaps the reason we're seeing increased CPU load is because the
purge_deleted call, which used to do a table scan over a relatively
small table, is now over the entire records table. This probably
doubles the amount of table scans we perform.

Let's disable this purge_deleted for a while. Tombstones will stay
tombstones for now. This won't be a problem for kintowe because
kintowe never actually generates tombstones (even deleted records are
encrypted as "live" records).

### DIFF
--- a/kinto/views/buckets.py
+++ b/kinto/views/buckets.py
@@ -35,7 +35,7 @@ def on_buckets_deleted(event):
                                collection_id=None,
                                with_deleted=False)
             # Remove remaining tombstones too.
-            storage.purge_deleted(parent_id=pattern,
-                                  collection_id=None)
+            # storage.purge_deleted(parent_id=pattern,
+            #                       collection_id=None)
             # Remove related permissions
             permission.delete_object_permissions(pattern)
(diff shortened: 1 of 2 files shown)


Natim:
Yes we could also imagine a `kinto purge-deleted` command.

glasserc:
I'm not really sure how to make a change that would only be kintowe specific, and I'm hesitant to try to devote effort to fixing the problem without being sure that it is, in fact, the problem. Assuming purge_deleted is the actual problem, how would we fix it? Is there a way we can determine that it is the problem for kintowe without changing the kinto code in this invasive way?

Regarding `with_deleted`, I think that should be `False`, as that ensures that deletes the records without leaving a tombstone. It is already `with_deleted=False` in the surrounding code.

Natim:
> Regarding with_deleted, I think that should be False, as that ensures that deletes the records without leaving a tombstone. It is already with_deleted=False in the surrounding code.

Yes sorry you are right.

glasserc:
Speaking about this on a voice call, Rémy and I agreed that this hack could be improved by making it conditional on a setting named something like `experimental_disable_purge_deleted`, so as not to afflict random users who installed this version by accident. Making this change now.

glasserc:
@Natim re-r?

PR #1424, opened 2017-12-16
add backoff percentage config setting for client

### LINKED ISSUE #890: Backoff a percentage of clients
Add a _percentage_ setting for backoff. If specified, then send backoff header only if `random number < percentage`.

AnthonyGaruccio:
Fixes #890 

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

Hi! This is my first contribution, so I welcome your feedback and would be happy to make any changes you suggest.

### DIFF
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -3,6 +3,7 @@ import re
 import warnings
 from datetime import datetime
 from dateutil import parser as dateparser
+import random
 
 from pyramid.events import NewRequest, NewResponse
 from pyramid.exceptions import ConfigurationError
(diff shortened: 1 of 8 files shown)


kinto/core/initialization.py
```diff
@@ -132,7 +133,12 @@ def on_new_response(event):
         # Add backoff in response headers.
         backoff = config.registry.settings['backoff']
         if backoff is not None:
-            event.response.headers['Backoff'] = str(backoff)
+            backoff_percentage = config.registry.settings['backoff_percentage']
```
Natim: Do we need to convert the value to integer?

Natim:
Hi @AnthonyGaruccio thank you for your contribution.

Can you add the changelog entry, your name in the contributor file and a line in the config documentation about that?

kinto/core/initialization.py
AnthonyGaruccio: I don't believe so, otherwise
```python
if random.random() > (backoff_percentage / 100.0)
```
would have raised a `TypeError` exception when the tests ran

AnthonyGaruccio:
Hi @Natim thanks for the welcome!  I can update the changelog, contributor file, and documentation tonight.

kinto/core/initialization.py
Natim: Well in your test you test with a mock of a integer, the question is did you try it with an actually config file?

leplatrem: I agree with @Natim, an explicit cast to int or float would not harm here

leplatrem:
Awesome! Thanks \o/

Issue #508, opened 2016-03-15
The Github oauth tutorial is not OpenID Connect compliant.


Natim:
As @jmandel described in [his comment on the tutorial](http://kinto.readthedocs.org/en/latest/tutorials/authentication-github.html#comment-2568178329)

> In order to use GitHub sign-in safely, you need to ensure the token you pass to Kinto API endpoints is intended for Kinto. In an ideal world, GitHub would help with this problem by creating a signed token designed for the application to consume. This is what the ID token accomplishes in OpenID connect. Since GitHub does not do this for you, you would need to do it yourself.

Using Github Bearer Token to authenticate directly to Kinto without any specific scoping of the Bearer Token means that any application having a Github Bearer Token for the user can manage, read, add and delete her Kinto data. This is not acceptable for production use.

This remembers me a discussion we had a month ago in London with @michielbdejong where he explained me that RemoteStorage embedded an OpenID Connect compatible server in charge of generating RemoteStorage scoped specific Bearer Token.

How should we handle this for Kinto, what is the plan forward to have a better OAuth2 integration.

Thoughts?

jmandel:
Fundamentally the challenge is that OAuth is designed as an authorization protocol. You can build an authentication protocol on top, but you have to be careful to get the details right (like OIDC). Forgot details see http://oauth.net/articles/authentication/ by @jricher.

leplatrem:
FWIW @ptgamr did https://github.com/ptgamr/kinto-github-token ;)

PR #1425, opened 2017-12-17
Openid connect support (fixes #939)

### LINKED ISSUE #939: Plug OpenID Connect authentication policy.
- LDAP is going to be replaced by auth.mozilla.auth0.com
- FxA is now supporting OpenId Connect too.

It looks like a standard for OpenID providers and it would make it configurable for a wide range of identity providers.

leplatrem:
Fixes #939 

- [x] Add docs
- [x] Move to plugins folder
- [x] Choose appropriate settings name
- [x] Add tests
- [x] Publish demo somewhere?
- [x] Create issue for kinto-admin support 
- [ ] Add mention in Github tutorial ? ref #508   
- [x] Polish OpenAPI spec
- [x] Reject scope that does not contain `email` if the configured userid_field is email
- [ ] Implement the scope filtering as we did for kinto-fxa

We could make this piece of code work with both Auth0 and Firefox Accounts \o/  

@rfk f?

### DIFF
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -479,13 +479,14 @@ In order to replace it by another one:
     multiauth.policies = basicauth
     multiauth.policy.basicauth.use = myproject.authn.BasicAuthPolicy
 
+.. _settings-openid:
 
 OpenID Connect
 ::::::::::::::
 
-First of all, you must find an Identity Provider. Google Identity Platform for example, but it may also be Auth0, Microsoft, Yahoo, Paypal, Bitbucket, Ebay, Salesforce, ... or which ever platform that publishes its discovery metadata as JSON.
+First of all, you must find an Identity Provider. Google Identity Platform for example, but it may also be Auth0, Microsoft, Yahoo, Paypal, Bitbucket, Ebay, Salesforce, ... or whichever platform that publishes its discovery metadata as JSON.
 
-Enable the OpenID authentication policy in Kinto settings like below. The ``oidc`` is arbitrary but will become the user ID prefix (e.g. ``google:someuser@gmail.com``)
+Enable the OpenID authentication policy in Kinto settings like below. The ``google`` name is arbitrary but will become the user ID prefix (e.g. ``google:someuser@gmail.com``)
 
 .. code-block:: ini
 
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -509,13 +509,13 @@ At this point, Kinto should be able to start and OpenID Authentication should wo
 .. code-block:: ini
 
     # User ID field name (Default: `sub`)
-    multiauth.policy.oidc.userid_field = email
+    multiauth.policy.google.userid_field = email
     # Authorization header prefix (Default: `Bearer`)
-    multiauth.policy.oidc.header_type = Bearer+OIDC
+    multiauth.policy.google.header_type = Bearer+OIDC
     # User information cache expiration (Default: 1 day)
-    multiauth.policy.oidc.verification_ttl_seconds = 86400
+    multiauth.policy.google.verification_ttl_seconds = 86400
     # Authentication state cache duration (Default: 1 hour)
-    multiauth.policy.oidc.state_ttl_seconds = 3600
+    multiauth.policy.google.state_ttl_seconds = 3600
 
 
 Custom Authentication
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -480,6 +480,44 @@ In order to replace it by another one:
     multiauth.policy.basicauth.use = myproject.authn.BasicAuthPolicy
 
 
+OpenID Connect
+::::::::::::::
+
+First of all, you must find an Identity Provider. Google Identity Platform for example, but it may also be Auth0, Microsoft, Yahoo, Paypal, Bitbucket, Ebay, Salesforce, ... or which ever platform that publishes its discovery metadata as JSON.
+
+Enable the OpenID authentication policy in Kinto settings like below. The ``oidc`` is arbitrary but will become the user ID prefix (e.g. ``google:someuser@gmail.com``)
+
+.. code-block:: ini
+
+    kinto.includes = kinto.plugins.openid
+
+    multiauth.policies = google
+    multiauth.policy.google.use = kinto.plugins.openid.OpenIDConnectPolicy
+
+Based on the information provided by the Identity Provider configure the ``issuer`` and ``client_id``. Potentially you may have a ``client_secret`` depending on how the application was setup. For example, Auth0 has a *Frontend app* option that does not require any client secret.
+
+.. code-block:: ini
+
+    multiauth.policy.google.issuer = https://accounts.google.com
+    multiauth.policy.google.client_id = 42XXXX365001.apps.googleusercontent.com
+    multiauth.policy.google.client_secret = UAlL-054uyh5in4b6u8jhg5o3hnj
+
+At this point, Kinto should be able to start and OpenID Authentication should work as described in the :ref:`API docs <authentication-openid>`.
+
+**Advanced settings**
+
+.. code-block:: ini
+
+    # User ID field name (Default: `sub`)
+    multiauth.policy.oidc.userid_field = email
+    # Authorization header prefix (Default: `Bearer`)
+    multiauth.policy.oidc.header_type = Bearer+OIDC
+    # User information cache expiration (Default: 1 day)
+    multiauth.policy.oidc.verification_ttl_seconds = 86400
+    # Authentication state cache duration (Default: 1 hour)
+    multiauth.policy.oidc.state_ttl_seconds = 3600
+
+
 Custom Authentication
 :::::::::::::::::::::
 
--- a/kinto/core/authentication.py
+++ b/kinto/core/authentication.py
@@ -1,7 +1,14 @@
+from jose import jwt
 from pyramid import authentication as base_auth
+from pyramid.interfaces import IAuthenticationPolicy
+from zope.interface import implementer
+
+import requests
 
 from kinto.core import utils
 from kinto.core.openapi import OpenAPI
+
+from kinto.core import logger
 
 
 class BasicAuthAuthenticationPolicy(base_auth.BasicAuthAuthenticationPolicy):
@@ -35,6 +42,93 @@ class BasicAuthAuthenticationPolicy(base_auth.BasicAuthAuthenticationPolicy):
             credentials = '{}:{}'.format(*credentials)
             userid = utils.hmac_digest(hmac_secret, credentials)
             return userid
+
+
+@implementer(IAuthenticationPolicy)
+class OpenIDConnectPolicy(base_auth.CallbackAuthenticationPolicy):
+    def __init__(self, realm='Realm'):
+        self.realm = realm
+        self._cache = None
+
+        self._jwt_keys = None
+
+    def unauthenticated_userid(self, request):
+        """Return the FxA userid or ``None`` if token could not be verified.
+        """
+        authorization = request.headers.get('Authorization', '')
+        try:
+            authmeth, payload = authorization.split(' ', 1)
+        except ValueError:
+            return None
+
+        if authmeth.lower() != 'bearer+oidc':
+            return None
+
+        try:
+            # Bearer+OIDC id_token=jwt, access_token=bearer
+            parts = payload.split(',')
+            credentials = {k: v for k, v in [p.strip().split("=") for p in parts]}
+            id_token = credentials["id_token"]
+            access_token = credentials["access_token"]
+        except (ValueError, KeyError):
+            # Bearer+OIDC jwt
+            id_token = payload
+            access_token = None
+
+        return self._verify_token(request, id_token, access_token)
+
+    def forget(self, request):
+        """A no-op. Credentials are sent on every request.
+        Return WWW-Authenticate Realm header for Bearer token.
+        """
+        return [('WWW-Authenticate', 'Bearer+OIDC realm="%s"' % self.realm)]
+
+    def _verify_token(self, request, id_token, access_token):
+        # Verify signature
+        audience = request.registry.settings["oidc.audience"]
+        issuer = request.registry.settings["oidc.issuer_url"]
+
+        if self._jwt_keys is None:
+            resp = requests.get(issuer.rstrip("/") + "/.well-known/openid-configuration")
+            jwks_uri = resp.json()["jwks_uri"]
+            resp = requests.get(jwks_uri)
+            self._jwt_keys = resp.json()["keys"]
+
+        try:
+            unverified_header = jwt.get_unverified_header(id_token)
+        except jwt.JWTError:
+            logger.debug("Invalid header. Use an RS256 signed JWT Access Token")
+            return None
+
+        if unverified_header["alg"] != "RS256":
+            logger.debug("Invalid header. Use an RS256 signed JWT Access Token")
+            return None
+
+        rsa_key = {}
+        for key in self._jwt_keys:
+            if key["kid"] == unverified_header["kid"]:
+                rsa_key = utils.dict_subset(key, ("kty", "kid", "use", "n", "e"))
+                break
+        if not rsa_key:
+            logger.debug("Unable to find appropriate key")
+            return None
+        try:
+            options = None
+            if access_token is None:
+                options = {"verify_at_hash": False}
+
+            payload = jwt.decode(id_token, rsa_key, algorithms=["RS256"],
+                                 audience=audience, issuer=issuer,
+                                 options=options, access_token=access_token)
+            return payload["sub"]
+
+        except jwt.ExpiredSignatureError as e:
+            logger.debug("Token is expired: %s" % e)
+        except jwt.JWTClaimsError as e:
+            logger.debug("Incorrect claims, please check the audience and issuer: %s" % e)
+        except Exception as e:
+            logger.exception("Unable to parse token")
+        return None
 
 
 def includeme(config):
--- a/kinto/core/authentication.py
+++ b/kinto/core/authentication.py
@@ -1,14 +1,7 @@
-from jose import jwt
 from pyramid import authentication as base_auth
-from pyramid.interfaces import IAuthenticationPolicy
-from zope.interface import implementer
-
-import requests
 
 from kinto.core import utils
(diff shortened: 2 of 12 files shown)


rfk:
> We could make this piece of code work with both Auth0 and Firefox Accounts \o/

FWIW I'm a strong 👍  to this goal :-)

kinto/core/authentication.py
```diff
@@ -37,6 +44,93 @@ def unauthenticated_userid(self, request):
...
+        try:
+            authmeth, payload = authorization.split(' ', 1)
+        except ValueError:
+            return None
+
+        if authmeth.lower() != 'bearer+oidc':
+            return None
+
+        try:
+            # Bearer+OIDC id_token=jwt, access_token=bearer
```
rfk: Out of curiosity, is this approach based on some prior art, or your own scheme?  I'm not sure how well it will interact with refresh tokens, which would allow you to generate a fresh access_token that's not bound to the id_token.

Natim: It is our own scheme. We asked ourselves a lot of question about that.

Looking at the Auth0 documentation is seems that the only way to verify an access_token is to use the id_token. (at_hash value) but it doesn't seems to be the way of doing it with Firefox Account where we would most likely use the `/verify` endpoint.

Natim: My understanding of OAuth was that sending the access_token should be enough but then how do you validate it with Auth0? How does services usually work with id_token and access_token ? Do they send the id_token?

rfk: > My understanding of OAuth was that sending the access_token should be enough

That's my understanding as well.  The id_token is a point-in-time confirmation that we successfully *authenticated*  a particular user, and the access_token is an ongoing grant of *authorization* to do a particular thing.  I don't know much about Auth0, but having to send both on every request doesn't mesh with my expectations of OIDC.

rfk: > Looking at the Auth0 documentation is seems that the only way to verify an access_token is to use the id_token

Can you please drop a link to this for my reference?

Natim: This was it: https://auth0.com/docs/api-auth/tutorials/verify-access-token

But this is what I also found: https://auth0.com/docs/api-auth/intro#calling-apis-with-access-tokens

Natim: So maybe the access_token is also a JWT that self validates?

Natim: The problem is that if you look at this table: https://auth0.com/docs/api-auth/intro#how-to-use-the-new-flows

You see that tokeninfo is disabled...

Natim: Should we rely on userinfo call instead?

leplatrem:
@cfguimaraes in order to try this branch with Auth0, you can use these settings:

```ini
multiauth.policies = oidc
multiauth.policy.oidc.use = kinto.core.authentication.OpenIDConnectPolicy
oidc.issuer_url = https://{yours}.auth0.com
oidc.audience = {api-id}
```

cfguimaraes:
@leplatrem thank you for the mention, I will try this on Codenvy

almet:
This seems great, thanks for working on it! Being able to use Openid connect will be a lot useful, so we can ditch basic auth default mechanism :-)

kinto/core/authentication.py
```diff
@@ -37,6 +44,93 @@ def unauthenticated_userid(self, request):
...
+            parts = payload.split(',')
+            credentials = {k: v for k, v in [p.strip().split("=") for p in parts]}
+            id_token = credentials["id_token"]
+            access_token = credentials["access_token"]
```
cfguimaraes: Are these lines mandatory in JWT from an OIDC provider?
I've created a non-interactive client on Auth0 dashboard, from the APIs gotten the JWT token using curl, however, I always got HTTP 401 Unauthorized from Kinto.
```
{
  "iss": "https://tenant.auth0.com/",
  "sub": "key@clients",
  "aud": "https://tenant.auth0.com/api/v2/",
  "iat": 1516993206,
  "exp": 1517079606,
  "azp": "gaJ9QUArBJTPpL76SYV8jDhqu9Cztq93",
  "gty": "client-credentials"
}
```
The code above is the Payload extracted from [JWT.io site](https://jwt.io/)

cfguimaraes: Accordingly to section 2 from [Open ID Core](http://openid.net/specs/openid-connect-core-1_0.html#IDToken) the keys required are all right.

cfguimaraes: My bad, I'm testing all day with a non-interactive client from Auth0, the purpose of that is to grant authorization to scripts, in a server-server communication channel.
Tomorrow I will plug Auth0 lock (login) and update this topic with discovers, I'm really happy with how simples this code are at the same time how great are thanks to OIDC discovery.
Thanks to @leplatrem too for spend resources on this code.

cfguimaraes: Status update:
1. First of all, I become testing this in an Ionic project, already configured with Auth0, but when trying to integrate with kinto.js I don't know why, but sadly doesn't work
2. Download quick start from Auth0's GitHub samples, inspecting the code on this branch I've seen that this is relying on python-jose for authentication, specifically to `decode` the `JWT`, also, @leplatrem  fixed an issue and make a PR to address an error on `at_hash`claim being required, but accordingly to OIDC that is not required at all
3. I've forked the @leplatrem repo from python-jose and build the branch with her contributions to generate a wheel (whl) and installed it with pip3, however, I'm always getting an error from Authentication.py 
`Incorrect claims, please check the audience and issuer: **at_hash claim missing from token**.`

I think that the local installation of python-jose (whl) is not being reflected on kinto source, can someone give me a help, I don't know what I have to do to make the decode function of python-jose work with enhanced code from @leplatrem in order to advance in these tests.

The good news is that after all, we already have a sample demonstrating how to integrate OIDC compliant like Auth0 to kinto.

kinto/core/authentication.py
```diff
@@ -37,6 +44,93 @@ def unauthenticated_userid(self, request):
...
+                break
+        if not rsa_key:
+            logger.debug("Unable to find appropriate key")
+            return None
+        try:
+            options = None
+            if access_token is None:
+                options = {"verify_at_hash": False}
+
+            payload = jwt.decode(id_token, rsa_key, algorithms=["RS256"],
```
cfguimaraes: I've logged the code and verified that the error `at_hash claim missing from token` is being raised from this function.
I cloned (https://github.com/leplatrem/python-jose/tree/75-at_hash-optional)[leplatrem python-jose] implementation that is supposed to address the issue in order to test, but I don't know if the changes on code is being reflected on kinto source

leplatrem: Sorry @cfguimaraes for the delay...

Basically you should be able to apply the fork with:
```
$ cd kinto/
$ source .venv/bin/activate
$ pip install -e <path-to-python-jose-fork>
```

But according to the code above if you provide a JWT but no access token, then this hash is not verified.

The status of this branch is still very rough, I want to land it ASAP, we'll need it in our production stacks at Mozilla :)

cfguimaraes:
Good news, working today in this feature I could connect with Auth0 OIDC Conformant JWT Token.

However, the code looks for a Bearer+OIDC Authorization Header, I couldn't find a place where this Header is described, only **Bearer** token.

The code in the branch doesn't seem to work properly, I've needed to fork python-jose implementation of the at_hash issue of leplatrem, after that all works fine.

I will clear the local code, and make new tests to ensure the python-jose code has a bug or not in the current version of Kinto.

In 4 hours I'll make a report.

cfguimaraes:
I've cleaned local code, make a pull and update the code to reflect the locations for the plugin and all work fine, without the code on python-jose.
Sadly I don't have many use cases to test, but Authentication is working fine, with a valid acess_token and Forbidden with an invalid JWT.

If there is something you'll want I do, I will be happy to help.

leplatrem:
Thanks @cfguimaraes for your detailed feedback!

Next step on my side is to write a little single page app demo (with Github auth or something) and update the docs / tutorial...

> However, the code looks for a Bearer+OIDC Authorization Header, I couldn't find a place where this Header is described, only Bearer token.

We invented it with @Natim :) Since Kinto can have several authentication policies enabled, with several ones that use bearer tokens, we wanted to have a special one. Maybe we could default it to just `Bearer` and make it optional in settings to customize it. If you thought that was confusing, it's a sign :) 

> Notes: 
> https://developer.github.com/apps/building-oauth-apps/

leplatrem:
I made some changes to this branch. The tests will fail. The `audience` setting was renamed to `client_id`.

I made it work with Google Accounts, this is the frontend demo that I wrote:
https://github.com/leplatrem/kinto-oidc-demo/

It's still Work-In-Progress though.

cfguimaraes:
I really think that a **Bearer** token as proposed is the best as default.
If one wants a different behavior that could configure that, I think that a tool has to be standard compatible, those that are exceptions have to write code for exception, IMO.

Natim:
> I really think that a Bearer token as proposed is the best as default.

To be honest I am not quite sure about that postulate. When using the bearer token with Kinto you still have a to build the Authorization header by hand.

If you know that Kinto expect a `Bearer+OIDC` realm then it doesn't change anything to write `Bearer+OIDC` rather than `Bearer`.

Plus this let you configure Kinto with multiple auth providers without messing around with configuration.

leplatrem:
> Plus this let you configure Kinto with multiple auth providers without messing around with configuration.

Honestly, having multiple authentication policies configured is not the most common use-case

kinto/plugins/openid/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+            authmeth, payload = authorization.split(' ', 1)
+        except ValueError:
+            return None
+
+        header_type = request.registry.settings.get("oidc.header_type", "bearer")
+        if authmeth.lower() != header_type:
+            return None
+
+        try:
+            # Bearer+OIDC id_token=jwt, access_token=bearer
```
Natim: Remove support for ID_Token (which is used by the frontend), on the backend we want to use access/bearer token

cfguimaraes:
@Natim, I like standard things, don't you think that being standard this tool will help much more developers at all?
[In OIDC Core, token response section says](http://openid.net/specs/openid-connect-core-1_0.html#TokenResponse)
> The OAuth 2.0 token_type response parameter value MUST be Bearer, as specified in OAuth 2.0 Bearer Token Usage [RFC6750], unless another Token Type has been negotiated with the Client
I think, having a non-standard Header like `Bearer+OIDC` will force 2 implementations, one for tools that accept content negotiation, and another for standard tools only. In the first case we will force developers to write and manage more code, to accept negotiation with Kinto endpoint, and to access other API's that are standard compatible **only** sending a `Bearer` jwt.
I as a developer don't like how this sounds, if I need a non-standard solution just implementing an interface and settings in kinto.ini (configuration file) will make me much happier.

Natim:
> unless another Token Type has been negotiated with the Client

That's my point.

> I think, having a non-standard Header like Bearer+OIDC will force 2 implementations

I don't think so.

> I as a developer don't like how this sounds, if I need a non-standard solution just implementing an interface and settings in kinto.ini (configuration file) will make me much happier.

I agree with you here.

cfguimaraes:
> I don't think so.

Could you say more about how you see this?
I still see 2 implementations, in this case, one for Bearer (standard), and another for any non-standard header (as any developer can extend kinto) like Bearer+OIDC as invented for @leplatrem in her code.

I've made a login in an Auth0 OIDC compliant client, sending a compliant acess_token **Authorization heared as a Bearer jwt** I've been Unauthorized, so I've to send a Kinto compliant `Bearer+OIDC` token or change the code to expect a `Bearer`.
I've made the first because I'm writing request in a REST plugin, but when I've made an app to make tests with Angular the tools I used sends standard `Bearer` token. I don't investigate how to extend the js lib I used to send a custom Header, but I really, don't wish to do that.
And is this I'm saying, if we'll deliver a built-in plugin with Kinto, this SHOULD be OIDC standard compliant, this way we can bring a more convenient tool for developers.
If one needs a different solution than that, this already will have to write custom code, so is acceptable.

What you think about this?

Natim:
Everything is the same (Bearer token, JWT) that only the Realm that has got a suffix.
You can use the same lib, you can use everything that already exists around OIDC grab the access_token and then when the clients call Kinto ti will use the proper OIDC plugin realm. It doesn't need another implementation.

kinto/plugins/openid/__init__.py
leplatrem: The thing is that we can have access tokens in JWT format. I haven't looked at those yet. 

Also, using id tokens saves a round-trip to the Identity Provider. With the inconvenient that their validity remains unquestioned until they expire.

cfguimaraes:
> Then maybe I am wrong. Can you show me how you use your plugin?

Of course, I can demonstrate.

Recently I cleaned local code that I changed for make tests.
I've put all apps (kinto, ng) running.
When I make a request with curl using an Authorization `Bearer+OIDC` as expected in the code, all works and I've been authenticated.
If I go to Angular app, authenticated to Auth0 (OIDC Conformant) and make a request using **angular2-jwt**, maintained by auth0, that put an Authorization `Bearer` token, the code is looking for another __AuthMethod__, I got a 401 and a header `Www-Authenticate | bearer+oidc realm="Realm"`

When I look for the www-authenticate definition I found in [mozilla developers network](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) 
> The HTTP WWW-Authenticate response header defines the authentication method that should be used to gain access to a resource.

Because this I'm saying, if we go with OIDC standard that says the `Bearer` is the default __transporter__ (if I coud say) for  JWT `acess_token` we can benefit to be standard compatible and write less code, the same time, other developers will work smooth.
However, if we choose another Authorization header that isn't a Bearer we'll deliver to third-party developers the responsibility to write code to manage custom authentication just for Kinto.

And as I said previously if one needs another Auth mechanism that is not OAuth, this one already will have to write custom code, almost easy for Kinto, just implementing an Interface and some change in configuration, and another in Client (s) code, is perfectly presumable, don't you think?

I don't have knowledge about Realm, but the definition I've found isn't the same I think I've understood of   @Natim written.

I've understood that Bearer+OIDC (Authentication Mechanism) is just a Realm.
In my understanding, **Bearer+OIDC** will be another Authentication Mechanism, like **Basic** is, and Realm is apart of that.

Natim:
Apparently angular2-jwt let you configure a headerPrefix here: https://github.com/auth0/angular2-jwt#configuration-options

But now I understand the confusion. Thanks for letting us know.

leplatrem:
I reworked the settings part, it now supports multiple OpenID policies, and I updated the demo.
I drafted some docs as well.

At this point I would take as much feedback as possible :)

docs/configuration/settings.rst
```diff
@@ -480,6 +480,44 @@ In order to replace it by another one:
...
+    multiauth.policy.google.client_secret = UAlL-054uyh5in4b6u8jhg5o3hnj
+
+At this point, Kinto should be able to start and OpenID Authentication should work as described in the :ref:`API docs <authentication-openid>`.
+
+**Advanced settings**
+
+.. code-block:: ini
+
+    # User ID field name (Default: `sub`)
+    multiauth.policy.oidc.userid_field = email
```
Natim: Should we replace oidc by google here?

leplatrem:
The code is ready to be thoroughly reviewed/merged. 

Now the question is: should we do the scope filtering feature in this PR as well as revamp the whole authentication documentation? My take is «no» :)

leplatrem:
@Natim final r?

almet:
> Now the question is: should we do the scope filtering feature in this PR as well as revamp the whole authentication documentation? My take is «no» :)

That's the easy way, for sure ! But the risk is that it will stay as it is. I remember that this auth documentation have had to be revamped for some time now and it's still not the case.

My take would be to have documentation before merging this ;) (but, it's easy to say when you're not the one doing it!)

Natim:
I think the authentication documentation already have been revamp to use account rather than the dummy Basic Auth, I don't think it is related to openid anyway but more to the account plugin so another pull-request will be fine.

I think it is good enough as it is to be merge right away, we should create the issue regarding documentation and scopes management once we know how we are supposed to handle it with Auth0.

leplatrem:
> I think it is good enough as it is to be merge right away

I'll create the issue once approved and merged :)

kinto/plugins/openid/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+        try:
+            authmeth, payload = authorization.split(' ', 1)
+        except ValueError:
+            return None
+
+        if authmeth.lower() != self.header_type.lower():
+            return None
+
+        try:
+            # Bearer+OIDC id_token=jwt, access_token=bearer
```
Natim: Is this on purpose? I though we said that we would only send Bearer token and not id_tokens anymore?

Natim:
Since OpenID is about Authentication and Oauth about Authorization, I believe it is fine not validating scopes here and using the userinfo_endpoint to validate bearer tokens.

In the meantime I believe we should remove the custom authentication flow using id_token.

leplatrem:
> Since OpenID is about Authentication and Oauth about Authorization, I believe it is fine not validating scopes here and using the userinfo_endpoint to validate bearer tokens.

Our original use case is to obtain different userids depending on scope (to isolate data between apps basically).
What if we just define two clients IDs and define two policies?

```ini
    multiauth.policy.notes.issuer = https://accounts.google.com
    multiauth.policy.notes.client_id = YYYYYYYY.apps.googleusercontent.com
    multiauth.policy.notes.client_secret = UAlL-YYYYYYYY

    multiauth.policy.lockbox.issuer = https://accounts.google.com
    multiauth.policy.lockbox.client_id = XXXXXX.apps.googleusercontent.com
    multiauth.policy.lockbox.client_secret = UAlL-XXXXXX
```
The user ID will be `notes:abc` on one side, and `lockbox:abc` on the other side... The IdP could restrict the access to one or another with the callback/redirect URIs settings.

> I believe we should remove the custom authentication flow using id_token.

I don't know, using ID tokens saves a round trip to the IdP. But yes, it would make everything simpler indeed. I wish I had an example of flow where we obtain JWT access tokens, but I don't. Yeah, you're right, let's make it super stupid and simple: just access tokens.

Natim:
> What if we just define two clients IDs and define two policies?

The issue is that in that case the bearer_token will be validated in both case right?

leplatrem:
> The issue is that in that case the bearer_token will be validated in both case right?

If it's for the second policy, yes.

PR #1436, opened 2018-01-03
Use index for delete all

glasserc:
This is a potential fix for e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1421999, where delete_all is very slow because it can't use any indices. This approach changes the collation for id, parent_id, and collection_id columns to "C" because they are not meant to be a human language. This allows the existing `idx_records_parent_id_collection_id_last_modified` (which is btree on `parent_id` and then some other stuff) to service queries like `SELECT * FROM records WHERE parent_id LIKE '/buckets/abcd%'`.

Other approaches are possible. If we would rather keep the collation the same, we could add a new index using `text_pattern_ops`. This should be more efficient than that, if we believe this is more technically correct.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,13 @@ This document describes changes between each past release.
 
 - Update the Docker compose configuration to use memcache for the cache backend (#1405)
 
+**Operational concerns**
+
+- *The schema for the Postgres ``storage`` backend has changed.* This
+  changes some ID columns to use the "C" collation, which will make
+  delete_all queries faster. This may change the sort order and
+  grouping of record IDs.
+
 **New features**
 
 - New setting ``kinto.backoff_percentage`` to only set the backoff header a portion of the time
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,10 +12,15 @@ This document describes changes between each past release.
 
 **Operational concerns**
 
-- *The schema for the Postgres ``storage`` backend has changed.* This
+- *The schema for the Postgres storage backend has changed.* This
   changes some ID columns to use the "C" collation, which will make
-  delete_all queries faster. This may change the sort order and
-  grouping of record IDs.
+  ``delete_all`` queries faster. (See
+  e.g. https://www.postgresql.org/docs/9.6/static/indexes-opclass.html,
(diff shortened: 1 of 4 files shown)


kinto/core/storage/postgresql/schema.sql
```diff
@@ -23,9 +23,9 @@ IMMUTABLE;
 -- Actual records
 --
 CREATE TABLE IF NOT EXISTS records (
-    id TEXT NOT NULL,
-    parent_id TEXT NOT NULL,
-    collection_id TEXT NOT NULL,
+    id TEXT COLLATE "C" NOT NULL,
```
peterbe: The comment in `migration_018_019.sql` about *why* the use of a specific collation would be useful here.

CHANGELOG.rst
```diff
@@ -10,6 +10,13 @@ This document describes changes between each past release.
 
 - Update the Docker compose configuration to use memcache for the cache backend (#1405)
 
+**Operational concerns**
+
+- *The schema for the Postgres ``storage`` backend has changed.* This
```
leplatrem: nit: I don't think the ` `` ` are rendered within italic blocks

CHANGELOG.rst
```diff
@@ -10,6 +10,13 @@ This document describes changes between each past release.
...
+  changes some ID columns to use the "C" collation, which will make
+  delete_all queries faster. This may change the sort order and
+  grouping of record IDs.
```
leplatrem: the *default* sort order ?

CHANGELOG.rst
```diff
@@ -10,6 +10,13 @@ This document describes changes between each past release.
 
 - Update the Docker compose configuration to use memcache for the cache backend (#1405)
 
+**Operational concerns**
+
+- *The schema for the Postgres ``storage`` backend has changed.* This
+  changes some ID columns to use the "C" collation, which will make
+  delete_all queries faster. This may change the sort order and
```
leplatrem:
```
``delete_all``
```

leplatrem: (suggestion) Maybe a link to a Postgresql reference that mentions why the C collation will leverage the index could be useful

PR #1439, opened 2018-01-05
Creates a settings blacklist for the PostgreSQL storage backend.

emamurho:
Fixes #

- [x] Create a black list of settings for postgresql storage client .
- [x] Uses existing test

PR #1443, opened 2018-01-09
Preparing release 8.1.0

glasserc:


requirements.txt
```diff
@@ -38,7 +38,7 @@ ujson==1.35
 urllib3==1.22
 venusian==1.1.0
 waitress==1.1.0
-WebOb==1.7.4
+WebOb==1.8.0rc1
```
Natim: Really?

Natim:
Can we check the changelog according to https://github.com/Kinto/kinto/compare/8.0.0...master ?

requirements.txt
glasserc: Hmm, oops? I just did `make build-requirements` and committed without really reviewing it. I'll roll it back..

Natim: No worries just checking :)

leplatrem:
It should be 8.1.0 because we have new features.

And since #1439 was reviewed we could include it

PR #1452, opened 2018-01-17
Prepare 7.6.6

glasserc:
This is to try to rule out the hypothesis that it's an SQLAlchemy dependency bump that's at fault.


glasserc:
I'm pretty sure the performance characteristics we're seeing are only due to DB migrations and changed queries.

PR #1453, opened 2018-01-18
Hawk request authentication

### LINKED ISSUE #1331: Add a way to handle HAWK credentials
A number of users are trying to figure out a way to prevent replays attack when using Kinto with game programming.

A solution to that problem would be to add a Hawk authentication handler.

TDress:
Fixes #1331

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.


dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
 bravado_core==4.11.3
 flake8==3.5.0
 mock==2.0.0
+mohawk==0.3.4
```
Natim: I don't think we should put it here since it is already in the setup.py

docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
...
+some implementations of Hawk in different programming languages.
+
+Mohawk auth is enabled by turning on ``kinto.hawk.enabled`` and providing 
+credential values for ``kinto.hawk.id``, ``kinto.hawk.secret``, and
+``kinto.hawk.algo``.  
+
+.. code-block:: ini
+
+    kinto.hawk.enabled = true
+    kinto.hawk.id = 'some client id'
```
Natim: Arguably this id and secret should be unique per users.

kinto/core/errors.py
```diff
@@ -52,6 +54,7 @@ class ERRORS(Enum):
     | 410         | 202   | Service deprecated                             |
     +-------------+-------+------------------------------------------------+
     """
+    INVALID_HAWK_AUTH_TOKEN = 103
```
Natim: We should reuse code 105 for that.

kinto/core/__init__.py
```diff
@@ -51,6 +51,7 @@
         'kinto.core.initialization.setup_storage',
         'kinto.core.initialization.setup_permission',
         'kinto.core.initialization.setup_cache',
+        'kinto.core.initialization.setup_hawk_auth',
```
Natim: I believe instead of putting it in core we should create a hawk plugin in the plugins repository.

kinto/core/initialization.py
```diff
@@ -252,6 +253,23 @@ def setup_cache(config):
...
+    settings = config.get_settings()
+    
+    if 'hawk.enabled' in settings and settings['hawk.enabled']:
+        def on_new_request(event):
+            request = event.request
+            authenticator = HawkAuth(settings['hawk.id'],   
+                                     settings['hawk.secret'], 
+                                     settings['hawk.algo'],
+                                     config)
+            if not authenticator.authenticate(request):
```
Natim: We should plug it to multiauth so that we can use multiple authentication mecanisms in parallel.

kinto/core/initialization.py
```diff
@@ -322,7 +340,7 @@ def install_middlewares(app, settings):
 
 def setup_logging(config):
     """Setup structured logging, and emit `request.summary` event on each
-    request, as recommanded by Mozilla Services standard:
+    request, as recommended by Mozilla Services standard:
```
Natim: Good catch, do you mind filling a pull-request for this one change only?

Natim:
Thank you @TDress for starting this work. I started to put together some comment.

I believe the path forward is:

- Create a hawk plugin in kinto/plugins/hawk/
- Use an Authentication backend (you can see [kinto-portier](https://github.com/Kinto/kinto-portier/blob/master/kinto_portier/authentication.py) [kinto-fxa](https://github.com/Kinto/kinto-fxa/blob/master/kinto_fxa/authentication.py) or even the [account plugin](https://github.com/Kinto/kinto/tree/master/kinto/plugins/accounts)

I believe for kinto we need to authenticate users, having only one id and secret in the config doesn't allow us to authenticate users but rather give access to the kinto instance.

We should tight the Hawk authentication with the account plugin so that we can create users there and then use the user id as the hawk id and grab an API secret that would be used as a Hawk secret that we can also rotate if needed.

kinto/core/__init__.py
leplatrem: ~~repository~~ → folder :)

kinto/core/initialization.py
TDress: No problem  [PR #1455](https://github.com/Kinto/kinto/pull/1455)

docs/configuration/settings.rst
TDress: Hey @Natim ,
So you think we should use one secret to authenticate all users on a kinto instance? So userid will be used as client ID and the secret will be defined in settings.  Am I misunderstanding that?   
When you say we can "rotate the secret as needed", do we want users to just update the plaintext secret in the settings when they want to, or do we want to provide a way to generate a new secret and rotate it from the CLI?    

thanks!

Natim: > So you think we should use one secret to authenticate all users on a kinto instance?

No not at all, that's what you've implemented but that's not what we are aiming at.

> So user ID will be used as client ID and the secret will be defined in settings.

I guess at some point we should have an endpoint to trade the account identifiers with a Hawk secret.

The secret will not be defined in the settings, but generated randomly and stored inside the database as a property of the account record.

> When you say we can "rotate the secret as needed"

For now you can ignore that part, I will try to put together a blueprint to explain that in details.

TDress:
Hey @Natim, please see new commit.  Some code to keep the discussion going, but it still needs work.  #1466 
image

Natim:
Superseeded by #1466

PR #1454, opened 2018-01-18
Add collations to the timestamps table

glasserc:
The trigger bump_timestamp does a query against the timestamps table
using the parent_id and collection_id from the records table (which
are now collation C). If the timestamps columns are not of collation
C, the primary key indices could potentially include different
sequences of letters that should be considered equal, or equal
sequences of bytes which are considered different by the collation, so
Postgres does not use the index. This makes all queries against the
timestamps table perform a table scan, which performs very
badly. Because every request of a default_bucket does at least one
create(), this makes every single HTTP request very expensive.

If we update the collation of the columns in the timestamp table to
match the records table, the queries should return to their previous speed.

- (n/a) Add documentation.
- (n/a) Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,11 @@ This document describes changes between each past release.
 8.1.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+**Operational concerns**
+
+- *The schema for the Postgres storage backend has changed.* This
+  changes some more ID columns to use the "C" collation, which fixes a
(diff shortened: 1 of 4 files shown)


Natim:
You need to update the expected version in the SQL file too and the test will pass. Other than that r+

glasserc:
Yeah, I always forget that one. I knew I was going to forget something and was leaving it to the tests to show me where it was. Thanks for fixing it.

PR #1461, opened 2018-01-23
Change collation of ID columns in permission backend to speed up delete_object_permissions

### LINKED ISSUE #1460: Flushing a server breaks migration forever
Using the Kinto `flush` plugin wipes the metadata table, which removes information about the storage schema version. This means it is impossible to do a migration, because we no longer have any idea what version to migrate *from*.

There is currently error-handling logic which tries to accommodate for this. It assumes that 1. if the metadata table is empty, then the server was flushed, and 2. if the server was flushed, the schema version in-place is the "current" maximal one. Point 2 is obviously wrong in the case of e.g. local development, where you might flush the server and then, much later, try to build and run a new Kinto version. Point 1 is about to become wrong as we add migration to the permission backend, which may store data in the same metadata table. On top of all of this, the code is not clearly marked as error-handling logic, although it really is.

glasserc:
Fixes #1460.

- [x] Reuse migration mechanism to cover permission backend.
- [x] Actually add a migration for the permission backend adding COLLATE "C" to tables that need it.
- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/kinto/core/permission/postgresql/__init__.py
+++ b/kinto/core/permission/postgresql/__init__.py
@@ -79,7 +79,7 @@ class Permission(PermissionBase, MigratorMixin):
     def initialize_schema(self, dry_run=False):
         return self.create_or_migrate_schema(dry_run)
 
-    def _get_installed_version(self):
+    def get_installed_version(self):
         """Return current version of schema or None if not any found.
 
         Migrations were only added to the permission backend in
--- a/kinto/core/permission/postgresql/__init__.py
+++ b/kinto/core/permission/postgresql/__init__.py
@@ -10,7 +10,7 @@ from kinto.core.storage.postgresql.migrator import Migrator
 
 logger = logging.getLogger(__name__)
 
-HERE = os.path.abspath(os.path.dirname(__file__))
+HERE = os.path.dirname(__file__)
 
 
 class Permission(PermissionBase, Migrator):
--- a/kinto/core/permission/postgresql/__init__.py
+++ b/kinto/core/permission/postgresql/__init__.py
@@ -89,8 +89,8 @@ class Permission(PermissionBase, MigratorMixin):
         table/permission_schema_version is therefore version 1.
 
         In version 8.1.2, the permission backend added a ``metadata``
-        table. If the permission and storage backends share
-        credentials, this will be the same table created by the
+        table. If the permission and storage backends point to the
+        same database, this will be the same table created by the
         storage backend. This means either backend could create the
         table without the knowledge of the other one. For this reason,
         be careful to handle the case where the metadata table exists
--- a/kinto/core/permission/postgresql/__init__.py
+++ b/kinto/core/permission/postgresql/__init__.py
@@ -5,12 +5,15 @@ from collections import OrderedDict
 
 from kinto.core.permission import PermissionBase
 from kinto.core.storage.postgresql.client import create_from_config
+from kinto.core.storage.postgresql.migrator import Migrator
 
 
 logger = logging.getLogger(__name__)
 
+HERE = os.path.abspath(os.path.dirname(__file__))
 
-class Permission(PermissionBase):
+
+class Permission(PermissionBase, Migrator):
     """Permission backend using PostgreSQL.
 
     Enable in configuration::
@@ -63,12 +66,68 @@ class Permission(PermissionBase):
 
     :noindex:
     """  # NOQA
+
+    name = 'permission'
+    schema_version = 2
+    schema_file = os.path.join(HERE, 'schema.sql')
+    migrations_directory = os.path.join(HERE, 'migrations')
+
     def __init__(self, client, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.client = client
 
     def initialize_schema(self, dry_run=False):
-        # Check if user_principals table exists.
+        return self.create_or_migrate_schema(dry_run)
+
+    def _get_installed_version(self):
+        """Return current version of schema or None if not any found.
+
+        Migrations were only added to the permission backend in
+        8.1.2. Before this, the permission backend was only the two
+        tables ``user_principals`` and ``access_control_entries``. The
+        presence of these two tables and absence of a metadata
+        table/permission_schema_version is therefore version 1.
+
+        In version 8.1.2, the permission backend added a ``metadata``
+        table. If the permission and storage backends share
+        credentials, this will be the same table created by the
+        storage backend. This means either backend could create the
+        table without the knowledge of the other one. For this reason,
+        be careful to handle the case where the metadata table exists
+        but no version exists.
+
+        """
+        query = "SELECT tablename FROM pg_tables WHERE tablename = 'metadata';"
+        with self.client.connect() as conn:
+            result = conn.execute(query)
+            table_exists = result.rowcount > 0
+
+        if table_exists:
+            query = """
+            SELECT value AS version
+              FROM metadata
+             WHERE name = 'permission_schema_version'
+             ORDER BY LPAD(value, 3, '0') DESC
+             LIMIT 1;
+            """
+            with self.client.connect() as conn:
+                result = conn.execute(query)
+                if result.rowcount > 0:
+                    return int(result.fetchone()['version'])
+
+        # Either the metadata table doesn't exist, or it doesn't have
+        # a permission_schema_version row. Many possiblities exist:
+        #
+        # - Maybe we are migrating from <8.1.2 and the permission
+        # backend doesn't have a metadata table.
+        #
+        # - Maybe we are on a new install and don't have any tables.
+        #
+        # - Maybe we are on a new install and the storage backend has
+        # created the metadata table but we haven't initialized yet.
+        #
+        # Check if user_principals table exists. If it does, we are
+        # migrating from pre-8.1.2 and we are version 1.
         query = """
         SELECT 1
           FROM information_schema.tables
@@ -77,23 +136,12 @@ class Permission(PermissionBase):
         with self.client.connect(readonly=True) as conn:
             result = conn.execute(query)
             if result.rowcount > 0:
-                logger.info('PostgreSQL permission schema is up-to-date.')
-                return
+                return 1
 
-        # Create schema
-        here = os.path.abspath(os.path.dirname(__file__))
-        sql_file = os.path.join(here, 'schema.sql')
-
-        if dry_run:
-            logger.info("Create permission schema from '{}'".format(sql_file))
-            return
-
-        # Since called outside request, force commit.
-        with open(sql_file) as f:
-            schema = f.read()
-        with self.client.connect(force_commit=True) as conn:
-            conn.execute(schema)
-        logger.info('Created PostgreSQL permission tables')
+        # Metadata table missing or has no
+        # permission_schema_version, and no user_principals table. We
+        # need to initialize.
+        return None
 
     def flush(self):
         query = """
--- a/kinto/core/permission/postgresql/__init__.py
+++ b/kinto/core/permission/postgresql/__init__.py
@@ -5,7 +5,7 @@ from collections import OrderedDict
 
 from kinto.core.permission import PermissionBase
 from kinto.core.storage.postgresql.client import create_from_config
-from kinto.core.storage.postgresql.migrator import Migrator
+from kinto.core.storage.postgresql.migrator import MigratorMixin
 
 
 logger = logging.getLogger(__name__)
@@ -13,7 +13,7 @@ logger = logging.getLogger(__name__)
 HERE = os.path.dirname(__file__)
 
 
-class Permission(PermissionBase, Migrator):
+class Permission(PermissionBase, MigratorMixin):
     """Permission backend using PostgreSQL.
 
     Enable in configuration::
--- a/kinto/core/permission/postgresql/migrations/migration_001_002.sql
+++ b/kinto/core/permission/postgresql/migrations/migration_001_002.sql
@@ -6,6 +6,9 @@ CREATE TABLE IF NOT EXISTS metadata (
     value VARCHAR(512) NOT NULL
 );
 
+-- IDs are not really human language text, so set them to be COLLATE
+-- "C" rather than the DB default collation. This also speeds up
+-- prefix-match queries (object_id LIKE '/bucket/abc/%').
 ALTER TABLE user_principals
     ALTER COLUMN user_id TYPE TEXT COLLATE "C";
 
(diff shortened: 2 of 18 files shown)


glasserc:
/cc @peterbe. If you were interested in getting your hands dirty in the Kinto codebase, this might be a PR to look at.

I'm not exactly sure what the test failure is about, but some of these tests seem to take a strange attitude about cleaning up after themselves -- looking at it now.

Natim:
Adding @leplatrem let's try to pair review it tomorrow

kinto/core/permission/postgresql/schema.sql
```diff
@@ -4,14 +4,14 @@
 SET client_min_messages TO ERROR;
 
 CREATE TABLE IF NOT EXISTS user_principals (
-    user_id TEXT,
+    user_id TEXT COLLATE "C",
```
peterbe: Would be nice with a comment here that justifies this hack.

kinto/core/permission/postgresql/schema.sql
```diff
@@ -4,14 +4,14 @@
...
     principal TEXT,
 
     PRIMARY KEY (user_id, principal)
 );
 
 CREATE TABLE IF NOT EXISTS access_control_entries (
-    object_id TEXT,
+    object_id TEXT COLLATE "C",
```
peterbe: ditto

kinto/core/permission/postgresql/__init__.py
```diff
@@ -5,12 +5,15 @@
 
 from kinto.core.permission import PermissionBase
 from kinto.core.storage.postgresql.client import create_from_config
+from kinto.core.storage.postgresql.migrator import Migrator
 
 
 logger = logging.getLogger(__name__)
 
+HERE = os.path.abspath(os.path.dirname(__file__))
```
peterbe: Why the `os.path.abspath`? `os.path.dirname` should always give a valid directory path.

kinto/core/permission/postgresql/__init__.py
```diff
@@ -77,23 +136,12 @@ def initialize_schema(self, dry_run=False):
...
-        # Since called outside request, force commit.
-        with open(sql_file) as f:
-            schema = f.read()
-        with self.client.connect(force_commit=True) as conn:
-            conn.execute(schema)
-        logger.info('Created PostgreSQL permission tables')
+        # Metadata table missing or has no
+        # permission_schema_version, and no user_principals table. We
+        # need to initialize.
+        return None
```
peterbe: nit: Ending a function without `return None` is the same as ending the function with a `return None`.

kinto/core/permission/postgresql/migrations/migration_001_002.sql
```diff
@@ -0,0 +1,15 @@
...
+    value VARCHAR(512) NOT NULL
+);
+
+ALTER TABLE user_principals
+    ALTER COLUMN user_id TYPE TEXT COLLATE "C";
+
+ALTER TABLE access_control_entries
+    ALTER COLUMN object_id TYPE TEXT COLLATE "C";
+
+INSERT INTO metadata VALUES ('permission_schema_version', '2');
```
peterbe: nit: If the `metadata` table already existed, it's not a guarantee that the column order is exactly `'name', 'value'`. A safer INSERT statement would be:
`INSERT INTO metadata (name, value) VALUES ('permission_schema_version', '2');`

tests/core/schema/postgresql-permission-1.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE IF NOT EXISTS user_principals (
+    user_id TEXT,
```
peterbe: I'm confused. Why isn't this `COLLATE "C"` too?

kinto/core/storage/postgresql/migrator.py
```diff
@@ -0,0 +1,95 @@
...
+A helper class to run migrations using a series of SQL files.
+"""
+
+import logging
+import os
+
+logger = logging.getLogger(__name__)
+
+
+class Migrator(object):
```
peterbe: No need to `(object)` in any py 3 code.

peterbe: Also, is `MigratorMixin` a better name to indicate to the code reader how you're supposed to use this?

kinto/core/storage/postgresql/migrator.py
```diff
@@ -0,0 +1,95 @@
...
+    migrations_directory = None
+
+    def _get_installed_version(self):
+        """Return current version of schema or None if none found.
+
+        Override this.
+
+        This may be called several times during a single migration.
+        """
+        pass
```
peterbe: This should probably be `raise NotImplementedError('method not overwritten')`

kinto/core/storage/postgresql/migrator.py
```diff
@@ -0,0 +1,95 @@
...
+    """The file to find the current "newest" schema in. Override this."""
+    schema_file = None
+
+    """The directory to find migration files in.
+
+    Migrations should each be a file named migration_nnn_mmm.sql, where mmm = nnn + 1.
+    """
+    migrations_directory = None
+
+    def _get_installed_version(self):
```
peterbe: Weird that a privately named method is expected to be overridden.

kinto/core/storage/postgresql/migrator.py
```diff
@@ -0,0 +1,95 @@
...
+            self.create_schema(dry_run)
+            return
+
+        logger.info('Detected PostgreSQL {} schema version {}.'.format(self.name, version))
+        if version == self.schema_version:
+            logger.info('PostgreSQL {} schema is up-to-date.'.format(self.name))
+            return
+
+        self.migrate_schema(version, dry_run)
+        return
```
peterbe: not necessary

tests/core/test_storage_migrations.py
```diff
@@ -8,9 +8,72 @@
...
 from kinto.core.permission import postgresql as postgresql_permission
 from kinto.core.storage import postgresql as postgresql_storage
+from kinto.core.storage.postgresql.migrator import Migrator
 from kinto.core.testing import skip_if_no_postgresql
 
 
+class MigratorTest(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.migrator = Migrator()
```
peterbe: This looks dangerous. You create an instance of `Migrator` but mutate it within the tests. 
It'd be much better to ...:
```python
    def setUp(self):
        self.migrator = Migrator()
        ...
```

tests/core/test_storage_migrations.py
```diff
@@ -8,9 +8,72 @@
...
+    def test_migration_files_are_listed_if_ran_with_dry_run(self):
+        self.migrator.schema_version = 6
+
+        versions = [6, 5, 4, 3, 3]
+        self.migrator._get_installed_version = lambda: versions.pop()
+
+        with mock.patch('kinto.core.storage.postgresql.migrator.logger') as mocked:
+            self.migrator.create_or_migrate_schema(dry_run=True)
+
+        output = ''.join([repr(call) for call in mocked.info.call_args_list])
```
peterbe: Why `repr()`? Won't it already be a perfectly fine string object?

tests/core/test_storage_migrations.py
```diff
@@ -8,9 +8,72 @@
...
+        sql_called = execute_sql.call_args_list[-1][0][0]
+        self.assertIn('migrations/migration_005_006.sql', sql_called)
+
+    def test_migration_files_are_listed_if_ran_with_dry_run(self):
+        self.migrator.schema_version = 6
+
+        versions = [6, 5, 4, 3, 3]
+        self.migrator._get_installed_version = lambda: versions.pop()
+
+        with mock.patch('kinto.core.storage.postgresql.migrator.logger') as mocked:
```
peterbe: Arguably, perhaps all logging you should be captured. E.g.

```python
    @mock.patch('kinto.core.storage.postgresql.migrator.logger')
    def setUp(self, mocked_logger):
        self.migrator = Migrator()
        self.mocked_logger = mocked_logger
```

tests/core/test_storage_migrations.py
```diff
@@ -8,9 +8,72 @@
...
+                self.migrator.create_or_migrate_schema()
+                self.assertFalse(execute_sql.called)
+
+    def test_migration_file_is_executed_for_every_intermediary_version(self):
+        self.migrator.schema_version = 6
+
+        versions = [6, 5, 4, 3, 3]
+        self.migrator._get_installed_version = lambda: versions.pop()
+
+        with mock.patch.object(self.migrator, '_execute_sql_file') as execute_sql:
```
peterbe: This is now repeated in 3 of the 5 tests so it should probably be mocked in the `setUp` method.

kinto/core/permission/postgresql/__init__.py
peterbe: What I mean is; If there's a strong reason for using `os.path.abspath` it should be a code comment that explains why. If there isn't a reason, it can be deleted and only use `os.path.dirname(__file__)`.

kinto/core/permission/postgresql/__init__.py
```diff
@@ -63,12 +66,68 @@ class Permission(PermissionBase):
...
+              FROM metadata
+             WHERE name = 'permission_schema_version'
+             ORDER BY LPAD(value, 3, '0') DESC
+             LIMIT 1;
+            """
+            with self.client.connect() as conn:
+                result = conn.execute(query)
+                if result.rowcount > 0:
+                    return int(result.fetchone()['version'])
+
```
peterbe: Indent this whole block one to the right and avoid the second connection creation.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -8,13 +8,15 @@
     DEFAULT_ID_FIELD, DEFAULT_MODIFIED_FIELD, DEFAULT_DELETED_FIELD,
     MISSING)
 from kinto.core.storage.postgresql.client import create_from_config
+from kinto.core.storage.postgresql.migrator import Migrator
 from kinto.core.utils import COMPARISON
 
 
 logger = logging.getLogger(__name__)
+HERE = os.path.abspath(os.path.dirname(__file__))
```
peterbe: ditto about use of `abspath`

kinto/core/permission/postgresql/__init__.py
glasserc: Apparently as of Python 3.4, `__file__` is almost always an absolute path. I have updated the uses of it in the codebase.

kinto/core/permission/postgresql/__init__.py
glasserc: While this is technically true, I believe returning `None` explicitly in this particular case is a better expression of intent here, since the function is meant to return None when it cannot identify the schema version in play. But I agree with your suggestion in a couple of other places.

tests/core/schema/postgresql-permission-1.sql
glasserc: This is the original schema, from which I want to verify that migration is possible.

tests/core/test_storage_migrations.py
glasserc: Oops, yeah. Good catch. I hate XUnit-style test cases..

tests/core/test_storage_migrations.py
glasserc: Each call is a tuple of arguments. I guess the idea is to make sure that forms such as `logger.info("Executing migration {}", "migrations/migration_003_004.sql")` are still OK?

tests/core/test_storage_migrations.py
glasserc: Rethinking how logging is done in the test suite is outside the scope of this PR, although I agree it needs work.

glasserc:
Thanks for the suggestions -- I'll try to clean up the mocking in the test cases tomorrow.

kinto/core/permission/postgresql/__init__.py
```diff
@@ -63,12 +66,68 @@ class Permission(PermissionBase):
...
+        Migrations were only added to the permission backend in
+        8.1.2. Before this, the permission backend was only the two
+        tables ``user_principals`` and ``access_control_entries``. The
+        presence of these two tables and absence of a metadata
+        table/permission_schema_version is therefore version 1.
+
+        In version 8.1.2, the permission backend added a ``metadata``
+        table. If the permission and storage backends share
+        credentials, this will be the same table created by the
```
leplatrem: nit: *share credentials* → *share the same database* ?

kinto/core/storage/postgresql/__init__.py
```diff
@@ -157,48 +118,68 @@ def _check_database_encoding(self):
...
-        """Delete records from tables without destroying schema. Mainly used
-        in tests suites.
+        """Delete records from tables without destroying schema.
+
+        This is used in test suites as well as in the flush plugin.
         """
         query = """
         DELETE FROM records;
         DELETE FROM timestamps;
-        DELETE FROM metadata;
```
leplatrem: So this fixes #1460

tests/core/schema/postgresql-permission-1.sql
```diff
@@ -0,0 +1,20 @@
...
+    principal TEXT,
+
+    PRIMARY KEY (object_id, permission, principal)
+);
+CREATE INDEX IF NOT EXISTS idx_access_control_entries_object_id
+    ON access_control_entries(object_id);
+CREATE INDEX IF NOT EXISTS idx_access_control_entries_permission
+  ON access_control_entries(permission);
+CREATE INDEX IF NOT EXISTS idx_access_control_entries_principal
+  ON access_control_entries(principal);
```
leplatrem: You might want to add a row in metadata here.

kinto/core/permission/postgresql/migrations/migration_001_002.sql
```diff
@@ -0,0 +1,18 @@
...
+-- migration status for both. Only one schema actually has to create
+-- it.
+CREATE TABLE IF NOT EXISTS metadata (
+    name VARCHAR(128) NOT NULL,
+    value VARCHAR(512) NOT NULL
+);
+
+-- IDs are not really human language text, so set them to be COLLATE
+-- "C" rather than the DB default collation. This also speeds up
+-- prefix-match queries (object_id LIKE '/bucket/abc/%').
```
peterbe: 👍

tests/core/schema/postgresql-permission-1.sql
glasserc: That would create a schema that was different from the PostgreSQL permission v1 schema.

kinto/core/storage/postgresql/__init__.py
glasserc: Yes (as described in the PR description as well as the commit comments).

PR #1466, opened 2018-01-29
HAWK request authentication plugin.  Relies on the account plugin for

TDress:
storing secret, and the account user ID is used as the client ID.
Account creation triggers a secret to be generated and exposed to the
user.

Fixes #

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ $(PYTHON):
 build-requirements:
 	$(VIRTUALENV) $(TEMPDIR)
 	$(TEMPDIR)/bin/pip install -U pip
-	$(TEMPDIR)/bin/pip install -Ue ".[monitoring,postgresql,memcached]"
+	$(TEMPDIR)/bin/pip install -Ue ".[monitoring,postgresql,memcached,hawk]"
 	$(TEMPDIR)/bin/pip freeze | grep -v -- '-e' > requirements.txt
 
 build-kinto-admin: need-npm
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -519,6 +519,28 @@ At this point, Kinto should be able to start and OpenID Authentication should wo
     multiauth.policy.google.state_ttl_seconds = 3600
 
 
+Mohawk request Auth
+:::::::::::::::::::
+
+`Mohawk <https://mohawk.readthedocs.io/en/latest/index.html>`_ implements the 
+`Hawk authorization scheme <https://github.com/hueniverse/hawk>`_, which 
+allows you to authenticate messages from your application backend to Kinto.
+In order to prevent replay attacks, nonce checking is enforced.  
+See `here <https://github.com/hueniverse/hawk#implementations>` for 
+some implementations of Hawk in different programming languages.
+
+Mohawk auth is enabled by turning on ``kinto.hawk.enabled`` and providing 
+credential values for ``kinto.hawk.id``, ``kinto.hawk.secret``, and
+``kinto.hawk.algo``.  
+
+.. code-block:: ini
+
+    kinto.hawk.enabled = true
+    kinto.hawk.id = 'some client id'
+    kinto.hawk.secret = 'some secret key'
+    kinto.hawk.algo = 'sha256'
+
+
 Custom Authentication
 :::::::::::::::::::::
 
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -519,26 +519,25 @@ At this point, Kinto should be able to start and OpenID Authentication should wo
     multiauth.policy.google.state_ttl_seconds = 3600
 
 
-Mohawk request Auth
-:::::::::::::::::::
+Hawk request Auth
+:::::::::::::::::
 
 `Mohawk <https://mohawk.readthedocs.io/en/latest/index.html>`_ implements the 
-`Hawk authorization scheme <https://github.com/hueniverse/hawk>`_, which 
-allows you to authenticate messages from your application backend to Kinto.
-In order to prevent replay attacks, nonce checking is enforced.  
-See `here <https://github.com/hueniverse/hawk#implementations>` for 
-some implementations of Hawk in different programming languages.
+`Hawk authorization scheme <https://github.com/hueniverse/hawk>`_, which
+allows you to authenticate requests. In order to prevent replay
+attacks, nonce checking is enforced.
+See `here <https://github.com/hueniverse/hawk#implementations>` for some
+implementations of Hawk in different programming languages.
 
-Mohawk auth is enabled by turning on ``kinto.hawk.enabled`` and providing 
-credential values for ``kinto.hawk.id``, ``kinto.hawk.secret``, and
-``kinto.hawk.algo``.  
+Hawk authentication can be enabled by turning on
+``kinto.hawk.enabled``. By default Hawk will use the ``sha256`` algorithm.
+
+But you can change this behaviour by setting ``kinto.hawk.algorithm`` value.
 
 .. code-block:: ini
 
     kinto.hawk.enabled = true
-    kinto.hawk.id = 'some client id'
-    kinto.hawk.secret = 'some secret key'
-    kinto.hawk.algo = 'sha256'
+    kinto.hawk.algorithm = sha256
 
 
 Custom Authentication
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -8,7 +8,7 @@ import random
 from pyramid.events import NewRequest, NewResponse
 from pyramid.exceptions import ConfigurationError
 from pyramid.httpexceptions import (HTTPTemporaryRedirect, HTTPGone,
-                                    HTTPBadRequest)
+                                    HTTPBadRequest, HTTPUnauthorized)
 from pyramid.renderers import JSON as JSONRenderer
 from pyramid.response import Response
 from pyramid.security import NO_PERMISSION_REQUIRED
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -7,8 +7,7 @@ import random
 
 from pyramid.events import NewRequest, NewResponse
 from pyramid.exceptions import ConfigurationError
-from pyramid.httpexceptions import (HTTPTemporaryRedirect, HTTPGone,
-                                    HTTPBadRequest, HTTPUnauthorized)
+from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPGone, HTTPBadRequest
 from pyramid.renderers import JSON as JSONRenderer
 from pyramid.response import Response
 from pyramid.security import NO_PERMISSION_REQUIRED
--- a/kinto/plugins/accounts/views.py
+++ b/kinto/plugins/accounts/views.py
@@ -5,6 +5,8 @@ from pyramid.security import Authenticated, Everyone
 from pyramid.settings import aslist
 from pyramid.events import subscriber
 
+from ..hawk.hawkauth import HawkAuth
+
 from kinto.views import NameGenerator
 from kinto.core import resource, utils
 from kinto.core.errors import raise_invalid, http_error
@@ -106,8 +108,9 @@ class Account(resource.ShareableResource):
 
     def process_record(self, new, old=None):
         new = super(Account, self).process_record(new, old)
-
         new['password'] = hash_password(new['password'])
+        # Generate HAWK client secret for use with the Kinto hawk plugin
+        new['hawk_secret'] = HawkAuth.generate_secret()
 
         # Administrators can reach other accounts and anonymous have no
         # selected_userid. So do not try to enforce.
--- a/kinto/plugins/hawk/__init__.py
+++ b/kinto/plugins/hawk/__init__.py
@@ -1,0 +1,6 @@
+def includeme(config):
+    config.add_api_capability(
+        'hawk',
+        description='Hawk request authentication',
+        url='https://kinto.readthedocs.io/en/latest/\
+        configuration/settings.html#plugins')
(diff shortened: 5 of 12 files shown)


docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
...
+some implementations of Hawk in different programming languages.
+
+Mohawk auth is enabled by turning on ``kinto.hawk.enabled`` and providing 
+credential values for ``kinto.hawk.id``, ``kinto.hawk.secret``, and
+``kinto.hawk.algo``.  
+
+.. code-block:: ini
+
+    kinto.hawk.enabled = true
+    kinto.hawk.id = 'some client id'
```
Natim: Not accurate anymore.

Makefile
```diff
@@ -65,7 +65,7 @@ $(PYTHON):
 build-requirements:
 	$(VIRTUALENV) $(TEMPDIR)
 	$(TEMPDIR)/bin/pip install -U pip
-	$(TEMPDIR)/bin/pip install -Ue ".[monitoring,postgresql,memcached]"
+	$(TEMPDIR)/bin/pip install -Ue ".[monitoring,postgresql,memcached,mohawk]"
```
leplatrem: nitpick: I think it could be just `hawk` since the underlying library may change overtime :)

docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
     multiauth.policy.basicauth.use = myproject.authn.BasicAuthPolicy
 
 
+Mohawk request Auth
```
leplatrem: What about `Hawk Authentication` as the title?

docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
...
+:::::::::::::::::::
+
+`Mohawk <https://mohawk.readthedocs.io/en/latest/index.html>`_ implements the 
+`Hawk authorization scheme <https://github.com/hueniverse/hawk>`_, which 
+allows you to authenticate messages from your application backend to Kinto.
```
leplatrem: authenticate *requests*.

The rest of the sentence *from your application backend to Kinto.* is not really helpful

docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
...
+In order to prevent replay attacks, nonce checking is enforced.  
+See `here <https://github.com/hueniverse/hawk#implementations>` for 
+some implementations of Hawk in different programming languages.
```
leplatrem: Note: Why would someone be interested in checking these implementations from the settings docs?

docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
...
+
+Mohawk auth is enabled by turning on ``kinto.hawk.enabled`` and providing
```
leplatrem: Hawk authentication can be enabled by ...

docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
...
+`Hawk authorization scheme <https://github.com/hueniverse/hawk>`_, which 
+allows you to authenticate messages from your application backend to Kinto.
+In order to prevent replay attacks, nonce checking is enforced.  
+See `here <https://github.com/hueniverse/hawk#implementations>` for 
+some implementations of Hawk in different programming languages.
+
+Mohawk auth is enabled by turning on ``kinto.hawk.enabled`` and providing 
+credential values for ``kinto.hawk.id``, ``kinto.hawk.secret``, and
+``kinto.hawk.algo``.  
+
```
leplatrem: And also Kinto must be installed with `[hawk]` optional dep

kinto/core/initialization.py
```diff
@@ -8,7 +8,7 @@
 from pyramid.events import NewRequest, NewResponse
 from pyramid.exceptions import ConfigurationError
 from pyramid.httpexceptions import (HTTPTemporaryRedirect, HTTPGone,
-                                    HTTPBadRequest)
+                                    HTTPBadRequest, HTTPUnauthorized)
```
leplatrem: ?

kinto/plugins/accounts/views.py
```diff
@@ -4,6 +4,8 @@
 from pyramid.security import Authenticated, Everyone
 from pyramid.settings import aslist
 
+from ..hawk.hawkauth import HawkAuth
```
leplatrem: use absolute imports `from kinto.plugins.hawk`

kinto/plugins/hawk/__init__.py
```diff
@@ -0,0 +1,8 @@
+
+
```
leplatrem: you can import `HawkAuth` here, so that we can have `from kinto.plugins.hawk import HawkAuth`

kinto/plugins/hawk/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+def includeme(config):
+    config.add_api_capability(
+        'hawk',
+        description='Hawk request authentication using mohawk',
```
leplatrem: `using mohawk` is an implementation detail that we don't need to expose here

kinto/plugins/hawk/authentication.py
```diff
@@ -0,0 +1,50 @@
...
+    def forget(self, request):
+        return [('WWW-Authenticate', 'Hawk')]
+        
+    def _test_credentials(self, request):
+        """Test credentials in request against account HAWK credentials"""
+        account_creds = self._get_account_hawk_creds(request)
+        try:
+            return account_creds['id']
+        except:
+            return None
```
leplatrem: why not merge `_test_credentials` and `_get_account_hawk_creds` methods?

kinto/plugins/hawk/authentication.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def _get_account_hawk_creds(self, request):
+        """Check storage for the request account HAWK credentials.
+        
+        The accounts plugin is a depency of this plugin, so we can rely
+        on Basic Auth credentials to get the account and its HAWK creds.
+        """
+        print('check hawk')
```
leplatrem: leftover

docs/configuration/settings.rst
```diff
@@ -480,6 +480,28 @@ In order to replace it by another one:
...
+.. code-block:: ini
+
+    kinto.hawk.enabled = true
+    kinto.hawk.id = 'some client id'
+    kinto.hawk.secret = 'some secret key'
+    kinto.hawk.algo = 'sha256'
+
```
leplatrem: Also the `account` plugin must be enabled as far as I understand

kinto/plugins/hawk/hawkauth.py
```diff
@@ -0,0 +1,106 @@
...
+            key = '{id}:{nonce}:{ts}'.format(id=sender_id, nonce=nonce,
+                                             ts=timestamp)
+            if self._cache_backend.get(key):
+                return True
+            else:
+                # Messages expire after 60 seconds, so we can be confident that
+                # we only need to store the nonce for slightly longer than the
+                # timeout (because a replayed nonce won't work if the entire message
+                # is expired).
+                self._cache_backend.set(key, True, 90)
```
leplatrem: 90 should be a setting (comment says 60 BTW :))

docs/configuration/settings.rst
TDress: Maybe it's not necessary, but I included those references because the user will need to implement the hawk client.

docs/configuration/settings.rst
TDress: thanks for reading through my docs change!  Most of the change is outdated and I shouldn't have included it in the PR.  Since we are moving in the direction of basically using hawk auth as a wrapper around the account plugin, I will need to redo this.  Your comments are helpful in understanding what needs to be in the docs

docs/configuration/settings.rst
TDress: Much better since the hawk implementation (library) might change in the future

Makefile
TDress: good call!

kinto/core/initialization.py
TDress: I accidentally let this slip in from a previous commit.  I will get rid of it!

kinto/plugins/accounts/views.py
TDress: thanks for the tip!  Is this better style or is there another reason to avoid relative imports here?

kinto/plugins/hawk/authentication.py
TDress: I must have had the idea that it was more modular, but looking at it again I think you are right-- it makes more sense as one method

TDress:
Thanks @leplatrem and @Natim  for the granular feedback on these PRs!  I'm enjoying the learning experience and would love to bring this all the way to fruition.  I'm going to get another PR ready with more progress.
Just to be clear that the design is on the right track: we are essentially wrapping the account plugin with a new hawk auth policy.  Kinto uses account IDs as hawk client IDs and generates secrets stored on account records.  The secrets are exposed to the account user when doing a GET account (and when creating account).  @Natim seemed to have some ideas about how to "rotate" secrets.

Natim:
I don't think the hawk secret should be returned each time we do a GET to the account.

It shouldn't be possible to retrieve a hawk secret that was already retrieved. And it should be possible to have multiple hawk secret at the same time.

If we are going this route maybe the ID should be random as well.

I am sorry i am not really clear because I am still processing the feature and its security capabilities.
I will write a blueprint about that and keep you posted but the flow I have in mind is the following:

- [ ] Create a user account
- [ ] Grab an Hawk session for that account (Is it possible to grab a session without entering the user password?)
- [ ] Hmac the session ID in order to grab an ID and a Secret linked to that user

# Open questions

- **How do we expire sessions?** I guess we can expire session to a given time and extend that time as soon as the session is used. After a while it will disappear.
- **How do we rotate sessions?** Each time we ask for a new one
- **How do we handle multiple concurrent sessions?** We keep a list of them all linked to a user.

Natim:
Here is the Blueprint: https://github.com/Kinto/kinto/wiki/Hawk-authentication-session

kinto/plugins/accounts/views.py
leplatrem: I think it's better for readability and style, but in this case, it's mainly for consistency with the rest of the codebase. I think we only have relative imports in tests :)

TDress:
Thanks for the blueprint-- looks good.  One question I have is about removing the current session (endpoint DELETE /accounts/(userid)/hawk-sessions/current).  Does the current session = most recently used?  If there are multiple sessions, and another user on the account uses a different session, then that DELETE call would remove the wrong one?

Natim:
> Does the current session = most recently used?

No the current session means the one we are currently logged in with. It is useful for a disconnect for instance.

TDress:
Ok, ammended my commit with a new route to create a session, and a test for the token in the response header.   more commits to come

TDress:
Worked my way through saving all creds on the account.  New dependency requests-hawk for doing the HKDF  work.

TDress:
@Natim, sorry this is kind of a delayed follow-up question, but when you say
> the current session means the one we are currently logged in with
this means that the DELETE session endpoints need to be hawk auth protected so that we can match the hawk creds to the correct session.  All new endpoints require hawk auth except for creating a new session, which will still require account basic auth.  

Does this sound correct?  In the blueprint the DELETE example doesn't have an 'authorization' request header for HAWK. Do you want me to add that to the blueprint?

Natim:
> Do you want me to add that to the blueprint?

Yes sure :+1: 

> how would we find the session /credentials to use to authenticate the request?

The session won't be tied to the account record but rather stored in the cache backend.
We will use a hmac of the hawk session ID to store the hawk secret and the user ID.

When we receive the Hawk Authorization header we will have the ID in the header.
We can then hmac it to find the key in the cache backend.

> Do you think we should save the sessions as their own independent records and then also save session tokens on account records when we first create sessions? 

Yes we should save session as their own indepedent records (I don't think it is necessary to store them in the storage backend) I guess if we store the hawk secret, hawk algorithm and user ID in the session record we will have the mandatory information to link it to the user.

TDress:
Thanks for the clarification on how you think sessions and creds should be stored.  The only remaining issue I see is for removing all of the sessions that are tied to an account 
`DELETE /accounts/(user_id)/hawk-sessions`
If we have the _account_ user id stored within the session record, then when it comes time to remove every session linked to an account, we will have to look at every session record to see if the account user id matches the account of the current request.  If there are N session records, we have to look at N records in every case.  Am I making sense?  The hawk client id is the key for the session record, so we can't look up all sessions linked to an account.  The performance will be weak if we have to scan all sessions

TDress:
Hey @Natim and @leplatrem ,
The commit I just added is a start for taking the approach of saving the creds to the session, like Natim suggested, and also saving client IDs on the account for fast lookup when it comes time to delete all sessions tied to an account (like I talked about in my previous comment).  Please let me know if I'm on the right track.

leplatrem:
@Natim since you're the one at the initiative of the Hawk support, I call for your help here, review, feedback and approval needed :)

Also, since the feature brings a lot of new code for a limited audience, what would you think of moving it to a dedicated package?

Natim:
I started to take over the PR to address the changes but I agree that we should create an hawk plugin even if we share its database with the account plugin.

Natim:
I moving the idea of this PR in https://github.com/Kinto/kinto-hawk

@TDress be sure that you work will be useful and that I will make you a co-author of every commit that has your code in it.

Again, thanks a lot for your help.

Issue #1470, opened 2018-01-31
pytest-sugar breaks pytest 3.4.0


Natim:
Refs https://github.com/Frozenball/pytest-sugar/pull/133

Natim:
Refs https://github.com/Kinto/kinto/commit/71ad7adc86465d062fd073c69f98d9de5c366b77

leplatrem:
What's the status of this?

Natim:
It was fixed and release in pytest-sugar 0.9.1

glasserc:
This no longer blocks #1485.

PR #1473, opened 2018-01-31
Allow inherited resources to set a custom model instance before instantiating (fixes #1472)

### LINKED ISSUE #1472: Do not enforce instantiation of resource model during init

leplatrem:
Fixes #1472


leplatrem:
@Natim should we merge? failing tests are due to #1470 on Py35

PR #1475, opened 2018-01-31
WIP - Collection timestamp with readonly (fixes #1474)

### LINKED ISSUE #1474: collection_timestamp() always fails when Kinto is readonly
Regression from [7.5.1 to 8.1.3](https://github.com/Kinto/kinto/compare/7.5.1...8.1.3)

In the past, we were only attempting to insert into the timestamp table when everything was empty. We could have an error when accessing the collection via a readonly instance on an empty table. But accessing via a writable instance was enough, since it would create the row and never try again. 

It seems that in the last version, it always try to create (UPSERT?) even if it's not empty. It thus always fails to execute collection_timestamp() on a readonly instance.

leplatrem:
Fixes #1474 

* [x] Find elegant solution
* [x] Add tests with readonly ?
* [ ] Update docs

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ This document describes changes between each past release.
 **Bug fixes**
 
 - Allow inherited resources to set a custom model instance before instantiating (fixes #1472)
+- Fix collection timestamp retrieval when the stack is configured as readonly (fixes #1474)
 
 8.1.3 (2018-01-26)
 ------------------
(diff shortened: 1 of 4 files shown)


glasserc:
I agree that this is basically the only thing we can do... did this ever used to work back when we wrote everything to timestamps? If so, how?

leplatrem:
> did this ever used to work back when we wrote everything to timestamps? If so, how?

Yeah it wasn't trying to create if a row was present, so we had the error only when the collection was empty

PR #1476, opened 2018-01-31
Preparing release 8.1.4

leplatrem:


glasserc:
There's a version number in `docs/configuration/production.rst`. However it's pretty late in the day so maybe I'll just pick up from here.

PR #1490, opened 2018-02-08
Check for existing record before creating one

### LINKED ISSUE #1487: Is it possible to reduce write load on create() without reintroducing TOCTOU bugs?
In #1406, we changed the create() method from doing "read, then insert" to doing "insert with on conflict". This introduced a lot of write load. Unlike read load, which can often be served using CPU and memory, write load always goes to disk and is therefore more expensive. This manifested by a dramatic increase on write load in the kintowe stack, because every request is creating a default bucket and a collection in that default bucket.

Is it possible to reduce write load without the risk of TOCTOU bugs? Now that we are fairly confident in the correctness of our `INSERT` + `ON CONFLICT` strategy, we could try to re-add the "read, then insert" strategy. The read would be optimistic -- if it's there, then we don't have to insert, but if it isn't there, we can insert and maybe get a conflict, no big deal. The risk here is that we could find that the row was there, and then fail to create it even though it's been deleted just afterwards by a different transaction. However, this may be acceptable, because child elements don't actually rely on the presence or absence of their ancestor records, and even if they do, such behavior is no different from trying the INSERT + ON CONFLICT and then having another thread delete the record just afterwards.

@leplatrem What do you think?

glasserc:
Fixes #1487.

- (n/a) Add documentation.
- (n/a) Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ This document describes changes between each past release.
 
 - Restore "look before you leap" behavior in the Postgres storage
   backend create() method to check whether a record exists before
-  running the INSERT query. This check is "optimistic" in the sense
+  running the INSERT query (#1487). This check is "optimistic" in the sense
   that we can still fail to INSERT after the check succeeded, but it
   can reduce write load in configurations where there are a lot of
   create()s (i.e. when using the default_bucket plugin).
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,14 @@ This document describes changes between each past release.
 8.1.5 (unreleased)
 ------------------
 
-- Nothing changed yet.
+**Bug fixes**
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -6,7 +6,14 @@ This document describes changes between each past release.
...
 
-- Nothing changed yet.
+**Bug fixes**
+
+- Restore "look before you leap" behavior in the Postgres storage
+  backend create() method to check whether a record exists before
+  running the INSERT query. This check is "optimistic" in the sense
+  that we can still fail to INSERT after the check succeeded, but it
+  can reduce write load in configurations where there are a lot of
+  create()s (i.e. when using the default_bucket plugin).
```
leplatrem: µnit: mention #1487

Issue #1347, opened 2017-10-03
Add an webhook plugin allowing to trigger another webservice


Natim:
This issue is to actually implement that Blueprint: https://github.com/Kinto/kinto/wiki/WebHooks

We would like to be able to trigger a Lambda Function on an Kinto update event.

The goal would be to be able to configure a URL that will be notified each time an event is sent.

A bit like you can integrate Github with other services. Here our main goal would be to trigger the kinto-blockpage-generator lambda each time we publish in the kinto/blocklists bucket.

emamurho:
@Natim @leplatrem can i look into this?

Natim:
@emamurho sure.

aimanparvaiz:
@emamurho @leplatrem Hi guys, I am new here. Very interesting project I must say. I was going through this issue and had a question. Pardon my ignorance if this is a stupid question. I understand that we use Redis list in our worker code to listen for events but how do we make kinto/blocklists bucket write to (publish to) that redis list on every create event. Assumption is that we want o trigger lambda only when we create a resource in kinto/blocklists bucket. I would be grateful if you could point me to the right direction.
Thanks

leplatrem:
Hi! Welcome here! And thanks for the kind words ;)

You are right, currently we can only filter events by action / resource (http://docs.kinto-storage.org/en/stable/configuration/settings.html?highlight=listeners#filtering). 

In that case we would need a `object_ids` filter or something like that, to be able to do:
```
kinto.event_listeners.redis.actions = create
kinto.event_listeners.redis.resources = bucket collection
kinto.event_listeners.redis.buckets_ids = blocklists blog
kinto.event_listeners.redis.collection_ids = articles
```
(a little bit like we do in [kinto-emailer settings](https://github.com/Kinto/kinto-emailer/#selection))

Or maybe be able to specify URIs in `resources`: 
```
kinto.event_listeners.redis.actions = create
kinto.event_listeners.redis.resources = /buckets/blocklists /buckets/blog/collections/articles
```
(a little bit like we do in [kinto-changes settings](https://github.com/Kinto/kinto-changes/#setup)


I think that could be a very good starting point: design an intuitive/flexible set of settings and implement this new filtering feature! 

What do you think?

Natim:
Also note that if you can guarantee that the HTTP call to trigger the lambda is really quick you could call the HTTP endpoint directly without relying on redis in between.

leplatrem:
[...]

PR #1499, opened 2018-02-16
Add event listener setting to filter by object id (ref #1347)

aimanparvaiz:
Ref #1347

- [x] Add documentation.
- [x] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -414,6 +414,10 @@ class EventResourceFilter:
         return not resource or not self.resources or resource in self.resources
 
 
+class EventResourceIdFilter:
+    pass
+
+
 def setup_listeners(config):
     # Register basic subscriber predicates, to filter events.
     config.add_subscriber_predicate('for_actions', EventActionFilter)
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -464,7 +464,6 @@ class EventResourceIdFilter:
                         return True
         return False
 
-
 def setup_listeners(config):
     # Register basic subscriber predicates, to filter events.
     config.add_subscriber_predicate('for_actions', EventActionFilter)
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -516,7 +516,6 @@ def setup_listeners(config):
         resource_names = aslist(resource_value)
 
         # Optional filter by resource object id
-<<<<<<< 67e57653ea4266428c9737a4b272dbd479edfa6e
         resource_ids_setting = prefix + 'resource_ids'
         resource_id_value = utils.read_env(
             '{}.{}'.format(
@@ -526,18 +525,6 @@ def setup_listeners(config):
         # Pyramid event predicates.
         options = dict(for_actions=actions, for_resources=resource_names,
                        for_resource_ids=resource_id_uris)
-=======
-        resource_id_setting = prefix + 'resource_ids'
-        resource_id_value = utils.read_env(
-            '{}.{}'.format(
-                project_name, resource_id_setting), settings.get(resource_id_setting, ''))
-        resource_id_names = aslist(resource_id_value)
-
-        # Pyramid event predicates.
-        options = dict(for_actions=actions, for_resources=resource_names,
-                       for_resource_ids=resource_id_names)
->>>>>>> Adding resource id based filtering. Extending core/utils to handle resource type record and bucket.
-
         if ACTIONS.READ in actions:
             config.add_subscriber(listener, ResourceRead, **options)
             if len(actions) == 1:
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -432,16 +432,17 @@ class EventResourceIdFilter:
         # return True
         for resource_ids in self.resource_ids:
             returned_resource_ids = utils.parse_resource(resource_ids)
+            bucket = matchdict.get('bucket_id')
+            collection = matchdict.get('collection_id')
+            returned_bucket = returned_resource_ids.get('bucket')
+            returned_collection = returned_resource_ids.get('collection')
             if event.payload.get('action') == 'create':
                 if r_name == 'collection':
-                    bucket = matchdict.get('bucket_id')
-                    if bucket == returned_resource_ids.get('bucket'):
+                    if bucket == returned_bucket:
                         return True
                 elif r_name == 'record':
-                    bucket = matchdict.get('bucket_id')
-                    collection = matchdict.get('collection_id')
-                    if bucket == returned_resource_ids.get('bucket') and \
-                            collection == returned_resource_ids.get('collection'):
+                    if bucket == returned_bucket and \
+                            collection == returned_collection:
                         return True
             else:
                 if r_name == 'bucket':
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -516,6 +516,7 @@ def setup_listeners(config):
         resource_names = aslist(resource_value)
 
         # Optional filter by resource object id
+<<<<<<< 67e57653ea4266428c9737a4b272dbd479edfa6e
         resource_ids_setting = prefix + 'resource_ids'
         resource_id_value = utils.read_env(
             '{}.{}'.format(
@@ -525,6 +526,17 @@ def setup_listeners(config):
         # Pyramid event predicates.
         options = dict(for_actions=actions, for_resources=resource_names,
                        for_resource_ids=resource_id_uris)
+=======
+        resource_id_setting = prefix + 'resource_ids'
+        resource_id_value = utils.read_env(
+            '{}.{}'.format(
+                project_name, resource_id_setting), settings.get(resource_id_setting, ''))
+        resource_id_names = aslist(resource_id_value)
+
+        # Pyramid event predicates.
+        options = dict(for_actions=actions, for_resources=resource_names,
+                       for_resource_ids=resource_id_names)
+>>>>>>> Adding resource id based filtering. Extending core/utils to handle resource type record and bucket.
 
         if ACTIONS.READ in actions:
             config.add_subscriber(listener, ResourceRead, **options)
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -415,13 +415,62 @@ class EventResourceFilter:
 
 
 class EventResourceIdFilter:
-    pass
+    def __init__(self, resource_ids, config):
+        self.resource_ids = resource_ids
+
+    def phash(self):
+        return 'for_resource_id = {}'.format(','.join(self.resource_ids))
+
+    def __call__(self, event):
+        request = event.request
+        uri = event.payload.get('uri')
+        r_name, matchdict = utils.view_lookup(request, uri)
+        """
+        Special case for Create bucket.
+
+        """
+        if event.payload.get('action') == 'create' and r_name == 'bucket':
+            return True
+        for resource_id in self.resource_ids:
+            returned_resource_ids = utils.parse_resource(resource_id)
+            if event.payload.get('action') == 'create':
+                if r_name == 'collection':
+                    bucket = matchdict.get('bucket_id')
+                    if bucket == returned_resource_ids.get('bucket'):
+                        return True
+                elif r_name == 'record':
+                    bucket = matchdict.get('bucket_id')
+                    collection = matchdict.get('collection_id')
+                    if bucket == returned_resource_ids.get('bucket') and \
+                            collection == returned_resource_ids.get('collection'):
+                        return True
+            elif event.payload.get('action') != 'create':
+                if r_name == 'bucket':
+                    bucket = matchdict.get('id')
+                    if bucket == returned_resource_ids.get('bucket'):
+                        return True
+                elif r_name == 'collection':
+                    bucket = matchdict.get('bucket_id')
+                    collection = matchdict.get('id')
+                    if bucket == returned_resource_ids.get('bucket') and \
+                            collection == returned_resource_ids.get('collection'):
+                        return True
+                elif r_name == 'record':
+                    bucket = matchdict.get('bucket_id')
+                    collection = matchdict.get('collection_id')
+                    record = matchdict.get('id')
+                    if bucket == returned_resource_ids.get('bucket') and \
+                            collection == returned_resource_ids.get('collection') and \
+                            record == returned_resource_ids.get('record'):
+                        return True
+        return False
 
 
 def setup_listeners(config):
     # Register basic subscriber predicates, to filter events.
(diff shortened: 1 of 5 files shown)


kinto/core/initialization.py
```diff
@@ -413,13 +413,62 @@ def __call__(self, event):
...
+        for resource_id in self.resource_ids:
+            returned_resource_ids = utils.parse_resource(resource_id)
+            if event.payload.get('action') == 'create':
+                if r_name == 'collection':
+                    bucket = matchdict.get('bucket_id')
+                    if bucket == returned_resource_ids.get('bucket'):
+                        return True
+                elif r_name == 'record':
+                    bucket = matchdict.get('bucket_id')
+                    collection = matchdict.get('collection_id')
```
leplatrem: you can put those `matchdict.get()` and `returned_resource_ids.get()` above the `if` to ease reading

kinto/core/initialization.py
```diff
@@ -462,12 +511,20 @@ def setup_listeners(config):
...
+        resource_value = utils.read_env(
+            '{}.{}'.format(project_name, resource_setting), settings.get(resource_setting, ''))
         resource_names = aslist(resource_value)
 
+        # Optional filter by resource object id
+        resource_id_setting = prefix + 'resource_ids'
+        resource_id_value = utils.read_env(
+            '{}.{}'.format(
+                project_name, resource_id_setting), settings.get(resource_id_setting, ''))
+        resource_id_names = aslist(resource_id_value)
```
leplatrem: Those are not really `names` but `URIs`

kinto/core/initialization.py
```diff
@@ -413,13 +413,62 @@ def __call__(self, event):
...
+        return 'for_resource_id = {}'.format(','.join(self.resource_ids))
+
+    def __call__(self, event):
+        request = event.request
+        uri = event.payload.get('uri')
+        r_name, matchdict = utils.view_lookup(request, uri)
+        """
+        Special case for Create bucket.
+
+        """
```
leplatrem: nitpick: this could be a single line comment

aimanparvaiz:
Hey @leplatrem 
I have updated the PR taking in to account various relevant points which you had brought up earlier. I wanted to run what I have done so far by you before I made any more progress. I have added a test for checking callback when resource id is not filtered, mocking out `kinto.core.utils.parse_resource` and `kinto.core.utils.view_lookup` functions since I understood this test is about whether call back function is being called or not and not about the functionality of the mocked functions. Please correct me if I am wrong here. If I get a green light on the test I will create similar tests for resource id - collection and tests around resource ids being filtered. For now I am not entertaining the special case of creating a bucket as you suggested in you previous comment.

I have fixed a couple of breaking tests too which were breaking because I added a subscriber predicate in `setup_listeners` function in initialization.py and it was calling `kinto.core.utils.parse_resource` and `kinto.core.utils.view_lookup` functions. I have mocked out those functions. In general I have used `mock.patch` decorator please let me know if this is not the best way to go about it.

Another change is in the test `test_malformed_url_raises_an_exception` in `/tests/core/test_utils.py` because one of the URI pattern which was illegal before is now a legal case.

I have incorporated all your comments from the PR except for one which I couldn’t understand where you mentioned the following:
```
you can put those matchdict.get() and returned_resource_ids.get() above the if to ease reading
```
I would be grateful if you could please throw more light on this and I will take care of this too.

If all of this is ok, I will continue working on fixing rest of the breaking tests and adding more tests where possible.
Thanks for your patience.

tests/core/test_listeners.py
```diff
@@ -64,10 +64,18 @@ def test_listener_module_can_be_specified_via_listeners_list(self):
         })
         self.assertTrue(self.demo_mocked.called)
 
-    def test_callback_called_when_action_is_not_filtered(self):
+    @mock.patch('kinto.core.utils.parse_resource',
+                return_value={'bucket': 'b_id',
+                              'collection': 'c_id'})
+    @mock.patch('kinto.core.utils.view_lookup',
+                return_value=('record',
+                              {'bucket_id': 'b_id', 'collection_id': 'c_id'}))
```
leplatrem: Are you sure you need to mock these?

tests/core/test_listeners.py
```diff
@@ -85,10 +93,18 @@ def test_callback_is_not_called_when_action_is_filtered(self):
...
+    def test_callback_called_when_resource_is_not_filtered(self, view_lookup, parse_resource):
         config = self.make_app({
             'event_listeners': 'demo',
             'event_listeners.demo.use': 'tests.core.listeners',
+            'event_listeners.demo.resources': 'mushroom',
+            'event_listeners.demo.resource_ids': '/buckets/b_id/collections/c_id',
```
leplatrem: I wouldn't touch these tests and just add new ones

tests/core/test_utils.py
```diff
@@ -335,7 +335,7 @@ def _assert_error(self, input_arr):
             self.assertEquals(str(excinfo.value), self.error_msg)
 
     def test_malformed_url_raises_an_exception(self):
-        input_arr = ['foo', '/', '/bar', 'baz/', '/fo/o', '/buckets/sbid/scid']
+        input_arr = ['foo', '/', '/bar', 'baz/', '/buckets/sbid/scid']
```
leplatrem: Why this change?

kinto/core/utils.py
```diff
@@ -479,19 +479,54 @@ def parse_resource(resource):
...
+        resource_record = True
     else:
         raise ValueError(error_msg)
-    if bucket == '' or collection == '':
-        raise ValueError(error_msg)
-    if not id_generator.match(bucket) or not id_generator.match(collection):
-        raise ValueError(error_msg)
+    if resource_bucket:
+        if bucket == '':
+            raise ValueError(error_msg)
```
leplatrem: `id_generator` does not match empty strings so this is useless

kinto/core/utils.py
```diff
@@ -479,19 +479,54 @@ def parse_resource(resource):
...
+        if not id_generator.match(bucket):
+            raise ValueError(error_msg)
+        return {
+            'bucket': bucket
+        }
+    elif resource_record:
```
leplatrem: nitpick: `elif` is useless since the previous `if` ends with a `return`

kinto/core/utils.py
```diff
@@ -479,19 +479,54 @@ def parse_resource(resource):
...
+        if bucket == '' or collection == '' or record == '':
+            raise ValueError(error_msg)
```
leplatrem: same as above

kinto/core/initialization.py
```diff
@@ -412,10 +412,60 @@ def __call__(self, event):
...
+    def __init__(self, resource_ids, config):
+        self.resource_ids = resource_ids
+
+    def phash(self):
+        return 'for_resource_ids = {}'.format(','.join(self.resource_ids))
+
+    def __call__(self, event):
+        request = event.request
+        uri = event.payload.get('uri')
+        r_name, matchdict = utils.view_lookup(request, uri)
```
leplatrem: don't you have these already in the `event.payload` or attributes?

tests/core/test_listeners.py
aimanparvaiz: The reason why I decided to mock these is, the test (test_callback_called_when_action_is_not_filtered)  was breaking when the newly added class (EventResourceIdFilter) to filter by resource by ID calls view_lookup in utils.py. It expects a upath_info to be passed in request (which can be easily added in test_listener.py) and a non blank uri. I thought that this test is meant to check if the callback is being called or not and not a test view_lookup function. I might be wrong here, as I can surmise from your comment. Please guide me how to handle this blank URI issue then, assuming that adding upath_info in the Request class in test_listener.py file is ok. Also, you made this comment for this test only or for other tests too in this file where I am mocking these 2 functions?
Thanks

kinto/core/initialization.py
```diff
@@ -412,10 +412,60 @@ def __call__(self, event):
...
+        """ Special case for Create bucket """
+        # if event.payload.get('action') == 'create' and r_name == 'bucket':
+        # return True
+        for resource_ids in self.resource_ids:
```
leplatrem: Maybe you would not need to mock the requests attributes in the other tests that are unrelated to `resource ids` if `utils.view_lookup()` does not get called when `self.resource_ids` is falsy (empty list). You could add `if not self.resource_ids: return` on top and this way this filter is completely bypassed when unused.

(*my 2 cents*)

leplatrem:
@aimanparvaiz would you be interested in bringing this to the end? Or should someone take over?

Thanks!

aimanparvaiz:
@leplatrem Sorry, I havent been able to work on this for some personal reasons. Give me a couple of days to catch up, I would like to continue working but I might need some help around writing tests. I will update you in a couple of days if thats ok with you.

kinto/core/initialization.py
aimanparvaiz: This was helpful, thanks for pointing me in this direction.

tests/core/test_utils.py
aimanparvaiz: My understanding is that we need to execute custom code for create, update, delete actions for a  given bucket. This logic is implemented in parse_resource by checking len(parts)==3 hence test input "/fo/o" is not a malformed URL anymore since it satisfies the len(parts) condition in parse_resource. 
Let me know if I am wrong in my understanding of the requirement for triggering custom code around a given bucket or there is a different way you want me to handle this.

tests/core/test_listeners.py
aimanparvaiz: I might have a bug here, let me take a deeper look.

leplatrem:
@aimanparvaiz of course that's OK! You can take as much time as you need! I just wanted to check with you that you're still interested with this :)

kinto/core/initialization.py
aimanparvaiz: ok, going out in the next commit

kinto/core/utils.py
aimanparvaiz: yes, fixed

kinto/core/utils.py
aimanparvaiz: removed id_generator.match check

kinto/core/utils.py
aimanparvaiz: yep, I am keeping the id_generator check but removing the check for blank resources

kinto/core/utils.py
aimanparvaiz: yes

leplatrem:
The diff of this pull-request contains a lot of unrelated changes that are already on master. 
If you are familiar with git you can try to rebase your work and see if you can only limit the changes to your contribution.
If you need help please don't hesitate to ask, there should be a way to start clean without loosing commits!

aimanparvaiz:
Sure, I will take a shot at rebase over the weekend most probably.

aimanparvaiz:
@leplatrem Sorry for the late response. I rebased my work over master and as you can see the commit history is clean now with my 16 commits only. I see that PR is lacking test coverage around  ```instance_uri_registry``` in ```kinto/core/utils.py```. I need a little assistance here, I believe I am not the author of this code, was wondering what would be the best approach here.
Thanks

aimanparvaiz:
@leplatrem Sorry for the late response. I rebased my work and the commit history is clean now. The tests are failing because of a system issue I guess
```curl: (7) Failed to connect to s3.amazonaws.com port 443: Connection timed out
Unable to download 3.5 archive. The archive may not exist. Please consider a different version.```
Please let me know if I am missing something here. Thanks for your patience.

**Comment by leplatrem** [comment]
Thanks @aimanparvaiz for your efforts on this! Tests pass which is good news :) 

I don't have the brain lately to dig into the code, but I'll come back to it when I find the energy :)

File: kinto/core/utils.py [hunk 12]
```diff
@@ -498,22 +498,48 @@ def parse_resource(resource):
...
+    if resource_record:
+        if not id_generator.match(bucket) \
+                or not id_generator.match(collection) \
+                or not id_generator.match(record):
+            raise ValueError(error_msg)
+        return {
+            'bucket': bucket,
+            'collection': collection,
+            'record': record
+        }
```
glasserc: Oh, I didn't realize this was here! Why do we have both this and `view_lookup` (which is what I used when I wanted to parse URL-y things in kinto-changes)? I'm guessing based on your change that the event filtering stuff uses this code and not `view_lookup`. This function supports something that is sort of URL-y but also `<bid>/<cid>` which I think is kind of gross. Since we're introducing a new feature, I think it would be better if it could only parse URLs and not semi-URLs like this function does -- is it possible for this feature to use `view_lookup` instead of this? If it's difficult, let me know, maybe we can find another approach.

kinto/core/utils.py
```diff
@@ -498,22 +498,48 @@ def parse_resource(resource):
     :param str resource: a uri formatted /buckets/<bid>/collections/<cid> or <bid>/<cid>.
     :returns: a dictionary with the bucket_id and collection_id of the resource
     """
-
     error_msg = 'Resources should be defined as '
-    "'/buckets/<bid>/collections/<cid>' or '<bid>/<cid>'. "
+    "'/buckets/<bid>/collections/<cid>' or '<bid>/<cid>' or "
+    "'/buckets/<bid>/collections/<cid>/records/<rid>.'"
```
glasserc: I see that you preserved the extra period at the end of this message, but actually this isn't the end of the message (there's another line afterwards), so let's get rid of the period (assuming we have to use `parse_resource` at all).

tests/core/test_listeners.py
```diff
@@ -108,6 +108,264 @@ def test_callback_is_not_called_when_resource_is_filtered(self):
...
         self.assertFalse(self.demo_mocked.return_value.called)
 
+    # uri field is blank in utils.view_lookup
+    # and there us no upath_info field in request
+    @mock.patch('kinto.core.utils.parse_resource',
+                return_value={'bucket': 'bid',
+                              'collection': 'cid'})
+    @mock.patch('kinto.core.utils.view_lookup',
+                return_value=('record',
+                              {'bucket_id': 'bid', 'collection_id': 'cid'}))
```
glasserc: There's a `DummyRequest` class in kinto.core.testing that provides at least `upath_info`. Do we still need these mocks if we can use that instead? I'd rather test against a DummyRequest but with real utils methods than a real Request but with mocked utils methods.

leplatrem:
Looks like this PR has stalled..

@aimanparvaiz are you still interested to pushing it forward? Should someone else take over?

Thanks for everything you've done so far!

aimanparvaiz:
@leplatrem Sorry, wont be able to work on this in the near future, please feel free to take over or delegate this. Thanks for the opportunity.

PR #1518, opened 2018-02-26
Get rid of make build-requirements (fixes #1512)

### LINKED ISSUE #1512: Get rid of `make build-requirements`
During a conversation with @peterbe, we realized there were several ways of managing requirements.

Obviously, we agreed on the main necessity: reproducibility and environments consistency. We want to control the dependencies versions going to production.

Currently, in Kinto we have:
* the top-level dependencies (with optional range) in `setup.py`
* the whole dependencies list and their versions in `requirements.txt`
* we deploy kinto using the constraints of `requirements.txt` to keep deps under control
* pyup that opens a PR when a new dependency is available (which runs the tests)

So far, so good.

What was discussed is the fact that:

* The requirements.txt constraints are not used to run the tests on TravisCI
* The requirements.txt constraints are not used to initialize the developer local virtualenv
* We don't edit requirements.txt constraints manually. We run `make build-requirements` during the release process (which basically runs `pip install + freeze` to take the latest versions in range of `setup.py`) and rely on unit-tests (NB. we have 100% coverage)

This leads to situations where:

* Newcomers will fail at installing their dev environment because of an undesired upgrade
* Undesired upgrades can block unrelated contributions.
  * Example:
  * A contributor pushes some documentation changes, no change to the code base
  * CI fails because a backward incompatible upgrade was released and makes the code fail
  * We open a PR to fix the compatible version range in `setup.py`, and merge.
  * We can then merge/rebase the documentation PR
* Some dependency upgrade was not never ran locally by any developer. The dependency upgrade was just tested in unit/integration tests, and deployed.

Alternatively we could:
* Use the requirements constraints in dev + CI
* Remove the  `make build-requirements` part and let individual developers (or pyup) upgrade dependencies by editing `requirements.txt` manually (or using something like `hashin`)

Thoughts?

leplatrem:
Fixes #1512

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,10 @@ This document describes changes between each past release.
 - Fix missing principals from user info in root URL when default bucket plugin is enabled (fixes #1495)
 - Fix crash in Postgresql when the value of url param is empty (fixes #1305)
 
+**Internal Changes**
+
+- Now fully rely on Pyup.io (or contributors) to update the versions in the `requirements.txt` file (fixes #1512)
+
 
 8.1.5 (2018-02-09)
 ------------------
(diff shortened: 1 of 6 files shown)


leplatrem:
@Natim do read your notifications between 2 snorkeling sessions?

PR #1556, opened 2018-03-27
Show bucket:create permission in permissions endpoint (ref #1510)

### LINKED ISSUE #1510: Add creation of buckets in permissions endpoint
Currently with these settings:

```ini
kinto.experimental_permissions_endpoint = true
kinto.bucket_create_principals = system.Everyone  # or whatever 
```

I should see the permission to create buckets in http://localhost:8888/v1/permissions

Something like:
```js
{
  "data": [
    {
      "uri": "/buckets",
      "resource_name": "bucket",
      "permissions": [
        "create",
      ],
    }
  ]
}
```

This is because creating bucket is a specific case: it does not involve parents (buckets are top level) and does not involve the permission backend (permission can only be set in `.ini` settings).
Currently we only iterate on existing buckets https://github.com/Kinto/kinto/blob/bf60689ae016a31b8c5bb1955397be46d7b4acca/kinto/views/permissions.py#L91

Would allow to fix https://github.com/Kinto/kinto-admin/issues/401

stloma:
Fixes #1510 

For the `kinto/core/utils.py` change, `''` is required for the `resource_name` value in the dictionary lookup in the permissions tree:

https://github.com/Kinto/kinto/blob/6d77dcb2b94f9198fa08db489776440f7df0c275/kinto/views/permissions.py#L123

Since the purpose of the `view_lookup` function is to map a `uri` to a `resource_name`, I thought it best in there.

I was also going under the assumption that the permissions response for the top bucket should not return an id, so there are changes to that effect as well.

Everything else should be pretty straightforward and please let me know if you see any issues or suggestions.

Thanks!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ This document describes changes between each past release.
 - Move from importing pip to running it in a subprocess (see https://github.com/pypa/pip/issues/5081).
 - Remove useless print when using the OpenID policy (ref #1509)
 - Try to recover from the race condition where two requests can delete the same record. (Fix #1557; refs #1407.)
+- Add resource permissions defined in settings to permissions endpoint (ref #1510)
 
 
 8.2.0 (2018-03-01)
--- a/kinto/views/permissions.py
+++ b/kinto/views/permissions.py
@@ -82,6 +82,14 @@ class PermissionsModel:
         # Check settings for every allowed resources.
         from_settings = allowed_from_settings(self.request.registry.settings, principals)
 
+        # Add additional resources and permissions defined in settings/plugins
+        for root_perm in from_settings.get('', []):
+            resource_name, _ = root_perm.split(':')
+            uri = core_utils.strip_uri_prefix(
+                    self.request.route_path('{0}-collection'.format(resource_name)))
+            perms_by_object_uri[uri] = {root_perm}
+            perms_descending_tree[resource_name].update({root_perm: {resource_name: {root_perm}}})
+
         # Expand permissions obtained from backend with the object URIs that
         # correspond to permissions allowed from settings.
         allowed_resources = {'bucket', 'collection', 'group'} & set(from_settings.keys())
@@ -114,8 +122,10 @@ class PermissionsModel:
                 # Skip permissions entries that are not linked to an object URI
                 continue
 
-            # For consistency with event payloads, prefix id with resource name
-            matchdict[resource_name + '_id'] = matchdict.get('id')
+            # For consistency with event payloads, if resource has an id,
+            # prefix it with its resource name
+            if "id" in matchdict:
+                matchdict[resource_name + '_id'] = matchdict['id']
 
             # Expand implicit permissions using descending tree.
             permissions = set(perms)
--- a/tests/plugins/test_accounts.py
+++ b/tests/plugins/test_accounts.py
@@ -233,6 +233,7 @@ class AccountViewsTest(AccountsWebTest):
 
 
 class PermissionsEndpointTest(AccountsWebTest):
+
     @classmethod
     def get_app_settings(cls, extras=None):
         settings = super().get_app_settings(extras)
@@ -252,7 +253,53 @@ class PermissionsEndpointTest(AccountsWebTest):
         assert uris == ['/buckets/a/collections/b/records/c',
                         '/buckets/a/collections/b',
                         '/buckets/a',
-                        '/accounts/alice']
+                        '/buckets',
+                        '/accounts/alice',
+                        '/accounts']
+
+    def test_account_create_read_write_permissions_(self):
+        resp = self.app.get('/permissions', headers=self.headers)
+        buckets = resp.json['data']
+        accounts = buckets[4]
+        account = buckets[5]
+        accounts['permissions'].sort()
+        self.assertEqual(accounts, {
+            'account_id': 'alice',
+            'id': 'alice',
+            'permissions': ['read', 'write'],
+            'resource_name': 'account',
+            'uri': '/accounts/alice'})
+        self.assertEqual(account, {
+            'permissions': ['account:create'],
+            'resource_name': 'account',
+            'uri': '/accounts'})
+
+
+class PermissionsEndpointTestUnauthenticatedCreatePermission(AccountsWebTest):
+
+    @classmethod
+    def get_app_settings(cls, extras=None):
+        settings = super().get_app_settings(extras)
+        settings['experimental_permissions_endpoint'] = 'True'
+        settings['bucket_create_principals'] = 'system.Everyone'
+        return settings
+
+    def setUp(self):
+        self.everyone_headers = get_user_headers('')
+
+    def test_permissions_endpoint_is_compatible_with_accounts_plugin(self):
+        resp = self.app.get('/permissions', headers=self.everyone_headers)
+        buckets = resp.json['data']
+        bucket_create = buckets[0]
+        account_create = buckets[1]
+        self.assertEqual(bucket_create, {
+            'permissions': ['bucket:create'],
+            'resource_name': 'bucket',
+            'uri': '/buckets'})
+        self.assertEqual(account_create, {
+            'permissions': ['account:create'],
+            'resource_name': 'account',
+            'uri': '/accounts'})
 
 
 class AdminTest(AccountsWebTest):
--- a/tests/test_views_permissions.py
+++ b/tests/test_views_permissions.py
@@ -19,6 +19,59 @@ class PermissionsViewTest(BaseWebTest, unittest.TestCase):
         return settings
 
 
+class PermissionsUnauthenticatedViewTest(BaseWebTest, unittest.TestCase):
+
+    @classmethod
+    def get_app_settings(cls, extras=None):
+        settings = super().get_app_settings(extras)
+        settings['experimental_permissions_endpoint'] = 'True'
+        settings['bucket_create_principals'] = 'system.Everyone'
+        return settings
+
+    def setUp(self):
+        super().setUp()
+        self.everyone_headers = get_user_headers('')
+        self.app.put_json('/buckets/beers', MINIMALIST_BUCKET,
+                          headers=self.everyone_headers)
+        self.app.put_json('/buckets/beers/collections/barley',
+                          MINIMALIST_COLLECTION,
+                          headers=self.everyone_headers)
+        self.app.put_json('/buckets/beers/groups/amateurs',
+                          MINIMALIST_GROUP,
+                          headers=self.everyone_headers)
+        self.app.put_json('/buckets/beers/collections/barley/records/{}'.format(RECORD_ID),
+                          MINIMALIST_RECORD,
+                          headers=self.everyone_headers)
+
+        # Other user.
+        self.app.put_json('/buckets/water', MINIMALIST_BUCKET,
+                          headers=get_user_headers('alice'))
+
+    def test_permissions_list_entries_for_everyone(self):
+        resp = self.app.get('/permissions', headers=self.everyone_headers)
+        buckets = resp.json['data']
+        self.assertEqual(len(buckets), 5)
+
+    def test_bucket_create_permission_exists(self):
+        resp = self.app.get('/permissions', headers=self.everyone_headers)
+        buckets = resp.json['data']
+        toplevel_bucket = buckets[4]
+        self.assertEqual(toplevel_bucket, {
+            'permissions': ['bucket:create'],
+            'resource_name': 'bucket',
+            'uri': '/buckets'})
+
+    def test_object_details_are_provided(self):
+        resp = self.app.get('/permissions', headers=self.everyone_headers)
+        entries = resp.json['data']
+        for entry in entries:
+            if entry['resource_name'] == 'record':
+                record_permission = entry
+        self.assertEqual(record_permission['record_id'], RECORD_ID)
+        self.assertEqual(record_permission['collection_id'], 'barley')
+        self.assertEqual(record_permission['bucket_id'], 'beers')
+
+
 class EntriesTest(PermissionsViewTest):
 
     def setUp(self):
@@ -42,7 +95,7 @@ class EntriesTest(PermissionsViewTest):
     def test_permissions_list_entries_for_current_principal(self):
         resp = self.app.get('/permissions', headers=self.headers)
         permissions = resp.json['data']
-        self.assertEqual(len(permissions), 4)
+        self.assertEqual(len(permissions), 5)
 
     def test_permissions_can_be_listed_anonymously(self):
         self.app.patch_json('/buckets/beers/collections/barley',
@@ -77,7 +130,7 @@ class EntriesTest(PermissionsViewTest):
         resp = self.app.get('/permissions?in_resource_name=bucket,collection',
                             headers=self.headers)
         permissions = resp.json['data']
-        self.assertEqual(len(permissions), 2)
+        self.assertEqual(len(permissions), 3)
 
     def test_filtering_with_unknown_field_is_not_supported(self):
         self.app.get('/permissions?movie=bourne',
@@ -92,7 +145,7 @@ class EntriesTest(PermissionsViewTest):
     def test_permissions_list_can_be_paginated(self):
         resp = self.app.get('/permissions?_limit=2',
                             headers=self.headers)
-        self.assertEqual(resp.headers['Total-Records'], '4')
+        self.assertEqual(resp.headers['Total-Records'], '5')
         self.assertIn('Next-Page', resp.headers)
         self.assertEqual(len(resp.json['data']), 2)
 
@@ -168,6 +221,8 @@ class GroupsPermissionTest(PermissionsViewTest):
     def test_permissions_inherited_are_not_listed(self):
         resp = self.app.get('/permissions', headers=self.admin_headers)
         collections = [e for e in resp.json['data']
+                       # top level '/buckets' does not have an id
+                       if e['uri'] != '/buckets'
                        if e['bucket_id'] == 'beers' and e['resource_name'] == 'collection']
         self.assertEqual(len(collections), 0)
 
@@ -261,4 +316,4 @@ class DeletedObjectsTest(PermissionsViewTest):
 
     def test_deleted_objects_are_not_listed(self):
         resp = self.app.get('/permissions', headers=self.headers)
-        self.assertEqual(len(resp.json['data']), 0)
+        self.assertEqual(len(resp.json['data']), 1)
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -66,6 +66,7 @@ Contributors
 * Stanisław Wasiutyński <https://github.com/stanley>
 * Stephen Daves <contact@stephendaves.com>
 * Stephen Donner <stephen.donner@gmail.com>
+* Stephen Martin <lockwood@opperline.com>
 * Shweta Oak <oakshweta11@gmail.com>
 * Sofia Utsch <sofia.utsch@gmail.com>
 * Sumit Sarin <sumitsarinofficial@gmail.com>


kinto/core/utils.py
```diff
@@ -445,6 +445,11 @@ def view_lookup(request, uri):
     :rtype: tuple
     :returns: the resource name and the associated matchdict.
     """
+
+    # If we are accessing the top level listing of /buckets, there is no id
+    if (uri == '/buckets'):
+        return '', {id: False}
```
leplatrem: I'm pretty sure `view_lookup(request, '/buckets')` should already work as expected, ie. return `"bucket", {}`

If it does not, then we can fix it. You could add a unit test in test_utils to assert that BTW

kinto/views/permissions.py
```diff
@@ -114,8 +118,12 @@ def get_records(self, filters=None, sorting=None, pagination_rules=None,
...
                 continue
 
-            # For consistency with event payloads, prefix id with resource name
-            matchdict[resource_name + '_id'] = matchdict.get('id')
+            # For consistency with event payloads, prefix id with resource name. In
+            # the case that the resource is the top level bucket, do not include an id
+            if matchdict.get('id'):
+                matchdict[resource_name + '_id'] = matchdict.get('id')
+            else:
+                matchdict = {}
```
leplatrem: this would be simpler like this:
```py
if "id" in matchdict:
     matchdict[resource_name + '_id'] = matchdict['id']
```

kinto/views/permissions.py
```diff
@@ -125,6 +133,11 @@ def get_records(self, filters=None, sorting=None, pagination_rules=None,
                 # (e.g "bucket:write" gives "bucket:read" but not "group:read")
                 permissions |= obtained[resource_name]
 
+                # For consistency with event payloads, change top level bucket
+                # resource_name from '' to bucket
+                if resource_name == '':
+                    resource_name = 'bucket'
```
leplatrem: this would not be necessary if `view_lookup()` would return the right value

kinto/authorization.py
```diff
@@ -25,7 +25,9 @@
 # Ex: bucket:read is granted by both bucket:write and bucket:read
 PERMISSIONS_INHERITANCE_TREE = {
     '': {  # Granted via settings only.
-        'bucket:create': {},
+        'bucket:create': {
+            '': ['bucket:create', 'create']
+        },
```
leplatrem: Are you sure this is necessary? I would say that the code in `from_settings()` is enough.

tests/test_views_permissions.py
```diff
@@ -19,6 +19,44 @@ def get_app_settings(cls, extras=None):
...
+
+    @classmethod
+    def get_app_settings(cls, extras=None):
+        settings = super().get_app_settings(extras)
+        settings['experimental_permissions_endpoint'] = 'True'
+        settings['bucket_create_principals'] = 'system.Everyone'
+        return settings
+
+
+class EntriesTestUnauthenticated(PermissionsUnauthenticatedViewTest):
```
leplatrem: You don't need that class, you can just put the test into `PermissionsUnauthenticatedViewTest`

tests/test_views_permissions.py
```diff
@@ -19,6 +19,44 @@ def get_app_settings(cls, extras=None):
...
+        self.app.put_json('/buckets/beers/collections/barley/records/{}'.format(RECORD_ID),
+                          MINIMALIST_RECORD,
+                          headers=self.everyone_headers)
+
+        # Other user.
+        self.app.put_json('/buckets/water', MINIMALIST_BUCKET,
+                          headers=get_user_headers('alice'))
+
+    def test_permissions_list_entries_for_everyone(self):
+        print(self.headers)
```
leplatrem: leftover

tests/test_views_permissions.py
```diff
@@ -19,6 +19,44 @@ def get_app_settings(cls, extras=None):
...
+        resp = self.app.get('/permissions', headers=self.everyone_headers)
+        permissions = resp.json['data']
+        self.assertEqual(len(permissions), 5)
```
leplatrem: There needs to be a test somewhere that asserts the content of the bucket creation entry

tests/test_views_permissions.py
```diff
@@ -168,6 +206,8 @@ def test_permissions_granted_via_groups_are_listed(self):
     def test_permissions_inherited_are_not_listed(self):
         resp = self.app.get('/permissions', headers=self.admin_headers)
         collections = [e for e in resp.json['data']
+                       # top level '/buckets' does not have an id
+                       if e['uri'] != '/buckets'
```
leplatrem: syntax error?

kinto/core/utils.py
stloma: Yes `view_lookup` does return `"bucket", {}` without that code, the issue I'm seeing is that I need it to return `'', {}` for this:

https://github.com/Kinto/kinto/blob/6d77dcb2b94f9198fa08db489776440f7df0c275/kinto/views/permissions.py#L123

`perms_descending_tree` is populated with:

https://github.com/Kinto/kinto/blob/6d77dcb2b94f9198fa08db489776440f7df0c275/kinto/authorization.py#L26

So I need `''` for `resource_name` in the `obtained = perms_descending_tree[resource_name][perm]` code above.

That's what's causing some of the back and forth between assigning `bucket` and `''` in some of my other changes.

kinto/views/permissions.py
stloma: Yes absolutely, thanks!

tests/test_views_permissions.py
stloma: Thanks, this was my first time doing real testing, I'll clean up all of that.

tests/test_views_permissions.py
stloma: This is due to getting a KeyError since the top level bucket doesn't have an id

tests/test_views_permissions.py
stloma: Will do

kinto/views/permissions.py
stloma: From my comment on `kinto/core/utils.py`, I need `''` to get the permissions in:

https://github.com/Kinto/kinto/blob/6d77dcb2b94f9198fa08db489776440f7df0c275/kinto/views/permissions.py#L123

Then I have to switch it back to `bucket` for the response payload the user sees. If I didn't make this change, the response would be:

```
"data": [
        {
            "permissions": [
                "bucket:create"
            ],
            "resource_name": "",
            "uri": "/buckets"
        }
    ]
```

Rather than:

```
"data": [
        {
            "permissions": [
                "bucket:create"
            ],
            "resource_name": "bucket",
            "uri": "/buckets"
        }
    ]
```

kinto/core/utils.py
leplatrem: Ok, I get it.

So we should keep the `view_lookup()` consistent with what it does. `/buckets` is the `bucket` resource, so it should return.

Now, what we want is that to read `perms_descending_tree[""]["bucket:create"]`. So we should manage this corner here, and not in `view_lookup()`.

For example:
```py
if resource == "bucket" and permission == "create":
    obtained = perms_descending_tree[""]["bucket:create"]
```

That way we don't have to change `kinto.authorization.PERMISSIONS_INHERITANCE_TREE`, and we dont' change the semantics of `view_lookup()` for that particular. Just imagine that `view_lookup()` can be used by any other component and would not expect this `"", {"id": False}` response

kinto/authorization.py
```diff
@@ -25,7 +25,9 @@
 # Ex: bucket:read is granted by both bucket:write and bucket:read
 PERMISSIONS_INHERITANCE_TREE = {
     '': {  # Granted via settings only.
-        'bucket:create': {},
+        'bucket:create': {
+            '': ['bucket:create', 'create']
```
Natim: It doesn't makes sense to inherit `bucket:create` from `bucket:create` because we already have it.

Natim:
Thank you for tackling this :+1:

kinto/authorization.py
leplatrem: No, I don't think this `PERMISSIONS_INHERITANCE_TREE` dict should be touched in this PR. 
This dict gives the permission inheritance, and there is permission that is above `bucket:create` since it can be given via settings.

However in the permissions endpoint code, we can manipulate the `perms_descending_tree` and adjust the `from_settings()` function.

stloma:
Hey Natim, sure thing!

Sorry, I have a silly question, how do I push my changes here?

I've made changes on my local branch and read that to push them to a pull request, I would do a commit --amend, push them to my my branch, and they would automatically integrate into this pull request, but it looks like they're just showing as outdated here. What is the best method for this?

Natim:
> Sorry, I have a silly question, how do I push my changes here?


You can keep committing to your branch and it will automatically update the PR.

kinto/authorization.py
Natim: Oh this is because we don't inherit settings permission just yet ok...

stloma:
Ok great, thanks. I'm just not seeing the updated code here.

In any case, I was able to cleanup the code and a solution to the `PERMISSIONS_INHERITANCE_TREE` if you all think that it works.

I also took out the `view_lookup` code, thanks a lot for the explanation on that leplatrem, it made a lot of sense.

If you all have any more suggestions, please let me know, otherwise I'll start the test cleanup.

Natim:
> I'm just not seeing the updated code here.

I can't see it there too: https://github.com/stloma/kinto/tree/show-bucket-create
So I believe your push wasn't successful.

stloma:
Hmm, I'm seeing it there. I took out the `bucket:create` perm from `''` here for instance:

https://github.com/stloma/kinto/blob/fa9652583cff52233c022cb864f12ef3da8a4466/kinto/authorization.py#L27-L29

Natim:
This commit, https://github.com/stloma/kinto/commit/fa9652583cff52233c022cb864f12ef3da8a4466 is the one that is linked to that PR and that you pushed force. (It updated the PR too btw) so we are good. What misled me is the fact that it still has the time of the initial commit. Try to avoid using `git commit --amend``

stloma:
Will do, I'll just do a regular commit next time with the same message, thanks for the help :)

kinto/views/permissions.py
```diff
@@ -82,6 +82,13 @@ def get_records(self, filters=None, sorting=None, pagination_rules=None,
         # Check settings for every allowed resources.
         from_settings = allowed_from_settings(self.request.registry.settings, principals)
 
+        # If 'bucket_create_principals' is in settings, setup the '/buckets' uri with
+        # permissions and add the permission lookup
+        bucket_creation = from_settings.get('')
+        if bucket_creation:
+            perms_by_object_uri['/buckets'] = bucket_creation
+            perms_descending_tree[''] = {'bucket:create': {'': {'bucket:create'}}}
```
leplatrem: Why not do `perms_descending_tree['bucket'] = {'bucket:create': {'bucket': {'bucket:create'}}}` here and remove the if below ?

stloma: The issue I saw with that is the way you wrote it, `perms_descending_tree['bucket'] = {'bucket:create': {'bucket': {'bucket:create'}}}`, would overwrite the existing `bucket` permissions taken in from `PERMISSIONS_INHERITANCE_TREE`

We could do something like `perms_descending_tree['bucket'][''] = {'bucket:create': {'': {'bucket:create'}}}`, but then `obtained = perms_descending_tree[resource_name][perm]` would get messed up and we'd be back to another if statement.

stloma: No, you're right, I can do the following and remove the if statement:

`perms_descending_tree['bucket'].update({'bucket:create': {'bucket': {'bucket:create'}}})`

stloma: Just let me know if that's acceptable and I'll start on the tests.

leplatrem: yep!

kinto/views/permissions.py
```diff
@@ -82,12 +82,14 @@ def get_records(self, filters=None, sorting=None, pagination_rules=None,
...
-            perms_by_object_uri['/buckets'] = bucket_creation
-            perms_descending_tree[''] = {'bucket:create': {'': {'bucket:create'}}}
+        # Setup additional bucket permissions from settings
+        bucket_perms = from_settings.get('')
+        if bucket_perms:
+            for bucket_perm in bucket_perms:
+                perms_by_object_uri['/buckets'] = {bucket_perm}
+                perms_descending_tree['bucket'].update({
+                    bucket_perm: {'bucket': {bucket_perm}}
+                })
```
stloma: I found that the accounts plugin also populates `from_settings` with `account:create`, so I added a for loop as well that will allow for multiple settings.

leplatrem: Oh! Very good catch!

I think we'll have to do the same thing for accounts, I'm pretty sure the permission to create and write account does not appear in the permission endpoint. We should open an issue and fix it in another PR.

stloma:
Actually, hold off on this commit if you can :(

I've been looking into displaying if the user has account:create permissions for the /permissions endpoint that you mentioned and think I found a bug.

stloma:
Ok, fixed :)

I thought with both the accounts plugin and `kinto.bucket_create_principals = system.Everyone` it was working, but there was an issue where `account:create` was applying to the `/buckets` uri. I fixed that and adding more testing so it wouldn't happen again.

By doing it this way, the `/accounts` uri and `/accounts/{username}` is now showing on the permissions endpoint without creating an additional issue and without going outside of the scope of fixing the original issue that this pull request references.

It should be good to go!

CHANGELOG.rst
```diff
@@ -15,6 +15,7 @@ This document describes changes between each past release.
 - Now fully rely on Pyup.io (or contributors) to update the versions in the `requirements.txt` file (fixes #1512)
 - Move from importing pip to running it in a subprocess (see https://github.com/pypa/pip/issues/5081).
 - Remove useless print when using the OpenID policy (ref #1509)
+- Add creation of buckets in permissions endpoint (ref #1510)
```
leplatrem: And accounts ;)

kinto/views/permissions.py
```diff
@@ -82,6 +82,17 @@ def get_records(self, filters=None, sorting=None, pagination_rules=None,
         # Check settings for every allowed resources.
         from_settings = allowed_from_settings(self.request.registry.settings, principals)
 
+        # Add additional resources and permissions defined in settings/plugins
+        bucket_perms = from_settings.get('')
+        if bucket_perms:
+            for bucket_perm in bucket_perms:
```
leplatrem: This should probably be renamed `root_perms` now.
And I guess it can be simplified:
```py
for root_perm in from_settings.get('', []):
    resource_name, _ = ...
```

kinto/views/permissions.py
```diff
@@ -82,6 +82,17 @@ def get_records(self, filters=None, sorting=None, pagination_rules=None,
...
+                resource_name, _ = bucket_perm.split(':')
+                uri = '/{0}s'.format(resource_name)
```
leplatrem: This will work, but if we want to be completely safe here, it should be:
```py
uri = core_utils.strip_uri_prefix(self.request.route_path(f'{resource_name}-collection'))
```

tests/plugins/test_accounts.py
```diff
@@ -252,7 +253,52 @@ def test_permissions_endpoint_is_compatible_with_accounts_plugin(self):
...
+        self.assertEqual(len(accounts), 5)
+        self.assertEqual(len(account), 3)
+        self.assertEqual(accounts['permissions'], ['read', 'write'])
+        self.assertEqual(accounts['resource_name'], 'account')
+        self.assertEqual(accounts['uri'], '/accounts/alice')
+        self.assertEqual(accounts['account_id'], 'alice')
+        self.assertEqual(accounts['id'], 'alice')
+        self.assertEqual(account['permissions'], ['account:create'])
+        self.assertEqual(account['resource_name'], 'account')
+        self.assertEqual(account['uri'], '/accounts')
```
leplatrem: Why `account` singular if it is for `/accounts`?

tests/plugins/test_accounts.py
```diff
@@ -252,7 +253,52 @@ def test_permissions_endpoint_is_compatible_with_accounts_plugin(self):
...
+                        '/buckets',
+                        '/accounts/alice',
+                        '/accounts']
+
+    def test_account_create_read_write_permissions_(self):
+        resp = self.app.get('/permissions', headers=self.headers)
+        buckets = resp.json['data']
+        accounts = buckets[4]
+        account = buckets[5]
+        self.assertEqual(len(accounts), 5)
```
leplatrem: Here you test the number of keys, I think it would be just easier to compare the dict directly:
```py
self.assertEqual(account, {
    'permissions': ['read', 'write'],
    'resource_name', ...
    ...
})
```

tests/plugins/test_accounts.py
```diff
@@ -252,7 +253,52 @@ def test_permissions_endpoint_is_compatible_with_accounts_plugin(self):
...
+    def setUp(self):
+        self.everyone_headers = get_user_headers('')
+
+    def test_permissions_endpoint_is_compatible_with_accounts_plugin(self):
+        resp = self.app.get('/permissions', headers=self.everyone_headers)
+        buckets = resp.json['data']
+        bucket_create = buckets[0]
+        account_create = buckets[1]
+        self.assertEqual(len(bucket_create), 3)
+        self.assertEqual(len(account_create), 3)
```
leplatrem: Same here, I think you can compare dicts

tests/test_views_permissions.py
```diff
@@ -19,6 +19,59 @@ def get_app_settings(cls, extras=None):
...
+        self.assertEqual(len(buckets), 5)
+
+    def test_bucket_create_permission_exists(self):
+        resp = self.app.get('/permissions', headers=self.everyone_headers)
+        buckets = resp.json['data']
+        toplevel_bucket = buckets[4]
+        self.assertEqual(len(toplevel_bucket), 3)
+        self.assertEqual(toplevel_bucket['permissions'], ['bucket:create'])
+        self.assertEqual(toplevel_bucket['resource_name'], 'bucket')
+        self.assertEqual(toplevel_bucket['uri'], '/buckets')
```
leplatrem: Compare dicts here as well?

tests/plugins/test_accounts.py
stloma: I was following the naming convention for `resource_name`, i.e., `bucket`, `collection`, `group`, and `record`, so thought `account` would be appropriate for the `uri` `/accounts`.

Thanks for the rest of the input too, I'll incorporate them. I'll just wait for your thoughts on `account` vs `accounts` for `resource_name` and then send the changes.

leplatrem: ok I get it ! leave it like that sorry :)

stloma:
Hey leplatrem, sorry, sick of me yet? :)

My latest push is hitting a test failure that I didn't pick up on my local machine. Where we changed one of my tests to test by dictionary comparison:

```
self.assertEqual(accounts, {
    ...
   'permissions': ['write', 'read'],
  ...
})
```

It is hitting a sort of race condition (if that's the correct term) since the value of the `permissions` key is a set, thus unordered, so sometimes I get `'permissions': ['write', 'read']`, other times `'permissions': ['read','write']`. These values are populated by:

https://github.com/Kinto/kinto/blob/7ea5061c6dd3e8a452be0d901948ab55f14e18c3/kinto/views/permissions.py#L73

So all permissions displayed on the permissions endpoint, e.g., `buckets`, `collections`, etc, will come out in different orders.

For example:

```
 "bucket_id": "mybucket", 
            ....
            "permissions": [
                "read", 
                "collection:create", 
                "group:create", 
                "write", 
                "read:attributes"
            ], 
```

vs.

```
 "bucket_id": "mybucket", 
            ....
            "permissions": [
                "collection:create", 
                "read", 
                "read:attributes",
                "group:create", 
                "write"    
            ], 
```


I don't know if this is an issue or not, or whether I can just go ahead and change my test so it deals with it, I just wanted to bring it to your attention before I moved forward.

Also, please correct me if there's a better way to handle these issues that are coming up, I'm not trying to be a bother :)

glasserc:
Hi @stloma, I can't speak for @leplatrem but I took a few minutes to look through the other tests in the codebase and there seem to be a few in `test_object_permissions.py` that use `sorted()` to compare, probably for exactly this reason. I think that's the best approach. It's not exactly a race condition because it doesn't have to do with two processes happening at the same time, just that we use a `set()` so even within one process, order isn't guaranteed. JSON doesn't have any way to represent a `set()` except as a JSON Array, so it looks like there's an order when we really don't intend for there to be.

Don't worry about asking questions or leaving comments on a PR! We like to work collaboratively! Thanks for your continued hard work!

stloma:
Awesome, thanks for your help!

PR #1576, opened 2018-04-03
Fix permissions

glasserc:
kinto-http.js tests started failing because of #1556. This is a quick workaround but I'm not really convinced it's the best solution -- I feel like having entries (in the permissions list, or indeed anywhere) without an ID is asking for trouble. What do you think @leplatrem @stloma ?

### DIFF
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -306,7 +306,7 @@ def extract_record_set(records, filters, sorting,
     paginated = {}
     for rule in pagination_rules or []:
         values = list(apply_filters(filtered, rule))
-        paginated.update(dict(((x.get(id_field), x) for x in values)))
+        paginated.update(dict(((x[id_field], x) for x in values)))
 
     if paginated:
         paginated = paginated.values()
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -306,7 +306,7 @@ def extract_record_set(records, filters, sorting,
     paginated = {}
     for rule in pagination_rules or []:
         values = list(apply_filters(filtered, rule))
-        paginated.update(dict(((x[id_field], x) for x in values)))
+        paginated.update(dict(((x.get(id_field), x) for x in values)))
 
     if paginated:
         paginated = paginated.values()
--- a/tests/test_views_permissions.py
+++ b/tests/test_views_permissions.py
@@ -84,20 +84,18 @@ class PermissionsUnauthenticatedViewTest(BaseWebTest, unittest.TestCase):
 
         for order in ['id', '-id', 'uri', '-uri']:
             for limit in range(1, 10):
-                records = []
-                resp = self.app.get('/permissions?_sort={}&_limit={}'.format(order, limit),
-                                    headers=self.everyone_headers)
-                records = records + resp.json['data']
-                while 'Next-Page' in resp.headers:
-                    next_url = resp.headers['Next-Page'][len('http://localhost/v1'):]
-                    resp = self.app.get(next_url, headers=self.everyone_headers)
+                with self.subTest(order=order, limit=limit):
+                    records = []
+                    resp = self.app.get('/permissions?_sort={}&_limit={}'.format(order, limit),
+                                        headers=self.everyone_headers)
                     records = records + resp.json['data']
+                    while 'Next-Page' in resp.headers:
+                        next_url = resp.headers['Next-Page'][len('http://localhost/v1'):]
+                        resp = self.app.get(next_url, headers=self.everyone_headers)
+                        records = records + resp.json['data']
 
-                message = "didn't get same permissions with sort={}, limit={}".format(
-                    order, limit)
-                self.assertEqual(sort_by_uri(real_permissions),
-                                 sort_by_uri(records),
-                                 message)
+                    self.assertEqual(sort_by_uri(real_permissions),
+                                     sort_by_uri(records))
 
     def test_object_details_are_provided(self):
         resp = self.app.get('/permissions', headers=self.everyone_headers)
--- a/tests/test_views_permissions.py
+++ b/tests/test_views_permissions.py
@@ -68,11 +68,11 @@ class PermissionsUnauthenticatedViewTest(BaseWebTest, unittest.TestCase):
     def test_bucket_create_permission_exists(self):
         resp = self.app.get('/permissions', headers=self.everyone_headers)
         buckets = resp.json['data']
-        toplevel_bucket = buckets[4]
-        self.assertEqual(toplevel_bucket, {
+        toplevel_bucket = [b for b in buckets if b['uri'] == '/buckets']
+        self.assertEqual(toplevel_bucket, [{
             'permissions': ['bucket:create'],
             'resource_name': 'bucket',
-            'uri': '/buckets'})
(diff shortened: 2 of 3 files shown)


leplatrem:
That's exactly it.

We need to show whether the user has the permission to create root level objects (buckets, accounts so far). Since they don't have any parent object, there is no field we could use as `id`.

We could put a `"id": ""` in the payload on those if you think that's more consistent for consuming clients.

tests/test_views_permissions.py
```diff
@@ -61,6 +61,13 @@ def test_bucket_create_permission_exists(self):
             'resource_name': 'bucket',
             'uri': '/buckets'})
 
+    def test_can_paginate(self):
+        """kinto_http.js uses this sort of request when listing permissions"""
+        resp = self.app.get('/permissions?_sort=id&_limit=1', headers=self.everyone_headers)
+        while 'Next-Page' in resp.headers:
+            next_url = resp.headers['Next-Page'][len('http://localhost/v1'):]
+            resp = self.app.get(next_url, headers=self.everyone_headers)
```
leplatrem: Ideally, I'd like the `/buckets` perm to come out first, could you please check that? (I never know whether `None` comes first or not)

kinto/core/storage/memory.py
```diff
@@ -306,7 +306,7 @@ def extract_record_set(records, filters, sorting,
     paginated = {}
     for rule in pagination_rules or []:
         values = list(apply_filters(filtered, rule))
-        paginated.update(dict(((x[id_field], x) for x in values)))
+        paginated.update(dict(((x.get(id_field), x) for x in values)))
```
leplatrem: could be `x.get(id_field, "")` too ;)

leplatrem:
OK, after having chatted with @glasserc this fix won't work, cause it will squash the accounts and buckets root permissions (both with `"id": ""`)

I propose that we set `"id": "root-{resource_name}"` instead (eg. `"id": "root-bucket"`) and avoid safety checks around `id`.

It would have to be done in a `else` here:
https://github.com/Kinto/kinto/blob/71fed133948d18f8c423ef4319e2c231a9580f6e/kinto/views/permissions.py#L125-L128

glasserc:
Yes, just to confirm, with the accounts plugin, the `bucket:create` permission "wipes out" the `account:create` permission if both are present and we're paginating, except if the pagination border falls between them. Either way, though, the "total records" field shows the correct total. However, thinking about this more, I don't think it makes sense for there to be two separate `bucket:create` and `account:create` records -- these are both things that can be created under a single `root` element. Similarly to how `bucket` permissions can have both `group:create` and `collection:create` listed, I think there should be a single record with `resource_name` `root` that has both `account:create` and `bucket:create`.

I'm also seeing that *sometimes* I get permissions lists which include buckets `b1` and `b10` twice. I think this is related; if the pagination border hits between `account:create` and `bucket:create`, the pagination filter is coming in as comparing against `Missing`, which I bet is random depending on the memory address of `Missing` vs. other real objects. (We don't handle `Missing` correctly if it comes on the filter, but only if it comes on the record.) In other words, I bet records without `id` probably violate a lot of assumptions throughout kinto. I don't love the name `root-account` or `root-bucket`, but honestly it doesn't really matter too much what it is as long as there is one.

However, looking at this more closely, we have other problems with this code, which assumes that `id` is a primary key for each record -- which it *normally* is, when retrieving a list of some resource such as collections in some bucket, but isn't here. (For example, you can have `/buckets/abcd/collections/def` and `/buckets/abcd/groups/def`. Both have `id` "def", and get "merged" the same way `account:create` and `bucket:create` do.) The actual primary key here is something like `(record_name, id)`, but of course most of the time `record_name` isn't available. So probably we need to be much smarter here. I think we have two options here -- either we can stop using the memory backend's semi-internal function for this, in which case we need to pick actually-unique IDs (perhaps `url`?), or we can write another function to handle this somewhat different use case, re-using only the parts that make sense for both. At the moment I'm leaning towards using `url` as the `id`, but it feels kind of hacky.

tests/test_views_permissions.py
glasserc: Apparently it always came last (per the test just above). `None` comes first, but `Missing` comes last to mimic the Postgres backend.

glasserc:
To reproduce the `groups` / `collections` thing, try doing:

```
http PUT 'localhost:8888/v1/buckets/abcd' --auth "user:user"
http PUT 'localhost:8888/v1/buckets/abcd/groups/def' --auth "user:user"
http PUT 'localhost:8888/v1/buckets/abcd/collections/def' --auth "user:user"
http PUT 'localhost:8888/v1/buckets/abb' --auth "user:user"   # because we need pages of length two, but for both the group and collection to be on a page after the first one
http 'localhost:8888/v1/permissions?_sort=id&_limit=2' --auth "user:user"
# Then, take the `Next-Page` header and request that with the same auth. For me it was:
http 'http://localhost:8888/v1/permissions?_sort=id&_token=eyJvZmZzZXQiOjIsImxhc3RfcmVjb3JkIjp7ImlkIjoiYWJjZCIsInVyaSI6IlwvYnVja2V0c1wvYWJjZCJ9LCJub25jZSI6InBhZ2luYXRpb24tdG9rZW4tODJjZjE0ZjctOTIxNC00ZjMzLThlYjItZWExNzk5Yzc5ZmU4In0%3D&_limit=2' --auth user:user
```

Looking at this more, I guess we should probably keep the `id` available for places where the `id` is available, but we should probably use `uri` as the unique field. I think this should be possible with the code as written, but I'm not 100% sure yet.

tests/test_views_permissions.py
```diff
@@ -50,16 +53,51 @@ def setUp(self):
...
+        toplevel_bucket = [b for b in buckets if b['uri'] == '/buckets']
+        self.assertEqual(toplevel_bucket, [{
             'permissions': ['bucket:create'],
             'resource_name': 'bucket',
-            'uri': '/buckets'})
+            'uri': '/buckets'}])
+
+    def test_permissions_can_be_paginated(self):
+        """Verify that pagination doesn't squash same IDs"""
+        # kinto_http.js uses this sort of request when listing permissions
```
leplatrem: I think this comment should go below close to the `?sort={}` request

tests/test_views_permissions.py
```diff
@@ -50,16 +53,51 @@ def setUp(self):
...
+        real_permissions = self.app.get('/permissions', headers=self.everyone_headers).json['data']
+
+        def sort_by_uri(l):
+            return sorted(l, key=lambda r: r['uri'])
+
+        for order in ['id', '-id', 'uri', '-uri']:
```
leplatrem: I think it is better to use `pytest` parametrize function for this (better failure messages etc)

glasserc: This is a `unittest` flavored test, which doesn't have any concept of "parametrization". I can convert it to a `pytest` style test, but it seems a bit aggressive of a change. Alternately I can use metaprogramming to generate all the tests, but it seems a bit ugly. In the meantime I used `subTest`, which is a little bit like `parametrize` in that it annotates the test failure with the values that were being tested. What do you think?

leplatrem: You're right, leave it like this...

leplatrem:
I had never heard of `subTest()`! Nice!

PR #1579, opened 2018-04-04
Add an id to root level resources

stloma:
Here's the idea for the fix regarding #1576


leplatrem:
Thanks! 

I think we'll follow the approach in #1581 

We'll leave this open until we make a final decision :) 

Sorry for the confusion! That was definitely an interesting patch :))

glasserc:
Closing in favor of #1581.

PR #1581, opened 2018-04-04
Fix resource names for account:create and bucket:create permission in permissions endpoint

glasserc:
Refs #1576. At first I thought this was part of the same issue and was going to fix it in the same PR, but thinking about it more, I decided it was better to split it off. The basic inconsistency here is that `/buckets/foo` is `resource_name` `bucket`, with `permissions` such as `collection:create` and `group:create`, but the `bucket:create` permission is on `/buckets` and not `/`. This would be OK if the `collection:create` permission were on `/buckets/foo/collections`, but it isn't, and I'd argue it shouldn't be.

The third commit gets a bit more strident about reorganizing the codebase. Up until now, we have this mysterious empty string resource name representing the root of the hierarchy, but things can be a little simpler if we just call it `root`, since that's what it is. It's possible that you don't approve of this change -- the empty string is symmetric with the empty-string `parent_id` that we use as the parents for e.g. buckets. I consider these two things different -- things in the inheritance tree are really resource names, and things in parent_id are really URLs -- but I don't feel strongly about it and will unwind the change if you prefer.

### DIFF
--- a/kinto/authorization.py
+++ b/kinto/authorization.py
@@ -24,7 +24,7 @@ from kinto.core import authorization as core_authorization
 # automatically provides it
 # Ex: bucket:read is granted by both bucket:write and bucket:read
 PERMISSIONS_INHERITANCE_TREE = {
-    '': {  # Granted via settings only.
+    'root': {  # Granted via settings only.
         'bucket:create': {},
         'write': {},
         'read': {},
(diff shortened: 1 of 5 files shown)


glasserc:
Hmm. This is kind of intertwingled with #1576, so maybe wait until that lands before reviewing this.

glasserc:
I rebased this onto master and I think all the tests are correct now. r? @leplatrem

PR #1583, opened 2018-04-05
Account verification and cache

leplatrem:
I was looking at the account cache (introduced in https://github.com/Kinto/kinto/pull/1413, Kinto 8.2.0) and I have trouble understanding how this can be secure:

https://github.com/Kinto/kinto/blob/ca3092dc58e6fc80129572cdbb172a140cbd6080/tests/plugins/test_accounts.py#L136-L142

Can someone explain me why the password is not used to build the cache key?


glasserc:
Yes, I also saw that and wondered the same thing ... it did seem a bit strange to note that the username was "verified", without noting how it was verified.

leplatrem:
Apparently the key changes were introduced by this innocent commit: https://github.com/Stanley/kinto/commit/00ef4bf2e3fcccf94b1a0cbd9e45364d074d6378#diff-ba2d1b0729b3d63fcf9111d377936b47

@Natim  @Stanley I really need your insights here, but it looks like a hole.

On our side we don't have any Kinto 8.2.0 running with the account plugin in production, but I would advise those who do to downgrade to 8.1.X until a fix is released.

kinto/plugins/accounts/utils.py
```diff
@@ -1,9 +1,15 @@
 import bcrypt
 
 
+# XXX: THIS IS WRONG OF COURSE. But it is the only way I found to have the password
```
Natim: :joy:

leplatrem: But then I don't have it in the `on_account_changed` event to delete the cache entries.

In that case we could introduce a new feature in the cache backend to delete multiple keys with `*` (`account:alice:*:verified`)

kinto/plugins/accounts/authentication.py
```diff
@@ -11,15 +11,16 @@ def account_check(username, password, request):
     settings = request.registry.settings
     hmac_secret = settings['userid_hmac_secret']
     cache_key = utils.hmac_digest(hmac_secret, ACCOUNT_CACHE_KEY.format(username))
-    cache_ttl = settings.get('account_cache_ttl_seconds', 30)
+    offuscated_password = utils.hmac_digest(cache_key, password)
+    cache_ttl = int(settings.get('account_cache_ttl_seconds', 0))
```
Natim: Shall we keep 30 by default?

Natim:
I am going to land this and release. Thanks for noticing.

Issue #1584, opened 2018-04-05
Why do we restart pagination from the beginning?


glasserc:
While working on #1576, I saw that occasionally we would generate a pagination token that would not match any records (i.e. it was at the very end of the list of objects to paginate). When this happened, the memory backend would restart pagination at the beginning. However, I can't tell if this behavior is intentional or not and what the reasoning for it is.

One reason it might be intentional: you can hit this behavior if someone inserts elements at the beginning of a list you're in the middle of paginating. In that case, restarting pagination from the beginning is the only way to eventually get to all of the items.

One reason it might not be intentional: once you show the first page of objects again, we no longer show a `Next-Page` header, so it's not really "restarting pagination"; it's "you get the first page of results again". This might mean that you never actually see the elements that were inserted, if they were inserted in the middle.

Natim:
If I remember correctly the pagination flow is the following:

1. Get the first page and keep the last_modified
2. Go through Next-Page header
3. At the end try to get a ?_since=kept-last-modified (it should be empty or something new came by while you where going through pagination)

Natim:
The last step is not automatic it is something the client does.

glasserc:
OK. How does the client identify that it's "at the end"?

Natim:
Another thing I recall is that because of the continuation token, it was not possible to prevent having an empty page at the end if `total_number_of_items % _limit == 0`

Natim:
It's at the end because there are no more Next-Page header

glasserc:
I won't claim to be an expert in this code, but it seems like we only issue a continuation token if there are still some items yet unseen, so we never get an empty page.

#1585 changes the memory backend so that we don't repeat the first page of objects, but I think it's possible to get into a situation where we generate an infinite number of empty pages, each with a `Next-Page` headers.

glasserc:
I looked more closely at the pagination behavior we have and I no longer believe it's possible to get into an infinite-pagination situation. We only generate a next-page header if we have a completely full page, and each time we have a completely full page, we advance the offset in the next page. I had gotten confused because I thought we would continue until `offset + len(records) == total_records`, but that isn't the case (we continue until `len(records) < limit`).

Because #1585 has landed, I'm considering this fixed. Closing.

Natim:
Thank you for investigating this. So in case the total number of records % the limit == 0 we may have an empty page at the end but that's a known trade off of continuation tokens

glasserc:
Yes, I agree.

PR #1585, opened 2018-04-05
Don't restart pagination if we don't match any objects

glasserc:
An attempt to answer #1584 empirically (using the test suite).


Natim:
Shall we add a test for this?

PR #1588, opened 2018-04-06
Preparing release 8.2.3

Natim:


leplatrem:
It should also be 8.3.0 because we upgraded a bunch of dependencies :)

PR #1592, opened 2018-04-16
343 add contains filtering

### LINKED ISSUE #343: How to search for records containing item(s) is a list property?
Let's say I have the following records:

```
{
    "data": {
        "title": "My title",
        "tags": ["astronomie", "science", "burundi"]
    }
}
```

**What would be the request to look for all science in burundi?**

http://localhost:8000/v1/buckets/wiki/collections/articles/records?tags_contains=science,burundi

Natim:
Fixes #343

- [ ] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.


kinto/core/storage/memory.py
```diff
@@ -336,6 +342,8 @@ def apply_filters(records, filters):
         COMPARISON.IN: operator.contains,
         COMPARISON.EXCLUDE: lambda x, y: not operator.contains(x, y),
         COMPARISON.LIKE: lambda x, y: re.search(y, x, re.IGNORECASE),
+        COMPARISON.CONTAINS_ALL: contains_all_filtering,
+        COMPARISON.CONTAINS: lambda record_value, search_term: set(record_value[1]).intersection(set(search_term[1]))
```
Natim: I'd like to understand where this [1] comes from it is a bit unclear to me right now.

glasserc:
This is kind of bikesheddy, but I think `contains_foo=x,y` should mean "the foo attribute contains `["x", "y"]`" (i.e. what you called `contains_all` here). If you want to write "the foo attribute overlaps with `["x", "y"]`", I think you should write `intersect` or `contains_any`.

Natim:
If you have any idea why pip is failling, I would be interested in knowing.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -737,14 +739,30 @@ def _format_conditions(self, filters, id_field, modified_field,
...
+
+                array_type = "TEXT"
+                if len(value) > 0 and isinstance(value[0], int):
+                    array_type = "INT"
+
+                sql_operator = operators.setdefault(filtr.operator,
+                                                    filtr.operator.value)
+                # Thanks https://dba.stackexchange.com/a/157186
+                cond = "TRANSLATE(({})\:\:jsonb\:\:text, '[]','{}')\:\:{}[] {} :{}".format(
+                    sql_field, '{}', array_type, sql_operator, value_holder)
```
glasserc: I think this is all way too complicated and I am not crazy about the use of string processing to convert from one type to another. I think you should be converting the filter value to JSONB. This way we can support checking for fields containing complex objects like `[1, 2]` (in a field like `[[1,2], [5, 9]]`). There's no way to do "overlap" testing with this type, but I think that's OK -- do we really need `contains_any` at this point in time or did you just figure it was easy to implement at the same time?

glasserc:
Pip failing might not be your fault... on https://github.com/mozilla/pyjexl/pull/13 I'm getting "too many redirects".

kinto/core/storage/postgresql/__init__.py
Natim: > do we really need `contains_any` 

Yes I do

Natim: > I think this is all way too complicated and I am not crazy about the use of string processing to convert from one type to another.

That's the only way I found so far to be able to add it in the WHERE clause.

We could use `jsonb_to_text_array` but it would be have to be a JOIN.

If you find another way of doing it, I would be pleased to use it.

It was also a bit complex to use the SQLAlchemy execute parameters with an array.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -737,14 +739,28 @@ def _format_conditions(self, filters, id_field, modified_field,
...
                 cond = '{} {}'.format(sql_field, operator)
+
+            elif filtr.operator == COMPARISON.CONTAINS_ANY:
+                value_holder = '{}_value_{}'.format(prefix, i)
+                holders[value_holder] = value
+                # Postgres's && operator doesn't support jsonbs.
+                # However, it does support Postgres arrays of any
+                # type. Assume that the referenced field is a JSON
+                # array and convert it to a Postgres array.
+                # N.B. This won't work on JSON objects.
```
leplatrem: What does it mean "won't work" ? "will always be false" ?

docs/api/index.rst
```diff
@@ -11,6 +11,12 @@ API
 Changelog
 ---------
 
+1.19 (unreleased)
+'''''''''''''''''
+
+- New `contains_` and `contains_any` filter operators (fixes #343)
+
```
leplatrem: Could you please add them to https://github.com/Kinto/kinto/blob/master/docs/api/1.x/filtering.rst ?

CHANGELOG.rst
```diff
@@ -3,6 +3,15 @@ Changelog
 
 This document describes changes between each past release.
 
+
+8.4.0 (unreleased)
+------------------
+
+**New feature**
+
+- Add ``contains`` and ``contains_any`` filtering (Fixes #343).
```
leplatrem: nitpick: usually when we add new features to the HTTP API we highlight it as such (see older entries) :)

tox.ini
```diff
@@ -9,7 +9,6 @@ commands =
     py.test --cov-report term-missing --cov-branch --cov-fail-under 100 --cov kinto {posargs}
 deps =
     -rdev-requirements.txt
-    -crequirements.txt
```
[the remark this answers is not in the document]
Natim: Yeah maybe we should keep it.

kinto/core/storage/postgresql/__init__.py
Natim: If you've got an array of json objects it won't work. The array should be only strings or int.

docs/api/index.rst
Natim: Yes @jlebunetel is working on the documentation part :+1:

kinto/core/storage/postgresql/__init__.py
Natim: Oh I just understood your remark, I guess we should add a test to make sure of it.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -737,14 +739,28 @@ def _format_conditions(self, filters, id_field, modified_field,
...
                 cond = '{} {}'.format(sql_field, operator)
+
+            elif filtr.operator == COMPARISON.CONTAINS_ANY:
+                value_holder = '{}_value_{}'.format(prefix, i)
+                holders[value_holder] = value
+                # Postgres's && operator doesn't support jsonbs.
+                # However, it does support Postgres arrays of any
+                # type. Assume that the referenced field is a JSON
+                # array and convert it to a Postgres array.
+                # N.B. This won't work on JSON objects and will never match.
```
glasserc: No, that's not right.

```
http --auth 'user:user' PUT 'http://localhost:8888/v1/buckets/abc/collections/def/records/r1' 'data:={"user": {"first": "Ethan", "id": 1234}, "aliases": [{"ll": "ls -l"}, {"gti": "git"}, {"vi": "vim"}]}'
http --auth 'user:user' 'http://localhost:8888/v1/buckets/abc/collections/def/records?overlaps_aliases=[{"ll":"ls -l"}]'   # matches
http --auth 'user:user' 'http://localhost:8888/v1/buckets/abc/collections/def/records?overlaps_aliases=[{"sl":"ls -l"}]'   #doesn't match
http --auth 'user:user' 'http://localhost:8888/v1/buckets/abc/collections/def/records?overlaps_user={"id":1234}'   # 503
```

(This uses `overlaps` here because `contains_any` has a different bug...)

Compare with `contains`:

```
http --auth 'user:user' 'http://localhost:8888/v1/buckets/abc/collections/def/records?contains_aliases=[{"ll":"ls -l"}]'   # matches
http --auth 'user:user' 'http://localhost:8888/v1/buckets/abc/collections/def/records?contains_aliases=[{"sl":"ls -l"}]'   #doesn't match
http --auth 'user:user' 'http://localhost:8888/v1/buckets/abc/collections/def/records?contains_user={"id":1234}'  # matches
```

N.B. these behaviors may not match in the memory backend.

docs/api/1.x/filtering.rst
```diff
@@ -56,6 +56,33 @@ Sub-objects
...
+
+  Same as the previous filter, but in case of values are a list, matches all
+  records whose field contains at least one of the values listed.
+
+Examples:
+
+* ``/collection?contains_colors=[red,blue]``
+
+  Matches any record whose ``colors`` array field contains ``red`` and
+  ``blue`` elements.
```
glasserc: This doesn't work.

```
http --auth 'user:user' POST 'http://localhost:8888/v1/buckets/abc/collections/def/records' data:='{"name": "Superman", "colors": ["red", "blue"]}'
http --auth 'user:user' POST 'http://localhost:8888/v1/buckets/abc/collections/def/records?contains_colors=[red,blue]'
```

Doesn't match..

docs/api/1.x/filtering.rst
```diff
@@ -56,6 +56,33 @@ Sub-objects
...
+
+* ``/collection?contains_any_colors=[red,blue]``
+
+  Matches any record whose ``colors`` array field contains ``red`` or
+  ``blue`` elements.
```
glasserc: The same problem above is present here.

docs/api/1.x/filtering.rst
```diff
@@ -56,6 +56,33 @@ Sub-objects
...
 
+Search in array fields
+----------------------
+
+* ``/collection?contains_field=value``
+
+  Matches any records whose ``field`` array field contains ``value``. Values can
+  be an integer, a string, a list of integers or a list of strings. In case of
+  values are a list, only matches records whose field contains all the values
+  listed.
```
glasserc: What happens if ``field`` is not an array?

This can also match objects, but I guess it doesn't matter because almost any object contains you could consider doing is better expressed as a bunch of equality filters?

tests/core/resource/test_filter.py
```diff
@@ -219,6 +219,34 @@ def test_exclude_returns_400_if_value_has_wrong_type(self):
...
+        values = [item['colors'] for item in result['data']]
+        for value in values:
+            assert 'red' in value or 'blue' in value
+
+    def test_contains_any_with_integer(self):
+        self.validated['querystring'] = {'contains_fib': [3, 5]}
+        result = self.resource.collection_get()
+        values = [item['fib'] for item in result['data']]
+        for value in values:
+            assert 3 in value or 5 in value
```
glasserc: These tests are inaccurately named.

docs/api/1.x/filtering.rst
Natim: I guess we should raise a 400 for objects here.

tests/core/resource/test_filter.py
Natim: Whould `test_contains_with_string_select_records_with_color` be better? Any suggestion?

PR #1596, opened 2018-04-17
Expose bump_timestamp and bump_and_store_timestamp as part as the MemoryBasedStorage class.

Natim:
Fix [a XXX](https://github.com/Kinto/kinto/pull/1596/files#diff-73e8dd09932ff9ff7ea1678c7a44fc39L116) in the code.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,10 @@ This document describes changes between each past release.
 
 - Introduce ``contains`` and ``contains_any`` filter operators (fixes #343).
 
+**Internal changes**
+
+- Expose the manage_timestamp storage method so that memory based storage can use it.
+
 **Documentation**
 
 - Version number is taken from package in order to ease release process
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,22 +4,22 @@ Changelog
 This document describes changes between each past release.
 
 
-8.4.0 (unreleased)
+9.0.0 (unreleased)
 ------------------
 
 **Protocol**
 
 - Introduce ``contains`` and ``contains_any`` filter operators (fixes #343).
 
-**Internal changes**
+**Breaking changes**
 
-- Expose the manage_timestamp storage method so that memory based storage can use it.
+- Expose the ``bump_timestamp`` and ``bump_and_store_timestamp`` methods
+  so that memory based storage can use them. (#1596)
 
 **Documentation**
 
-- Version number is taken from package in order to ease release process
-- Copyright year is now dynamic
-
+- Version number is taken from package in order to ease release process (#1594)
+- Copyright year is now dynamic (#1595)
 
 
 8.3.0 (2018-04-06)
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -46,9 +46,9 @@ class MemoryBasedStorage(StorageBase):
     def set_record_timestamp(self, collection_id, parent_id, record,
                              modified_field=DEFAULT_MODIFIED_FIELD,
                              last_modified=None):
-        timestamp = self._bump_and_store_timestamp(collection_id, parent_id, record,
-                                                   modified_field,
-                                                   last_modified=last_modified)
+        timestamp = self.bump_and_store_timestamp(collection_id, parent_id, record,
+                                                  modified_field,
+                                                  last_modified=last_modified)
         record[modified_field] = timestamp
         return record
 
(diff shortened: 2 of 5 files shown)


kinto/core/storage/memory.py
```diff
@@ -113,36 +143,12 @@ def _bump_timestamp(self, collection_id, parent_id, record=None,
...
-        else:
-            # Not specified, use a new one.
-            current = utils.msec_time()
-            # If two ops in the same msec, bump it.
-            if current <= collection_timestamp:
-                current = collection_timestamp + 1
-            collection_timestamp = current
+        current_collection_timestamp = self._timestamps[parent_id].get(collection_id, 0)
 
+        current, collection_timestamp = self._manage_collection_timestamp(
```
Natim: There is one case where `current` and `collection_timestamp` are different it is when `current` is lower than `collection_timestamp`

kinto/core/storage/memory.py
```diff
@@ -67,6 +67,36 @@ def extract_record_set(self, records,
                                   pagination_rules=pagination_rules,
                                   limit=limit)
 
+    def _manage_collection_timestamp(self, collection_timestamp,
+                                     record, modified_field, last_modified):
```
leplatrem: If you expose it, it should not start with `_`, and probably have a doctstring with expected parameters and returned value ;)

kinto/core/storage/memory.py
```diff
@@ -113,36 +143,12 @@ def _bump_timestamp(self, collection_id, parent_id, record=None,
...
-            # If two ops in the same msec, bump it.
-            if current <= collection_timestamp:
-                current = collection_timestamp + 1
-            collection_timestamp = current
+        current_collection_timestamp = self._timestamps[parent_id].get(collection_id, 0)
 
+        current, collection_timestamp = self._manage_collection_timestamp(
+            current_collection_timestamp,
+            record, modified_field,
+            last_modified)
```
leplatrem: nit: name the arguments, it's easier to read/refactor

kinto/core/storage/memory.py
leplatrem: nit: modified_field could be last ;)

Natim: So we would have `bump_timestamp` and `_bump_timestamp`?

kinto/core/storage/memory.py
```diff
@@ -46,11 +46,16 @@ def strip_deleted_record(self, collection_id, parent_id, record,
...
-        record[modified_field] = timestamp
-        return record
+        current_collection_timestamp = self._timestamps[parent_id].get(collection_id, 0)
+        current, collection_timestamp = self.bump_timestamp(
+            collection_timestamp=current_collection_timestamp,
+            record=record,
+            last_modified=last_modified,
+            modified_field=modified_field,
+        )
+        self._timestamps[parent_id][collection_id] = collection_timestamp
```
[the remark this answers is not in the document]
Natim: What do you mean it doesn't belong here? It is part of the two places you wanted to remove the _bump_timestamp and add the new call to bump_timestamp.

Do you want me to revert and keep _bump_timestamp instead?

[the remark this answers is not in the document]
leplatrem: `self._timestamps` is not defined in `MemoryBased` so it should not be used here

CHANGELOG.rst
```diff
@@ -11,6 +11,10 @@ This document describes changes between each past release.
 
 - Introduce ``contains`` and ``contains_any`` filter operators (fixes #343).
 
+**Internal changes**
+
+- Expose the manage_timestamp storage method so that memory based storage can use it.
```
leplatrem: leftover ;)

kinto/core/storage/memory.py
```diff
@@ -46,9 +46,9 @@ def strip_deleted_record(self, collection_id, parent_id, record,
     def set_record_timestamp(self, collection_id, parent_id, record,
                              modified_field=DEFAULT_MODIFIED_FIELD,
                              last_modified=None):
-        timestamp = self._bump_timestamp(collection_id, parent_id, record,
-                                         modified_field,
-                                         last_modified=last_modified)
+        timestamp = self._bump_and_store_timestamp(collection_id, parent_id, record,
+                                                   modified_field,
+                                                   last_modified=last_modified)
```
leplatrem: If the `MemoryBasedStorage` relies on  a `self._bump_and_store_timestamp()`, then every child class will have to implement it. So it should not be prefixed with `_`. And should raise `NotImplementedError` etc. :)

Natim: Yes sure thing.

CHANGELOG.rst
```diff
@@ -4,18 +4,22 @@ Changelog
 This document describes changes between each past release.
 
 
-8.4.0 (unreleased)
+9.0.0 (unreleased)
```
leplatrem: :+1:

PR #1606, opened 2018-04-23
Make the AuthenticationPolicy name/prefix configurable.

Natim:


leplatrem:
Hmm. Can you explain why do we need this?

Why can't we stick with the current behaviour:
```ini
multiauth.policy.account.use = kinto_hawk.authentication.HawkAuthenticationPolicy
```

leplatrem:
I'm usually -1 on introducing numerous internal tweaking customizations that give users many options/strategies to choose from...  Unless we're really really stuck, I'd prefer to keep the current approach : auth policy name → user prefix.

CHANGELOG.rst
```diff
@@ -16,6 +16,10 @@ This document describes changes between each past release.
 - Expose the ``bump_timestamp`` and ``bump_and_store_timestamp`` methods
   so that memory based storage can use them. (#1596)
 
+**Internal changes**
+
+- Add the ability for a policy to define its name and prefix for the prefixed_userid.
```
leplatrem: `Authentication policies can now hard code and override the name specified in settings`

tests/core/test_authentication.py
```diff
@@ -33,6 +33,17 @@ def test_basic_auth_is_declined_if_disabled_in_settings(self):
...
         assert 'basicauth' not in resp.json['capabilities']
 
+    @mock.patch('kinto.core.authentication.BasicAuthAuthenticationPolicy')
+    def test_policy_name_is_used(self, basicAuth):
+        basicAuth.return_value.name = "foobar"
+        app = self.make_app({
+            'multiauth.policies': 'dummy',
+            'multiauth.policy.dummy.use': ('kinto.core.authentication.'
+                                           'BasicAuthAuthenticationPolicy')})
+        # Check that the capability is exposed on the homepage.
```
leplatrem: comment is wrong

kinto/core/initialization.py
```diff
@@ -114,7 +114,7 @@ def setup_authentication(config):
     # Track policy used, for prefixing user_id and for logging.
     def on_policy_selected(event):
         request = event.request
-        authn_type = event.policy_name.lower()
+        authn_type = getattr(event.policy, 'name', event.policy_name.lower())
```
leplatrem: Add a small comment :  `If the authn policy has a `name` attribute, we ignore the name specified in settings`

leplatrem:
> The second rationale is because for Hawk I want to define two authentication backend that gives the same user ID: for Basic Auth and one for Hawk

Hmm, it sounds super complicated when reading this...

Natim:
> it sounds super complicated when reading this...

It's not: https://github.com/Kinto/kinto-hawk/#installation

```
multiauth.policies = account hawk

# Enable Account authenticated policy.
multiauth.policy.account.use = kinto.plugins.accounts.authentication.AccountsAuthenticationPolicy

# Enable Hawk authenticated policy.
multiauth.policy.hawk.use = kinto_hawk.authentication.HawkAuthenticationPolicy
```

leplatrem:
Would be easier to have

```ini
multiauth.policies = account 
multiauth.policy.account.use = kinto_hawk.authentication.HawkAuthenticationPolicy
```

PR #1608, opened 2018-04-24
Update kinto-admin to 1.16.0

Natim:
Fixes https://github.com/Kinto/kinto-admin/issues/526


leplatrem:
Some stuff is missing on the kinto-admin release page, I'll add them

PR #1612, opened 2018-04-25
Simplify native_value

glasserc:
Even if the JSON parse succeeds, we end up with the same string value
we started with.


Natim:
nice catch

Issue #1205, opened 2017-05-02
Throwing an exception in ResourceChanged fails the entire batch?


glasserc:
I tried writing a plugin like this:

```python
from pyramid import httpexceptions

from kinto.core.events import ACTIONS, ResourceChanged


def maybe_block_delete(event):
    for change in event.impacted_records:
        collection = change['old']
        if collection['id'] == 'storage-sync-crypto':
            raise httpexceptions.HTTPMethodNotAllowed("You may not delete the storage-sync-crypto collection.")


def includeme(config):
    # Redirect default to the right endpoint
    config.add_subscriber(maybe_block_delete, ResourceChanged,
                          for_resources=('collection',),
                          for_actions=(ACTIONS.DELETE,))

    config.add_api_capability(
        "dont_delete_crypto",
        description="Try to prevent deletion of storage-sync-crypto collections.",
        url="https://bugzilla.mozilla.org/show_bug.cgi?id=1359879")
```

I tried to test it like this:

```
http -v -j --auth 'user:pass2' POST localhost:8888/v1/batch 'requests:=[{"method": "PUT", "path": "/buckets/new-bucket/collections/my-collection-2"}, {"method":"DELETE", "path":"/buckets/default/collections/my-collection"}, {"method": "DELETE", "path": "/buckets/default/collections/storage-sync-crypto"}]'
```

I got:

```
HTTP/1.1 405 Method Not Allowed
Access-Control-Expose-Headers: Backoff, Content-Length, Retry-After, Alert
Content-Length: 102
Content-Type: application/json
Date: Mon, 01 May 2017 22:16:38 GMT
Server: waitress
X-Content-Type-Options: nosniff

{
    "code": 405,
    "errno": 115,
    "error": "Method Not Allowed",
    "message": "Method not allowed on this endpoint."
}
```

Is this the correct behavior? I assumed that only the one request would fail, similarly to how it would if I had put a nonmatching If-Match on a single request. On the other hand, maybe the entire batch has to fail because I'm rejecting the transaction?

glasserc:
OK, this seems like it's a duplicate of #624 then, thanks, closing.

PR #1620, opened 2018-05-03
503 error when expected 409 constraint violation to batch endpoint fixes 1569

### LINKED ISSUE #1569: IntegrityError if you try to create two records with the same last_modified value in a batch
```
demo:
  collections:
    demo:
      records:
        abc:
          data:
            last_modified: 123
        efg:
          data:
            last_modified: 123
```

```
$ kinto-wizard load -s https://kinto.dev.mozaws.net/v1 -a admin:admin  demo.yaml --force -b demo -c demo
kinto_http.exceptions.KintoException: POST /v1/batch - 503 503 - 
{'message': 'Service temporary unavailable due to overloading or maintenance, please retry later.',
 'code': 503, 'errno': 201, 'error': 'Service Unavailable'}
```

stloma:
Fixes #1569 

I believe this is the best way to fix this error. If agreed, I'll write the tests.

The issue was caused when Kinto processed a batch request. If one of the requests failed, e.g., a contraint violation, the `parent` process would not know and continue on into event notifications, e.g., `ResourceChanged`, after its child failed.

This wasn't triggering any noticable errors until the history plugin was turned on. In this case, `history/listener.py` was triggered from a change event and would try to query the database while `psycopg2` was in the following error state from the constraint violation: `current transaction is aborted, commands ignored until end of transaction block and can’t do anything else!` It was this condition that caused Kinto to return `503: Service temporary unavailable due to overloading or maintenance, please retry later.`

The solution I came up with was to use Pyramid's builtin transaction manager to `doom` the `parent` process so it would not continue on into triggering event notifications.

### DIFF
--- a/kinto/core/views/batch.py
+++ b/kinto/core/views/batch.py
@@ -144,13 +144,13 @@ def post_batch(request):
             resp, subrequest = request.follow_subrequest(subrequest,
                                                          use_tweens=False)
         except httpexceptions.HTTPException as e:
+            # Since some request in the batch failed, we need to stop the parent request
+            # through Pyramid's transaction manager.
+            if e.status_code == 409 or e.status_code >= 500:
+                request.tm.abort()
+
             if e.content_type == 'application/json':
                 resp = e
-                # Since some request in the batch failed, we need to "doom" the parent process
-                # through Pyramid's transaction manager.
-                transaction_manager = request.tm
-                transaction_manager.begin()
-                transaction_manager.doom()
             else:
                 # JSONify raw Pyramid errors.
                 resp = errors.http_error(e)
--- a/kinto/core/views/batch.py
+++ b/kinto/core/views/batch.py
@@ -145,8 +145,10 @@ def post_batch(request):
                                                          use_tweens=False)
         except httpexceptions.HTTPException as e:
             # Since some request in the batch failed, we need to stop the parent request
-            # through Pyramid's transaction manager.
-            if e.status_code == 409 or e.status_code >= 500:
+            # through Pyramid's transaction manager. 5XX errors are already caught by
+            # pyramid_tm's commit_veto
+            # https://github.com/Kinto/kinto/issues/624
+            if e.status_code == 409:
                 request.tm.abort()
 
             if e.content_type == 'application/json':
--- a/kinto/core/views/batch.py
+++ b/kinto/core/views/batch.py
@@ -146,6 +146,11 @@ def post_batch(request):
         except httpexceptions.HTTPException as e:
             if e.content_type == 'application/json':
                 resp = e
+                # Since some request in the batch failed, we need to "doom" the parent process
+                # through Pyramid's transaction manager.
+                transaction_manager = request.tm
+                transaction_manager.begin()
+                transaction_manager.doom()
             else:
                 # JSONify raw Pyramid errors.
                 resp = errors.http_error(e)
--- a/tests/core/test_views_transaction.py
+++ b/tests/core/test_views_transaction.py
@@ -86,9 +86,8 @@ class TransactionTest(PostgreSQLTest, unittest.TestCase):
         resp = self.app.post_json('/batch', body, headers=self.headers)
         response = resp.json['responses'][1]
         self.assertEqual(response['status'], 409)
-        resp = self.app.get('/buckets/mushrooms', headers=self.headers)
+        resp = self.app.get('/mushrooms', headers=self.headers)
         self.assertEqual(len(resp.json['data']), 0)
-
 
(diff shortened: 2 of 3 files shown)


stloma:
Local tests weren't running correctly I guess, I'll have to rethink this fix. The failing test might present a problem to my solution.

kinto/core/views/batch.py
```diff
@@ -146,6 +146,11 @@ def post_batch(request):
         except httpexceptions.HTTPException as e:
             if e.content_type == 'application/json':
                 resp = e
+                # Since some request in the batch failed, we need to "doom" the parent process
+                # through Pyramid's transaction manager.
+                transaction_manager = request.tm
+                transaction_manager.begin()
+                transaction_manager.doom()
```
leplatrem: I think it needs to check the `resp.status` (eg. if 5XX) like does the commit veto (see https://github.com/Pylons/pyramid_tm/blob/ffed3d3e352b800064dbe9e1f0270f11fdda63ff/pyramid_tm/__init__.py#L50) 

Also, why should it only apply to exception with content-type JSON ? Why should we `.begin()` again?

If you want to add some new tests, you'll find some in `core/test_views_transaction.py`

stloma: Yes, the resp.status check is a good idea. I'm trying to figure out how that should look and there seems to have been some discussion in the past on how a response should respond when a sub-response breaks.

It seems from the following discussions that it should be assumed when a sub-response breaks with a 4XX, the entire batch should come back with a 4XX:

https://github.com/Kinto/kinto/issues/1205
https://github.com/Kinto/kinto/issues/624

I guess I'm confused as the following test seems to contridact that:

https://github.com/Kinto/kinto/blob/382adcdc2dccff7151798bc7d250769dc818d83e/tests/core/test_views_transaction.py#L59

If the former is true, I will need to check that the `resp.status` is 4XX (as this contraint violation is a 409) and 5XX, but not 401 as the test asserts 401s should not effect other responses being committed.

stloma: As far as the `begin()`, short answer is I used it because it didn't work without it :) As far as my understanding, I agree, it shouldn't be necessary to use it. `abort()` does work without a `begin()`, but I'll look into this and get an answer.

stloma:
When I looked further into why I needed the `begin` method before the `doom` method in order to halt the request, it seems like `begin` was the one actually aborting the request; `doom` actually just saves a `isDoomed` boolean onto the request for checks later, but doesn't abort the transaction.

So, I added the `abort` method which does what I wanted.

kinto/core/views/batch.py
```diff
@@ -144,6 +144,11 @@ def post_batch(request):
             resp, subrequest = request.follow_subrequest(subrequest,
                                                          use_tweens=False)
         except httpexceptions.HTTPException as e:
+            # Since some request in the batch failed, we need to stop the parent request
+            # through Pyramid's transaction manager.
+            if e.status_code == 409 or e.status_code >= 500:
```
leplatrem: Are you sure the 500 is necessary here? I'm really not sure, since the final batch response status will be 5XX, the classic pyramid_tm commit veto is going to play its role.
(also don't hesitate to add this as a comment..)

stloma:
I still need to cleanup the tests a bit and understanding the `v0` endpoint. If there are any suggestions if that's the correct or incorrect way to do it, I would appreciate it!

tests/core/test_views_transaction.py
```diff
@@ -71,6 +71,25 @@ def test_modifications_are_not_rolled_back_on_4XX_error(self):
...
+        bucket_ccreate = {
+            'method': 'POST',
+            'path': '/mushrooms',
+            'body': {'data': {'name': 'Vesse de loup', 'last_modified': 123}}
+        }
+        body = {'requests': [bucket_create, bucket_ccreate]}
+        resp = self.app.post_json('/batch', body, headers=self.headers)
+        response = resp.json['responses'][1]
+        self.assertEqual(response['status'], 409)
+        resp = self.app.get('/buckets/mushrooms', headers=self.headers)
```
leplatrem: Here we are in `kinto.core`, there is no notion of `buckets`. Try with `/mushrooms`

kinto/core/views/batch.py
```diff
@@ -144,6 +144,11 @@ def post_batch(request):
             resp, subrequest = request.follow_subrequest(subrequest,
                                                          use_tweens=False)
         except httpexceptions.HTTPException as e:
+            # Since some request in the batch failed, we need to stop the parent request
+            # through Pyramid's transaction manager.
```
leplatrem: Can you please add a link to this issue please : #624

Issue #1616, opened 2018-04-26
Do we need to support get_all with a wildcard parent_id?


glasserc:
While investigating #1507, @peterbe discovered that the `DISTINCT` doesn't really make sense unless you have a wildcard `parent_id`. He proved this to his own satisfaction by removing `DISTINCT` and observing that no tests break.

Under what circumstances do we even have a wildcard `parent_id`? Unlike `delete`, where you might want to delete a thing and all its children, I couldn't think of a way to invoke this mechanism using the HTTP API. Indeed, there's a subtle bug in the current query and the fact that nobody has reported it makes me suspect that nobody ever actually uses it. Can we get rid of it?

leplatrem:
My position on stuff like this has always been: tests are specifications. If tests don't fail you can remove whatever you want :)

glasserc:
It turns out there actually are two tests that cover `get_all` with a wildcard `parent_id`. But my assertion is that outside of those tests, nobody actually uses it. (The only code path I could see is if someone creates a `Model` subclass with a `parent_id` that has a wildcard in it.) What if I delete the functionality and the two tests that cover it, can I still remove it? :grinning:

Natim:
Are you sure it's not a feature that we added to support the quota features?

leplatrem:
Can you give us pointers to the tests?

Which PR / blame ? Any mention in CHANGELOG? Are those tests present in the original cliquet repo?

leplatrem:
OK, I grep the usage of `get_all` and found this in the CHANGELOG:
`The storage backend now allows ``parent_id`` pattern matching in ``kinto.core.storage.get_all``. (#821)`

@Natim implemented it and I approved it in https://github.com/Kinto/kinto/pull/821

leplatrem:
I found its usage by searching "pattern matching" :) 

https://github.com/Kinto/kinto/blob/5904085b9ce87ea8772fe2fb19cf4b8c00db03a8/kinto/plugins/accounts/views.py#L82-L83

I remember now that we suffered when implementing the accounts parent id stuff. Each account entry is isolated using `parent_id=user_id`. As admin, I can see every entries, thus `parent_id=*`. 

There might other ways to implement this, and I don't think the «account admin» feature is important enough to justify having lower perfs on `get_all()`

Natim:
I also found why I implemented it in the first place:

[...]

PR #1622, opened 2018-05-04
Cleanup Postgres get_all

glasserc:
This is a more conservative approach towards addressing #1507. In particular, it doesn't introduce any race conditions, and probably reduces query times by much less.

The existing query's performance comes down to three factors:

1. It matches every record in the database that matches the queries, and outputs them in any random order.
2. It then sorts those records according to the order given in the sort order.
3. It does a distinct/join as part of paginating.

The performance characteristics of these steps varies depending on how many records the filters match and how many they don't match, as well as what proportion of the database is under the given parent_id.

1. If most of the database is under this parent, then doing a sequential scan is the fastest way to do this. Otherwise, you want to use an index scan to narrow it down as much as possible before applying filters.
2. Sorting is O(n log n), so if a lot of records match the filters, doing a sort afterwards is expensive. In those cases, you generally want to do an index scan to produce the records in the right order to begin with.
3. Doing a distinct/join is probably O(n log n), so it becomes more expensive the more records match.

Per our discussion in #1616, there are a couple of subtle bugs in the use of wildcard `parent_id`, which I used as the crowbar to try to pry open this query. First, introduce some tests to firm up the behavior we expect for wildcard `parent_ids`. It turns out that the third step above introduces a bug in pagination of results when doing a wildcard `parent_id` match, so just delete it.

Doing the sorting at the end seems wasteful to me; you need to sort anyhow in order to do the pagination, so why not generate records in the right order to begin with? I moved the 2nd step into the first step. This encourages Postgres to use indexes when called for. However, it seems that the benefit to the second step of using an index is outweighed by the cost to the first step of doing an index scan when very few records match, so in a Kinto instance with all records under one parent, doing a query with filters that match comparatively few records, this query is slower than the previous one.

I tested this by filling a Kinto instance with about 10m records, all in the same collection, each of the form `{"id": "id-XXXXXX", "my_data": XXXXXX}`, and then PATCHing one record to add a `"foo": true` field to it. I then measured the time of `http GET  'localhost:8888/v1/buckets/a/collections/b/records?_limit=10'`, `http GET  'localhost:8888/v1/buckets/a/collections/b/records?_limit=10&has_foo=true'`, and `http GET  'localhost:8888/v1/buckets/a/collections/b/records?_limit=10&has_foo=false'`. The first and last requests tend to perform the same, because both match the vast majority of the records in the database; on these, the new query takes about 60s (to the old query's ~90s, so about 33% faster). The second query is a little slower than the previous one (4.52s to 2.65s, or about 80% slower).

Perhaps it's possible to get the benefit of both queries by using heuristics to choose the sequential scan or the index scan, but then we're in the situation of trying to beat the Postgres query planner and I'm not sanguine about our chances. On the other hand, note that the Postgres query planner is hamstrung by the lack of indexing on the `data` column.

My conclusions are:

- It isn't really clear what the scalability characteristics of Kinto are or should be. Is Kinto meant to store 10m records? I would say probably not. What about 1m records, but all in one collection? I'm not certain.
- It isn't really clear (to me, at least) what queries Kinto should support as "first-class" and which are "gravy", i.e. added bonuses that we threw in because they were easy to implement but are not as well-loved, performant etc. as others. Naïvely I would say syncing collections is more important than supporting arbitrary filters, but then again, if that's what you want, CouchDB is available and may be a better fit for you.
- It isn't really clear to me what the relative priorities are in the Kinto ecosystem. As an example, in order to produce a `Total-Records` header, we have to add a `COUNT` to our queries, which means enumerating all the matching records rather than just the ones that we're going to render in the response. In other distributed systems, you might just paginate until you get an empty page. Is the ease-of-use/sanity checking provided by the `Total-Records` header worth the performance cost?

- [ ] Add documentation.
- [x] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -319,11 +319,10 @@ def extract_record_set(records, filters, sorting,
     total_records = len(filtered)
 
     if pagination_rules:
-        paginated = {}
+        paginated = []
         for rule in pagination_rules:
             values = apply_filters(filtered, rule)
-            paginated.update((id(x), x) for x in values)
-        paginated = paginated.values()
+            paginated.extend(values)
     else:
         paginated = filtered
 
--- a/kinto/core/storage/memory.py
+++ b/kinto/core/storage/memory.py
@@ -322,7 +322,7 @@ def extract_record_set(records, filters, sorting,
         paginated = {}
         for rule in pagination_rules:
             values = apply_filters(filtered, rule)
-            paginated.update(((x[id_field], x) for x in values))
+            paginated.update((id(x), x) for x in values)
         paginated = paginated.values()
     else:
         paginated = filtered
--- a/kinto/core/storage/testing.py
+++ b/kinto/core/storage/testing.py
@@ -297,6 +297,19 @@ class BaseTestStorage:
         self.assertEquals(len(records), 1)
         self.assertEquals(len(records), total_records)
 
+    def test_get_all_parent_id_handles_collisions(self):
+        abc1 = self.create_record(parent_id='abc1', collection_id='c',
+                                  record={'id': 'abc', 'secret_data': 'abc1'})
+        abc2 = self.create_record(parent_id='abc2', collection_id='c',
+                                  record={'id': 'abc', 'secret_data': 'abc2'})
+        records, total_records = self.storage.get_all(parent_id='ab*', collection_id='c',
+                                                      include_deleted=True)
+        self.assertEquals(len(records), 2)
+        self.assertEquals(len(records), total_records)
+        records.sort(key=lambda record: record['secret_data'])
+        self.assertEquals(records[0], abc1)
+        self.assertEquals(records[1], abc2)
+
     def test_get_all_return_all_values(self):
         for x in range(10):
             record = dict(self.record)
--- a/kinto/core/storage/testing.py
+++ b/kinto/core/storage/testing.py
@@ -1541,6 +1541,8 @@ class DeletedRecordsTest:
     def test_get_all_parent_id_paginates_correctly(self):
         """Verify that pagination doesn't squash or duplicate some records"""
 
+        # Create records with different parent IDs, but the same
+        # record ID.
         for parent in range(10):
             parent_id = 'abc{}'.format(parent)
             self.storage.create(parent_id=parent_id, collection_id='c',
@@ -1555,15 +1557,13 @@ class DeletedRecordsTest:
         GT = utils.COMPARISON.GT
         LT = utils.COMPARISON.LT
         for order in [('secret_data', 1), ('secret_data', -1)]:
-            order_field, order_direction = order
             sort = [Sort(*order), Sort('last_modified', -1)]
-            pagination_direction = GT if order_direction == 1 else LT
             for limit in range(1, 10):
                 with self.subTest(order=order, limit=limit):
                     records = []
                     pagination = None
                     while True:
-                        (page, total_records) = self.storage.get_all(
+                        page, total_records = self.storage.get_all(
                             parent_id='abc*', collection_id='c', sorting=sort,
                             limit=limit, pagination_rules=pagination)
 
@@ -1576,7 +1576,11 @@ class DeletedRecordsTest:
                         # IndexError.
                         if not page:  # pragma: nocover
                             break
+                        # Simulate paging though the records as
+                        # though following the logic in Resource._build_pagination_rules.
                         last_record = page[-1]
+                        order_field, order_direction = order
+                        pagination_direction = GT if order_direction == 1 else LT
                         threshhold_field = last_record[order_field]
                         threshhold_lm = last_record['last_modified']
                         pagination = [
--- a/kinto/core/storage/testing.py
+++ b/kinto/core/storage/testing.py
@@ -1538,6 +1538,56 @@ class DeletedRecordsTest:
         self.assertEqual(total_records, 10)
         self.assertEqual(len(records), 4)
 
+    def test_get_all_parent_id_paginates_correctly(self):
+        """Verify that pagination doesn't squash or duplicate some records"""
+
+        for parent in range(10):
+            parent_id = 'abc{}'.format(parent)
+            self.storage.create(parent_id=parent_id, collection_id='c',
+                                record={'id': 'some_id', 'secret_data': parent_id})
+
+        real_records, _ = self.storage.get_all(parent_id='abc*', collection_id='c')
+        self.assertEqual(len(real_records), 10)
+
+        def sort_by_secret_data(l):
+            return sorted(l, key=lambda r: r['secret_data'])
+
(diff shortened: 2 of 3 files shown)


kinto/core/storage/testing.py
```diff
@@ -1525,6 +1538,56 @@ def test_get_all_handle_all_pagination_rules(self):
...
+        for order in [('secret_data', 1), ('secret_data', -1)]:
+            order_field, order_direction = order
+            sort = [Sort(*order), Sort('last_modified', -1)]
+            pagination_direction = GT if order_direction == 1 else LT
+            for limit in range(1, 10):
+                with self.subTest(order=order, limit=limit):
+                    records = []
+                    pagination = None
+                    while True:
+                        (page, total_records) = self.storage.get_all(
```
leplatrem: nit: superfluous parenthesis

kinto/core/storage/testing.py
```diff
@@ -1525,6 +1538,56 @@ def test_get_all_handle_all_pagination_rules(self):
...
+                        if not page:  # pragma: nocover
+                            break
+                        last_record = page[-1]
+                        threshhold_field = last_record[order_field]
+                        threshhold_lm = last_record['last_modified']
+                        pagination = [
+                            [Filter(order_field, threshhold_field, utils.COMPARISON.EQ),
+                             Filter('last_modified', threshhold_lm, utils.COMPARISON.LT)],
+                            [Filter(order_field, threshhold_field, pagination_direction)]
+                        ]
```
leplatrem: Maybe add a comment that what happens in reality:
https://github.com/Kinto/kinto/blob/master/kinto/core/resource/__init__.py#L1055-L1057

kinto/core/storage/testing.py
```diff
@@ -1525,6 +1538,56 @@ def test_get_all_handle_all_pagination_rules(self):
...
+
+        def sort_by_secret_data(l):
+            return sorted(l, key=lambda r: r['secret_data'])
+
+        GT = utils.COMPARISON.GT
+        LT = utils.COMPARISON.LT
+        for order in [('secret_data', 1), ('secret_data', -1)]:
+            order_field, order_direction = order
+            sort = [Sort(*order), Sort('last_modified', -1)]
+            pagination_direction = GT if order_direction == 1 else LT
```
leplatrem: I know it makes no sense to redo that at each loop iteration, but I would have had less trouble reading all this if `order_field` and `pagination_direction` would have been define right along the pagination rules below :)

kinto/core/storage/testing.py
```diff
@@ -1525,6 +1538,56 @@ def test_get_all_handle_all_pagination_rules(self):
         self.assertEqual(total_records, 10)
         self.assertEqual(len(records), 4)
 
+    def test_get_all_parent_id_paginates_correctly(self):
+        """Verify that pagination doesn't squash or duplicate some records"""
+
+        for parent in range(10):
```
leplatrem: Please add a comment like `Create records with different parent ids`

kinto/core/storage/memory.py
```diff
@@ -322,7 +322,7 @@ def extract_record_set(records, filters, sorting,
         paginated = {}
         for rule in pagination_rules:
             values = apply_filters(filtered, rule)
-            paginated.update(((x[id_field], x) for x in values))
+            paginated.update((id(x), x) for x in values)
```
leplatrem: index by memory address? I realize I don't understand why we don't just build a list :)

leplatrem:
> It isn't really clear what the scalability characteristics of Kinto are or should be. Is Kinto meant to store 10m records? I would say probably not. What about 1m records, but all in one collection? I'm not certain.

10m records accross several collections, I would say yes. 

1m records in one collection, it should support it, but maybe not in optimal conditions. Since synchronization is going to be painful anyway (especially on the first time).

> It isn't really clear (to me, at least) what queries Kinto should support as "first-class" and which are "gravy", i.e. added bonuses that we threw in because they were easy to implement but are not as well-loved, performant etc. as others. Naïvely I would say syncing collections is more important than supporting arbitrary filters, but then again, if that's what you want, CouchDB is available and may be a better fit for you.

I definitely agree that syncing is more important than arbitrary filters (for which we have the ElasticSearch plugin)

> It isn't really clear to me what the relative priorities are in the Kinto ecosystem. As an example, in order to produce a Total-Records header, we have to add a COUNT to our queries, which means enumerating all the matching records rather than just the ones that we're going to render in the response. In other distributed systems, you might just paginate until you get an empty page. Is the ease-of-use/sanity checking provided by the Total-Records header worth the performance cost?

That was a naive choice at the beginning of the project. The `Total-Records` is not even exposed in our clients API...

On the server side, the total count is just used to figure out whether we have a next page. Which, as you say, could be managed differently (ie. paginate until empty page).

glasserc:
One reason why `Total-Records` might be nice is to show pagination stuff on the client side. (e.g. if Total-Records is 50, we can show pages 1..5.) We do expose it, not as part of other requests, but as a `getTotalRecords` call in kinto-http.js. PostgREST handles this by exposing a `Content-Range` header, but that header doesn't contain a total count unless you add `Prefer: count=exact` header. They even recommend stripping this header in prod: http://postgrest.org/en/stable/admin.html?highlight=Content-Range#count-header-dos.

glasserc:
We spoke about this online today. Removing the `Total-Records` header is kind of an API-breaking change, even though to our knowledge, our clients don't actually rely on this header.

peterbe:
I'm not sure if it's worth it but we *could* skip the count part if the request is for the next page. E.g.
```
$ curl localhost:8888/v1/collection/records
< 200 OK
< Total-Records: 1,234,567
< Next-URL: http://localhost:8888/v1/paginate=somelonghashthing
$ curl http://localhost:8888/v1/paginate=somelonghashthing
< 200 OK
< Total-Records: 1,234,567
< Next-URL: http://localhost:8888/v1/paginate=anotherlonghash
```
In this case, the `Total-Records` number could come from the cache rather than having to do another query again. 

Granted, I don't really know how the pagination works in Kinto but if the number really is only there for helping pagination then storing the first count in the cache would definitely make it possible to skip it for the next couple of pages. 

Too complicated or every-little-helps?

kinto/core/storage/memory.py
glasserc: Well, I just tried building a list and none of the tests failed. Maybe the person who wrote it originally was afraid that different pagination rules would re-match a record twice? In that case, the tests probably won't fail because we don't construct pagination rules in such a way where this could happen (but a pathological user could conceptually find a way if they were motivated). Since none of the tests fail, I'll take it out..

Natim:
Let's prepare a release then :rose:

leplatrem:
@glasserc could you add a CHANGELOG entry please?

glasserc:
Oops, yes, this PR took so long I forgot about the unchecked boxes in the description. I've opened #1635 with a CHANGELOG entry.

PR #1646, opened 2018-05-23
Remove useless `permission`

glasserc:


leplatrem:
So this will work for anynomous users too?

glasserc:
It will work the same as it does today (i.e. no).

glasserc:
Sorry, let me rephrase... this line contributes nothing to whether this endpoint works for anonymous users or not. That's controlled elsewhere. At present, that setting is for the endpoint to allow anonymous access.

PR #1648, opened 2018-05-24
Add administrative view to delete user info

### LINKED ISSUE #442: List users / delete all records?
I understand there is no [user management](http://kinto.readthedocs.org/en/latest/api/1.x/authentication.html#a-word-about-users), but...
- Is it possible to list all users (user ids)?
- How to delete all records owned by a particular user?

Thanks!

glasserc:
Fixes #442.

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -33,8 +33,7 @@ class Deleted(resource.ResourceSchema):
 
 
 class UserDataFactory(RouteFactory):
-    def _find_required_permission(self, request, service):
-        return ("user_data", "delete")
+    method_permissions = {"delete": "delete"}
 
 
 class UserDataViewSet(viewset.ShareableViewSet):
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -48,7 +48,9 @@ def get_parent_uri(object_uri):
 
     """
     path = object_uri.rsplit("/", 2)
-    if len(path) < 2:
+    # len(path) == 1: no '/', probably a broken URL?
+    # len(path) == 2: one '/', doesn't conform to our URL scheme
+    if len(path) < 3:
         return ""
 
     return path[0]
@@ -105,8 +107,6 @@ class UserData(resource.ShareableResource):
         write_perm_principals = permission.get_objects_permissions(object_uris, ["write"])
         to_delete = set()
         for object_uri, principals in zip(object_uris, write_perm_principals):
-            if "write" not in principals:
-                continue
             principals = principals["write"]
             # "Ownership" isn't a real concept in Kinto, so instead we
             # define ownership as meaning "this user is the only one
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -176,6 +176,7 @@ class UserData(resource.ShareableResource):
                     *objects_by_parent_uri[(parent_uri, resource_name)]
                 )
 
+        # Remove this principal from existing users.
         permission.remove_principal(principal)
 
         return {"data": {"principal": principal}}
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -121,6 +121,10 @@ class UserData(resource.ShareableResource):
 
         # Group by (parent_uri, resource of child) to make fewer
         # requests to storage backend.
+        # Store the parsed object IDs, since those are what we
+        # actually give to the storage backend.
+        object_ids_by_parent_uri = collections.defaultdict(list)
+        # Store also the object URIs, which we give to the permission backend.
         objects_by_parent_uri = collections.defaultdict(list)
         # We have to get the matchdict of the child here anyhow, so
         # keep that to generate events later.
@@ -128,13 +132,14 @@ class UserData(resource.ShareableResource):
         for object_uri in to_delete:
             parent_uri = get_parent_uri(object_uri)
             resource_name, matchdict = core_utils.view_lookup(self.request, object_uri)
-            objects_by_parent_uri[(parent_uri, resource_name)].append(matchdict["id"])
+            objects_by_parent_uri[(parent_uri, resource_name)].append(object_uri)
+            object_ids_by_parent_uri[(parent_uri, resource_name)].append(matchdict["id"])
             # This overwrites previous matchdicts for the parent, but
             # we'll only use the fields that are relevant to the
             # parent, which will be the same for each child.
             matchdicts_by_parent_uri[parent_uri] = matchdict
 
-        for (parent_uri, resource_name), object_ids in objects_by_parent_uri.items():
+        for (parent_uri, resource_name), object_ids in object_ids_by_parent_uri.items():
             # Generate the parent matchdict from an arbitrary child's matchdict.
             matchdict = {**matchdicts_by_parent_uri[parent_uri]}
             matchdict.pop("id", None)
@@ -161,8 +166,15 @@ class UserData(resource.ShareableResource):
                     resource_name=resource_name,
                     resource_data=matchdict,
                 )
-                # FIXME: no way to purge just some tombstones for just this principal
-                permission.delete_object_permissions(*batch)
+                # FIXME: need to purge the above tombstones, but no
+                # way to purge just some tombstones for just this
+                # principal
+
+                # Clear permissions from the deleted objects, for
+                # example those of other users.
+                permission.delete_object_permissions(
+                    *objects_by_parent_uri[(parent_uri, resource_name)]
+                )
 
         permission.remove_principal(principal)
 
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -186,4 +186,9 @@ class UserData(resource.ShareableResource):
         # Remove this principal from existing users.
         permission.remove_principal(principal)
 
+        # Remove this principal from all groups that contain it.
+        associated_principals = permission.get_user_principals(principal)
+        for associated_principal in associated_principals:
+            permission.remove_user_principal(principal, associated_principal)
+
         return {"data": {"principal": principal}}
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -124,9 +124,7 @@ class UserData(resource.ShareableResource):
             for perm in permissions:
                 permission.remove_principal_from_ace(object_uri, perm, principal)
 
-        print("Going to delete", to_delete)
         to_delete = condense_under_parents(self.request, to_delete)
-        print("Condensed into", to_delete)
 
         # Group by (parent_uri, resource of child) to make fewer
         # requests to storage backend.
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -10,8 +10,7 @@ from kinto.authorization import RouteFactory
 from kinto.core import utils as core_utils, resource
 from kinto.core.events import notify_resource_event, ACTIONS
 from kinto.core.resource import viewset
-from kinto.core.storage import Filter, Sort
-from kinto.core.storage.utils import paginated
+from kinto.core.storage import Filter
 
 
 def slice_into_batches(iterable, batch_size):
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -158,10 +158,10 @@ class UserData(resource.ShareableResource):
                 filters = [Filter("id", batch, core_utils.COMPARISON.IN)]
                 timestamp = storage.resource_timestamp(resource_name, parent_uri)
                 records, _ = storage.get_all(
-                    collection_id=resource_name, parent_id=parent_uri, filters=filters
+                    resource_name=resource_name, parent_id=parent_uri, filters=filters
                 )
                 tombstones = storage.delete_all(
-                    collection_id=resource_name, parent_id=parent_uri, filters=filters
+                    resource_name=resource_name, parent_id=parent_uri, filters=filters
                 )
                 notify_resource_event(
                     self.request,
--- a/kinto/views/admin.py
+++ b/kinto/views/admin.py
@@ -66,6 +66,7 @@ def condense_under_parents(request, object_uris):
 
     """
     # Sort object_uris so we see ancestors before descendants.
+    object_uris = list(object_uris)
     object_uris.sort()
 
     ancestor_object_uris = set()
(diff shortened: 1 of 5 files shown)


kinto/views/admin.py
```diff
@@ -0,0 +1,134 @@
...
+class Deleted(resource.ResourceSchema):
+    principal = colander.SchemaNode(colander.String())
+
+    class Options:
+        preserve_unknown = False
+
+
+class UserDataFactory(RouteFactory):
+    def _find_required_permission(self, request, service):
+        return ('user_data', 'delete')
```
glasserc: I did this so I could control the permission name I wanted to show up in the config. (I wanted `user-data_delete` instead of `user-data_write`.) I just noticed this one says `user_data` instead of `user-data`, even though my config file says `user-data`, so I guess I don't 100% understand how this works. There might also be a better way to get this to behave the way I want to, rather than overriding the factory.

kinto/views/admin.py
```diff
@@ -0,0 +1,134 @@
...
+                        filters=filters
+                    )
+                    tombstones = storage.delete_all(
+                        collection_id=resource_name, parent_id=parent_id,
+                        filters=filters
+                    )
+                    # FIXME: where do I get this from?
+                    timestamp = None
+                    notify_resource_event(self.request, parent_id, timestamp,
+                                          tombstones, ACTIONS.DELETE, old=records)
```
glasserc: There's definitely some bugs here. In particular, the `quota` plugin fails hilariously to deal with what I'm doing here, and I'm not even sure the `buckets` / `collections` listeners are being fired (because they should have deleted the quota records).

kinto/views/admin.py
```diff
@@ -0,0 +1,134 @@
...
+
+    schema = Deleted
+
+    def delete(self):
+        principal = self.request.matchdict['principal']
+        storage = self.request.registry.storage
+        permission = self.request.registry.permission
+        object_uris_and_permissions = permission.get_accessible_objects([principal])
+        object_uris = list(object_uris_and_permissions.keys())
+        object_uris.sort()
```
glasserc: This `sort()` turns out not to be necessary, I misread how the `get_objects_permissions` function works.

kinto/views/admin.py
```diff
@@ -0,0 +1,143 @@
...
+        storage = self.request.registry.storage
+        permission = self.request.registry.permission
+        object_uris_and_permissions = permission.get_accessible_objects([principal])
+        object_uris = list(object_uris_and_permissions.keys())
+        object_uris.sort(reverse=True)
+        write_perm_principals = permission.get_objects_permissions(object_uris, ['write'])
+        to_delete = []
+        for object_uri, principals in zip(object_uris, write_perm_principals):
+            principals = principals['write']
+            if principals == set([principal]):
```
leplatrem: not contains because we only delete objects where the user is the only writer?

leplatrem:
What is the status of this @glasserc ? How far is it from being done?

kinto/views/admin.py
glasserc: Right.

glasserc:
It's not very close. I think it should be a little easier now that https://github.com/Kinto/kinto/pull/1671/ has landed, because we can just rely on the existing listeners to clean up after we delete a resource, but I haven't confirmed that. The use case I needed this for was already solved by https://github.com/Kinto/kinto-fxa/pull/55 so I felt much less urgency about it. Maybe we should close it until someone else has the energy to pick it up.

kinto/views/admin.py
```diff
@@ -0,0 +1,168 @@
...
+            ancestor_object_uris.add(object_uri)
+
+    return list(ancestor_object_uris)
+
+@resource.register(name='user-data',
+                   description='Delete the data owned by a user',
+                   plural_path='/__user_data__',
+                   object_path='/__user_data__/{{principal}}',
+                   viewset=UserDataViewSet(),
+                   collection_methods=('DELETE',))
```
leplatrem: `plural_methods` I guess

But `plural_delete(self)` isn't defined :)

kinto/views/admin.py
```diff
@@ -0,0 +1,168 @@
...
+class UserData(resource.ShareableResource):
+
+    schema = Deleted
+
+    def delete(self):
```
leplatrem: this is for a single object ( `object_methods=("DELETE",)` )

kinto/views/admin.py
```diff
@@ -0,0 +1,168 @@
...
+                    filters=filters
+                )
+                # FIXME: where do I get this from?
+                timestamp = None
+                notify_resource_event(self.request, parent_uri, timestamp,
+                                      tombstones, ACTIONS.DELETE, old=records,
+                                      resource_name=resource_name, resource_data=matchdict
+                )
+                # FIXME: no way to purge just some tombstones for just this principal
+                permission.delete_object_permissions(*batch)
```
leplatrem: I'm surprised this works. Should it be given object uris?

kinto/views/admin.py
```diff
@@ -0,0 +1,168 @@
...
+            # we'll only use the fields that are relevant to the
+            # parent, which will be the same for each child.
+            matchdicts_by_parent_uri[parent_uri] = matchdict
+
+        for (parent_uri, resource_name), object_ids in objects_by_parent_uri.items():
+            # Generate the parent matchdict from an arbitrary child's matchdict.
+            matchdict = {**matchdicts_by_parent_uri[parent_uri]}
+            matchdict.pop('id', None)
+
+            # FIXME: Deletes are paginated too.. ugh
```
leplatrem: Why is that?

I think what limits storage operations is this setting https://github.com/Kinto/kinto/blob/f1da55010ae7fe4a4be7d2c8378a35d82eecf84c/kinto/core/__init__.py#L92

kinto/views/admin.py
```diff
@@ -0,0 +1,168 @@
...
+                tombstones = storage.delete_all(
+                    collection_id=resource_name, parent_id=parent_uri,
+                    filters=filters
+                )
+                # FIXME: where do I get this from?
+                timestamp = None
+                notify_resource_event(self.request, parent_uri, timestamp,
+                                      tombstones, ACTIONS.DELETE, old=records,
+                                      resource_name=resource_name, resource_data=matchdict
+                )
```
leplatrem: I wouldn't mind skipping resource events if that simplifies the code

kinto/views/admin.py
```diff
@@ -0,0 +1,168 @@
...
+                filters = [Filter('id', batch, core_utils.COMPARISON.IN)]
+                records, _ = storage.get_all(
+                    collection_id=resource_name, parent_id=parent_uri,
+                    filters=filters
+                )
+                tombstones = storage.delete_all(
+                    collection_id=resource_name, parent_id=parent_uri,
+                    filters=filters
+                )
+                # FIXME: where do I get this from?
```
leplatrem: The only idea that comes into my mind is a call to `collection_timestamp()` before deleting,..

kinto/views/admin.py
```diff
@@ -0,0 +1,168 @@
...
+                timestamp = None
+                notify_resource_event(self.request, parent_uri, timestamp,
+                                      tombstones, ACTIONS.DELETE, old=records,
+                                      resource_name=resource_name, resource_data=matchdict
+                )
+                # FIXME: no way to purge just some tombstones for just this principal
+                permission.delete_object_permissions(*batch)
+
+        permission.remove_principal(principal)
```
leplatrem: Maybe we could add a comment here to make it clear that this principal is going to be removed from every permission set on every object

kinto/views/admin.py
glasserc: Good catch. I bet it doesn't work. Permissions are probably being deleted by cascading event handlers.

kinto/views/admin.py
glasserc: Getting resource events right is pretty important. For example, it's how the quota plugin handles adjusting quotas.

kinto/views/admin.py
glasserc: That's it, thanks. I was trying to figure out a way to get access to it rather than hard-coding a random number..

kinto/views/admin.py
glasserc: That isn't what this function does. Should it?

leplatrem: Yes it does :wink:

glasserc:
I think this is now a legitimate PR and should be reviewed accordingly.

I decided not to put the `user-data_delete_principals` setting in the kinto.tpl, figuring it wasn't a common enough endpoint to make it worth cluttering things up.

tests/test_views_admin.py
```diff
@@ -0,0 +1,242 @@
...
+from kinto.core import utils
+from kinto.views.admin import get_parent_uri
+
+from .support import BaseWebTest, MINIMALIST_BUCKET, MINIMALIST_COLLECTION, MINIMALIST_RECORD
+
+
+class DeleteUserDataTest(BaseWebTest, unittest.TestCase):
+    doomed_bucket_url = "/buckets/user_to_delete"
+    doomed_collection_url = doomed_bucket_url + "/collections/collection_to_delete"
+    doomed_bucket_collection_url = doomed_bucket_url + "/collections/collection"
```
leplatrem: I think it would help if you would briefly explain (here or in setup) what you mean by doomed and safe :)

tests/test_views_admin.py
```diff
@@ -0,0 +1,242 @@
...
+    safe_group_containing_doomed_user = safe_bucket_url + "/groups/silly-group"
+    shared_bucket_url = "/buckets/both"
+    doomed_user = get_user_headers("doomed", "trustno1")
+    hmac_secret = "secret_hmac_for_userids"
+    user_hash = utils.hmac_digest(hmac_secret, "mat:secret")
+    user_principal = f"basicauth:{user_hash}"
+    doomed_hash = utils.hmac_digest(hmac_secret, "doomed:trustno1")
+    doomed_user_principal = f"basicauth:{doomed_hash}"
+    delete_user_url = f"/__user_data__/{doomed_user_principal}"
+    subscribed = (events.ResourceChanged,)
```
leplatrem: Maybe this doesn't have to be a class attribute IDK

tests/test_views_admin.py
```diff
@@ -0,0 +1,242 @@
...
+        self.app.delete(self.delete_user_url, headers=self.headers)
+
+    def test_doomed_bucket_was_deleted(self):
+        with self.assertRaises(exceptions.ObjectNotFoundError):
+            self.storage.get("bucket", "", "user_to_delete")
+
+    def test_doomed_bucket_collection_was_deleted(self):
+        # Everything under this bucket should be gone
+        with self.assertRaises(exceptions.ObjectNotFoundError):
+            self.storage.get("collection", self.doomed_bucket_url, "collection")
```
leplatrem: I realized (when refactoring the core notions stuff), that using kwargs explicitly helps a lot. (`resource_name=...`)

tests/test_views_admin.py
```diff
@@ -0,0 +1,242 @@
...
+        self.storage.get("record", self.safe_collection_url, "record_safe")
+
+    def test_safe_collection_doomed_record_was_deleted(self):
+        with self.assertRaises(exceptions.ObjectNotFoundError):
+            self.storage.get("record", self.safe_collection_url, "record_to_delete")
+
+    def test_safe_group_doomed_user_was_removed(self):
+        principals = self.permission.get_user_principals(self.doomed_user_principal)
+        for principal in principals:
+            if principal.startswith("system:"):
```
leplatrem: `system.` ?

tests/test_views_admin.py
```diff
@@ -0,0 +1,242 @@
...
+        self.assertEqual(permissions, {})
+
+    def test_doomed_bucket_collection_no_longer_has_any_perms(self):
+        permissions = self.permission.get_object_permissions(self.doomed_bucket_collection_url)
+        self.assertEqual(permissions, {})
+
+    def test_safe_bucket_no_longer_has_doomed_user_perm(self):
+        permissions = self.permission.get_object_permissions(self.safe_bucket_url)
+        for perm, principals in permissions.items():
+            self.assertNotIn(self.doomed_user_principal, principals)
```
leplatrem: Assert that we still have `self.user_principal` ?

(permissions should not be empty)

kinto/views/admin.py
```diff
@@ -0,0 +1,193 @@
...
+    the URI, identify the parent resource, and generate a new URI.
+
+    """
+    path = object_uri.rsplit("/", 2)
+    # len(path) == 1: no '/', probably a broken URL?
+    # len(path) == 2: one '/', doesn't conform to our URL scheme
+    if len(path) < 3:
+        return ""
+
+    return path[0]
```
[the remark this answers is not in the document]
glasserc: To be honest, I'm not sure. It feels like there's a concept of "hierarchy" in Kinto -- what resources are contained by what other resources and at what levels -- but nowhere does there seem to be any kind of formalization of that hierarchy. I feel like the whole concept needs to be structured somewhere, but I think it's outside the scope of this PR. This is a quick-and-dirty hack that responds to what this endpoint needs.

tests/test_views_admin.py
```diff
@@ -0,0 +1,242 @@
...
+
+class GetParentUriTest(unittest.TestCase):
+    def test_parent_uri_behaves_sensibly_for_unknown_resources(self):
+        unknown_url = "/cities/prague/monuments/castle/foods/mushrooms"
+        self.assertEqual(get_parent_uri(unknown_url), "/cities/prague/monuments/castle")
+
+    def test_parent_uri_accepts_pathological_urls(self):
+        # This shouldn't ever actually happen, and if it does happen
+        # it's probably a sign of a larger incompatibility of this
+        # method with the URL scheme
```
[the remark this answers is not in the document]
glasserc: If we try to support `/buckets`, then we should also try to support `/buckets/abcd/collections`. The method (as scoped right now) doesn't have to care too much about these kinds of URLs because it only considers actual object IDs.

tests/test_views_admin.py
glasserc: I was basing this on the code in `test_events`, although I admit it's a little excessive for this case. There might be a larger solution to cleaning up how we initialize apps in the test suites generally..

PR #1651, opened 2018-05-25
Upgrade kinto-admin to 1.18.0

leplatrem:


leplatrem:
Yes, so that our community users can deploy it during the weekend :)

PR #1653, opened 2018-05-26
Cache backend prompt in init

aimanparvaiz:
Fixes #

- [ ] Add documentation.
- [x] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -174,7 +174,7 @@ def main(args=None):
                                        'install', 'kinto[redis]'])
         elif cache_backend == 'memcached':
             try:
-                import memcache
+                import memcache     # NOQA
             except ImportError:
                 subprocess.check_call([sys.executable, '-m', 'pip',
                                        'install', 'python-memcached'])
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -174,7 +174,7 @@ def main(args=None):
                                        'install', 'kinto[redis]'])
         elif cache_backend == 'memcached':
             try:
-                import memcache     # NOQA
+                import memcache  # NOQA
             except ImportError:
                 subprocess.check_call([sys.executable, '-m', 'pip',
                                        'install', 'kinto[memcached]'])
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -57,6 +57,11 @@ def main(args=None):
                                    dest='backend',
                                    required=False,
                                    default=None)
+            subparser.add_argument('--cache-backend',
+                                   help='{memory,redis,postgresql,memcached}',
+                                   dest='cache-backend',
+                                   required=False,
+                                   default=None)
             subparser.add_argument('--host',
                                    help='Host to listen() on.',
                                    dest='host',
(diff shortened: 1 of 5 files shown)


kinto/__main__.py
```diff
@@ -138,21 +144,40 @@ def main(args=None):
...
-        elif backend == 'redis':
+        elif backend == 'redis' or cache_backend == 'redis':
             try:
                 import kinto_redis  # NOQA
             except ImportError:
                 subprocess.check_call([sys.executable, '-m', 'pip',
                                        'install', 'kinto[redis]'])
+        elif cache_backend == 'memcached':
+            try:
+                import memcache     # NOQA
```
Natim: Only two spaces before the #

kinto/config/__init__.py
```diff
@@ -29,7 +29,19 @@ def render_template(template, destination, **kwargs):
             output.write(rendered)
 
 
-def init(config_file, backend, host='127.0.0.1'):
+def get_cache_backend_url(cache_be_value):
+    if cache_be_value == 'kinto.core.cache.postgresql':
+        url = 'postgres://postgres:postgres@localhost/postgres'
+    elif cache_be_value == 'kinto.core.cache.redis':
+        url = 'redis://localhost:6379' + '/2'
```
Natim: ``'redis://localhost:6379/2'``

tests/test_config.py
```diff
@@ -55,13 +55,13 @@ def test_create_destination_directory(self):
 
     @mock.patch('kinto.config.render_template')
     def test_hmac_secret_is_text(self, mocked_render_template):
-        config.init('kinto.ini', 'postgresql')
+        config.init('kinto.ini', 'postgresql', 'postgresql')
```
Natim: We should use kwargs here to make it more readable:

```config.init('kinto.ini', backend='postgresql', cache_backend='postgresql')```

Should we also default cache_backend to backend if the parameter is None?

File: tests/test_config.py [hunk 4]
```diff
@@ -97,15 +96,15 @@ def test_init_redis_values(self, mocked_render_template):
...
+            'cache_url': redis_url + '/2',
             'permission_url': redis_url + '/3',
             'kinto_version': __version__,
             'config_file_timestamp': strftime('%a, %d %b %Y %H:%M:%S %z')
         })
 
     @mock.patch('kinto.config.render_template')
     def test_init_memory_values(self, mocked_render_template):
-        config.init('kinto.ini', 'memory')
+        config.init('kinto.ini', 'memory', 'memory')
```
Natim: We should add a test with postgresql and memcache.

tests/test_config.py
aimanparvaiz: quick question, in my understanding cache_backend cant really be none (it can be default, which would be memory) since we are asking for user input in a while loop in __main__.py. Please let me know if I am missing something here. 
Thanks

Natim: There are two things, the CLI tool that use the config.init function and the config.init function that can be use from somewhere else and notably in the tests.

For the former, you are right, but one could use the config.init function and I find it better not to have to pass twice the same value.

leplatrem:
@Natim final r?

Issue #1555, opened 2018-03-27
Schema validation at bucket/collection level too


leplatrem:
Currently if there's a `schema` attribute in the collection metadata, we validate the posted records.
See
https://docs.kinto-storage.org/en/stable/api/1.x/collections.html#collection-json-schema

I propose that when we set `collection:schema`, `group:schema` or `record:schema` in the bucket metadata, we validate the collection, group or record respectively :)

Use case: the «per bucket» configuration of Firefox remote settings where every record in every collection could have a `target` field that is validated against the same schema

saurabhShrivastava1995:
Working on this

leplatrem:
@ssWitcher I'm sorry I hadn't seen your comment :( I needed this for something we have in production and jumped on it!
I hoped you hadn't spent too much energy on it :(

PR #1654, opened 2018-05-28
Add support JSON schema at bucket level (#1555)

leplatrem:
Ref #1555 

- [x] Basic feature
- [x] Add collection and group validation
- [x] Add validation of `record:schema`, `group:schema` and `collection:schema` properties
- [x] Add docs

### DIFF
--- a/kinto/schema_validation.py
+++ b/kinto/schema_validation.py
@@ -43,6 +43,8 @@ def validate_schema(data, schema, ignore_fields=[]):
     else:
         schema = {f: v for f, v in schema.items() if f != 'required'}
 
+    data = {f: v for f, v in data.items() if f not in ignore_fields}
+
     try:
         validate(data, schema)
     except ValidationError as e:
@@ -56,7 +58,7 @@ def validate_schema(data, schema, ignore_fields=[]):
         raise e
 
 
-def validate_from_parent_schema_or_400(data, resource_name, request, ignore_fields=[]):
+def validate_from_bucket_schema_or_400(data, resource_name, request, ignore_fields=[]):
     """Lookup in the parent objects if a schema was defined for this resource.
 
     If the schema validation feature is enabled, if a schema is/are defined, and if the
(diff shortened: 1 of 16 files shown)


leplatrem:
I generalized to collections and groups.

Maybe we improve the current code, which repeats a lot of lines between record / collection / bucket.

Also I still have to update the documentation.

tests/test_views_schema_collection.py
```diff
@@ -0,0 +1,81 @@
...
+                'type': 'string'
+            },
+        },
+    },
+}
+
+VALID_COLLECTION = {'uiSchema': {'schema': {'ui:widget': 'hidden'}}, 'displayFields': ['name']}
+
+
+class DeactivatedSchemaTest(BaseWebTest, unittest.TestCase):
```
glasserc: It's kind of unfortunate that the schema validation feature doesn't have to be turned on here, because it claims `collection:schema` as a reserved word..

leplatrem: What do you mean?

That it's a pity to reserve the field `collection:schema` when the feature is disabled?

glasserc: Yes, that.

leplatrem:
@glasserc I completely changed the organization of the code. There's less repetition, IMO it's better but I can change my mind.

kinto/schema_validation.py
```diff
@@ -0,0 +1,112 @@
...
+        return
+
+    # For records, there can be several schemas (collection level + bucket level).
+    schemas = []
+
+    bucket_id = request.matchdict["bucket_id"]
+    bucket_uri = utils.instance_uri(request, 'bucket', id=bucket_id)
+
+    # For records, the schema defined on the collection will be validated first.
+    if resource_name == "record":
```
glasserc: I'm a little nervous about this. It hard-codes part, but only part, of the containment hierarchy of Kinto. To be honest, I preferred that this logic be part of the Record resource.

PR #1655, opened 2018-05-28
Limit network requests to current domain in Kinto Admin using CSP

leplatrem:
See ulfr recommandation https://github.com/Kinto/kinto-admin/issues/460#issuecomment-375127642

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ This document describes changes between each past release.
 **New features**
 
 - Kinto Admin plugin now supports OpenID Connect
+- Limit network requests to current domain in Kinto Admin using `Content-Security Policies <https://hacks.mozilla.org/2016/02/implementing-content-security-policy/>`_
 
 **Internal changes**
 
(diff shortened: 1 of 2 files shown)


kinto/plugins/admin/views.py
```diff
@@ -11,16 +11,10 @@
 # Configured home page
 @cache_forever
 def admin_home_view(request):
-    settings = {
-        'authMethods': aslist(request.registry.settings.get('multiauth.policies'))
-    }
-    globalSettings = '<script>window.globalSettings = {};</script>'.format(json.dumps(settings))
```
Natim: If we don't need to generate the settings from the installed kinto plugins, we don't need to have a python view anymore to render the html page. (except maybe to route between the index.html page and the help page but do we really?)

kinto/plugins/admin/src/index.js
```diff
@@ -5,8 +5,7 @@ import * as signoffPlugin from "kinto-admin/lib/plugins/signoff";
 
 const settings = {
   maxPerPage: 50,
-  singleServer: document.location.toString().split('/admin/')[0],
-  ...window.globalSettings,
```
Natim: Why is this related to the CSP feature?

leplatrem: I opened #1658 :)

PR #1658, opened 2018-05-28
Clean-up unused globalSettings in admin plugin

leplatrem:


Natim:
Is this because the new kinto-admin is able to read the capabilities to detect the activated authentication mecanisms?

leplatrem:
> Is this because the new kinto-admin is able to read the capabilities to detect the activated authentication mecanisms?

Yes!

Natim:
Will it automatically select the current server with this changes?

leplatrem:
> Will it automatically select the current server with this changes?

Yes, it's done in `index.js`

PR #1666, opened 2018-05-31
Extract instance_uri_registry too

glasserc:
This is slightly less theoretically correct in the
instance_uri_registry case:

- request.route_path handles certain "special" params such as
  _script_url, _scheme, _host, _port, _query, and _anchor. However, we
  never use these in calls to instance_uri.

- request.route_path calls route.pregenerator, which allows Pyramid
  plugins to do fancy things like customize the routes generated
  according to the request. This is apparently used by e.g. Pylons
  subdomain support. However, we don't appear to use this either.

- (n/a) Add documentation.
- (n/a) Add tests.
- (n/a) Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,10 @@ This document describes changes between each past release.
 
 - Kinto Admin plugin now supports OpenID Connect
 - Limit network requests to current domain in Kinto Admin using `Content-Security Policies <https://hacks.mozilla.org/2016/02/implementing-content-security-policy/>`_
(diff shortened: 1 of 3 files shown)


glasserc:
Thinking about this further, I am wondering if it makes more sense to just have the caller use `Request.blank` to gin up a fake request. I could also define `instance_uri_registry` by using `Request.blank`, but writing tests for such a silly function almost seems not worth it. What do you think @leplatrem ?

glasserc:
I think we should close this in favor of #1667.

PR #1670, opened 2018-06-01
Add package.lock of kinto-admin

leplatrem:


leplatrem:
@peterbe r?

peterbe:
Not sure how to review this. I can't really look at the file. 

Is it documented that developers have to remember update this if they upgrade anything in `package.json`? Something [similar to this](https://buildhub2.readthedocs.io/en/latest/ui.html#upgrading-adding-dependencies)?

leplatrem:
> Is it documented that developers have to remember update this if they upgrade anything in package.json?

Yeah I think we should do that.

leplatrem:
I'll write it while upgrading to 1.19

Issue #945, opened 2016-11-25
Event payload attributes on batch requests


leplatrem:
> This is a follow-up bug of #942

On batch requests, an event is sent for each (parent id / resource / action), and the list of objects is provided in `event.impacted_records`. But the content of `event.payload` makes very little sense in this context, and can lead to serious confusions:

```
>>> event.payload

{ 'action': 'update',
  'bucket_id': u'staging',
  'collection_id': u'gfx',
  'record_id': u'fd6a9531-af10-79a0-e645-4249218dbf4f',
  'resource_name': 'record',
  'timestamp': 1479999631700L,
  'uri': u'/buckets/staging/collections/gfx/records/fd6a9531-af10-79a0-e645-4249218dbf4f',
  'user_id': 'basicauth:cbd3731f18c97ebe1d31d9846b5f1b95cf8eeeae586e201277263434041e99d1'}
```

I propose that when there are more than one impacted object then we drop the `${resource_name}_id` and the `uri` keys.

Thoughts?

glasserc:
Would this mean that some events will have `${resource_name}_id` and `uri`, and others won't, depending on what request triggered them? That sounds bad to me -- I would rather drop them from all events, to have consistent events.

leplatrem:
We could remove `${resource_name}_id` and leave `uri` with the value of the plural endpoint for example.

gabisurita:
From what I've worked with this plugin, I couldn't understand why we need the `${resource_name}_id` fields. Couldn't we just use only `uri` all the times?

Natim:
For instance for attachment you need it to know on which record you want to assign the attachment to.

We can probably have helper that let us get this information from the URI but I like the idea of having them already there in the payload.

If event's are different ones I don't mind having a different payload because a plugin most of the time subscribe to one event at the time, so we won't need different event handling. Also if record_id is not present it probably means that you don't need it right?

glasserc:
Well, but are the events different ones? It seems like this proposal is to change `update` events to have certain fields if it was a batch request, and to lack those fields if it wasn't a batch request. Wouldn't a plugin want to subscribe to both of those cases, and handle them similarly?

leplatrem:
I think there is some confusion here. The problem concerns the events that have several impacted records.
[...]

PR #1671, opened 2018-06-01
Allow events to be raised on other resources

glasserc:
I was going to update the tests, but there aren't any for this module?

- (n/a) Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,12 +7,22 @@ This document describes changes between each past release.
 9.3.0 (unreleased)
 ------------------
 
+**Breaking changes**
+
+- ``kinto.core.events.get_resource_events`` now returns a generator
+  rather than a list.
+
+
 **New features**
 
 - Include Python 3.7 support.
-- kinto.core.events now supports ``resource_name`` and
-  ``resource_data``. These are useful when emitting events from one
-  view "as though" they came from another view.
+- ``kinto.core.events.notify_resource_event`` now supports
+  ``resource_name`` and ``resource_data``. These are useful when
+  emitting events from one view "as though" they came from another
+  view.
+- Resource events can now trigger other resource events, which are
+  handled correctly. This might be handy if one resource wants to
+  simulate events on another "virtual" resource, as in ``kinto-changes``.
 
 9.2.3 (2018-07-05)
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ This document describes changes between each past release.
 **New features**
 
 - Include Python 3.7 support.
+- kinto.core.events now supports ``resource_name`` and
+  ``resource_data``. These are useful when emitting events from one
+  view "as though" they came from another view.
 
 9.2.3 (2018-07-05)
 ------------------
--- a/kinto/core/events.py
+++ b/kinto/core/events.py
@@ -74,7 +74,7 @@ def notify_resource_events_before(handler, registry):
     """
     def tween(request):
         response = handler(request)
-        for event in request.get_resource_events():
+        for event in request.drain_resource_events():
             request.registry.notify(event)
 
         return response
@@ -95,7 +95,7 @@ def setup_transaction_hook(config):
         if not success:  # pragma: no cover
             return
 
-        for event in request.get_resource_events(after_commit=True):
+        for event in request.drain_after_resource_events():
             try:
                 request.registry.notify(event)
             except Exception:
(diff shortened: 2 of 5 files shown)


glasserc:
There's also a reference to request.path in this method that should probably get updated.

glasserc:
Actually, because of #945, `uri` is essentially always wrong, so I'll just leave it alone for now.

glasserc:
I discovered while working on Kinto/kinto-changes#41 that having an event trigger another event doesn't always cause the second event to be fired. I added a commit to handle this by reorganizing the `kinto.core.events` module, but I'm not really sure that it works and I haven't added tests yet. f? @leplatrem

kinto/core/events.py
```diff
@@ -116,31 +116,50 @@ def on_new_request(event):
...
         else:
-            if action == ACTIONS.READ:
-                event_cls = ResourceRead
-            else:
-                event_cls = ResourceChanged
-        event = event_cls(payload, impacted, request)
-        events.append(event)
-    return events
+            after_impacted = afterwards[key][2]
+            after_impacted.extend(impacted)
```
glasserc: Basically, I changed `resource_events` to be mutated (drained), but this poses a problem because we would no longer be able to fire `AfterResourceChanged` and `AfterResourceRead`, so now I keep a second explicit dictionary of those after they've been drained from the first one.

glasserc:
N.B. This means it's possible to write an event listener that sends another very similar event, which can break the event "grouping". I'm not sure if this is something we should support, reject, or merely tolerate.

kinto/core/events.py
```diff
@@ -116,31 +116,50 @@ def on_new_request(event):
...
+            event_cls = AfterResourceRead
+        else:
+            event_cls = AfterResourceChanged
+
+        yield event_cls(payload, impacted, request)
 
 
 def notify_resource_event(request, parent_id, timestamp, data, action,
-                          old=None):
+                          old=None, resource_name=None, resource_data=None):
```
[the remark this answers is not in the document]
glasserc: `matchdict` sounds an awful lot like a dictionary of information describing the URL -- what route it matched, what resource was named, what its ID was, etc. However, by definition, if you are providing this information using this argument, then the data is about some resource *different* from the one specified by the current URL. This is why I gravitated away from `matchdict`.

glasserc:
I rebased onto current master and pushed a couple new commits. In particular, I decided that it made sense to extract an `EventCollector` class which seemed like a reasonable place to keep the logic for how to merge events. This meant unwinding some of the changes I made to move `get_resource_events` into `drain_resource_events` -- I kept the existing name/calling convention but it's now a generator rather than returning the events. That's probably OK for the use demonstrated in `kinto-signer`. I also added a test for an event triggering another event. I am thinking about adding a test to demonstrate that "cascading" events are merged the same way as regular events.

CHANGELOG.rst
```diff
@@ -10,6 +10,9 @@ This document describes changes between each past release.
 **New features**
 
 - Include Python 3.7 support.
+- kinto.core.events now supports ``resource_name`` and
```
leplatrem: `kinto.core.events.notify_resource_event()` now supports...

leplatrem: Should it be added to http://kinto.readthedocs.io/en/stable/core/index.html ?

tests/core/resource/test_events.py
```diff
@@ -424,6 +425,52 @@ def test_events_are_not_sent_if_subrequest_fails(self):
...
+        self.assertEqual(len(resource_changed_events), 2)
+        self.assertEqual(resource_changed_events[0].payload['action'], ACTIONS.CREATE.value)
+        self.assertEqual(len(resource_changed_events[0].impacted_records), 1)
+        self.assertEqual(resource_changed_events[1].payload['action'], ACTIONS.CREATE.value)
+        self.assertEqual(len(resource_changed_events[1].impacted_records), 1)
+        self.assertEqual(resource_changed_events[1].impacted_records[0]['new']['name'],
+                         'de New York')
+
+    def test_cascading_events_are_merged(self):
+        print("Posted to", self.collection_url)
```
leplatrem: leftover?

tests/core/resource/test_events.py
```diff
@@ -424,6 +425,52 @@ def test_events_are_not_sent_if_subrequest_fails(self):
...
+            notify_resource_event(event.request, parent_id, event.payload['timestamp'], new_record,
+                                  ACTIONS.CREATE)
+
+    def test_event_can_trigger_other_event(self):
+        self.app.post_json(self.collection_url, self.body,
+                           headers=self.headers, status=201)
+        resource_changed_events = [e for e in self.events if isinstance(e, ResourceChanged)]
+        self.assertEqual(len(resource_changed_events), 2)
+        self.assertEqual(resource_changed_events[0].payload['action'], ACTIONS.CREATE.value)
+        self.assertEqual(len(resource_changed_events[0].impacted_records), 1)
```
leplatrem: what's `resource_changed_events[0].impacted_records[0]['new']['name']` here?

CHANGELOG.rst
glasserc: I added it using ``autofunction`` to ``notifications.rst``. On the one hand, you could say I was trying to integrate my documentation with what was already there. On the other hand, you could say I was being lazy...

tests/core/resource/test_events.py
glasserc: It should match the order of the events, so `de Paris` (which triggers the `de New York`).

PR #1688, opened 2018-06-13
Add black

Natim:
Following DjangoCon it seems that using black is the right thing to do. Same as using prettier for our Javascript code, using black make the codebase coherent.

Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.

Blackened code looks the same regardless of the project you're reading. Formatting becomes transparent after a while and you can focus on the content instead.

Black makes code review faster by producing the smallest diffs possible.

- [ ] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,11 +19,12 @@ import sys
 # have an absolute __file__
 __HERE__ = os.path.dirname(os.path.abspath(__file__))
 
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
+on_rtd = os.environ.get("READTHEDOCS", None) == "True"
 
 if not on_rtd:  # only import and set the theme if we're building docs locally
     import sphinx_rtd_theme
-    html_theme = 'sphinx_rtd_theme'
+
+    html_theme = "sphinx_rtd_theme"
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
 # otherwise, readthedocs.io uses their theme by default, so no need to specify
@@ -32,7 +33,7 @@ if not on_rtd:  # only import and set the theme if we're building docs locally
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('..'))
+sys.path.insert(0, os.path.abspath(".."))
 
 # -- General configuration ------------------------------------------------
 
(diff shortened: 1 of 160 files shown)


docs/conf.py
```diff
@@ -43,31 +44,29 @@
...
 
 # The master toctree document.
-master_doc = 'index'
+master_doc = "index"
 
 # General information about the project.
-project = 'Kinto'
-copyright = '2015-%s — Mozilla Services' % datetime.datetime.now().year
+project = "Kinto"
+copyright = "2015-%s — Mozilla Services" % datetime.datetime.now().year
```
Natim: Move to ``.format()``

kinto/__init__.py
```diff
@@ -44,44 +43,44 @@ def main(global_config, config=None, **settings):
...
     # Force project name, since it determines settings prefix.
-    config.add_settings({'kinto.project_name': 'kinto'})
+    config.add_settings({"kinto.project_name": "kinto"})
 
-    kinto.core.initialize(config,
-                          version=__version__,
-                          default_settings=DEFAULT_SETTINGS)
+    kinto.core.initialize(
+        config, version=__version__, default_settings=DEFAULT_SETTINGS
+    )
```
leplatrem: That kind of change is a no go for me :) :trollface:

kinto/__init__.py
```diff
@@ -44,44 +43,44 @@ def main(global_config, config=None, **settings):
...
     if schema_enabled:
         config.add_api_capability(
-            'schema',
-            description='Validates collection records with JSON schemas.',
-            url='https://kinto.readthedocs.io/en/latest/api/1.x/'
-                'collections.html#collection-json-schema')
+            "schema",
+            description="Validates collection records with JSON schemas.",
+            url="https://kinto.readthedocs.io/en/latest/api/1.x/"
+            "collections.html#collection-json-schema",
```
leplatrem: «That must be a bug, that can't possibly be intentional» says my neighboor

Natim: :sob:

leplatrem:
Can we vote? Am I the only one to be hesitant?

glasserc:
I'm hesitant too. I think what we have already (verification using flake8) is working pretty well. I think we jumped on prettier too early in the JS world with the consequence that periodically we'd get a PR to update prettier and have to reformat everything. I'd like to avoid this mistake here.

I think if we can articulate exactly what we don't like about the Black style and reflect it in issues upstream, that would put us in a much better position.

PR #1689, opened 2018-06-14
handle json pointer errors, fixes #1685

### LINKED ISSUE #1685: Unhandled JSON pointer errors
When setting a schema on resources, unresolvable references are not checked upon creation.

When creating a resource that is checked against that schema, a `RefResolutionError` is raised by `jsonschema` and not caught.

For example, creating the schema:

```json
{"data": {
    "schema": {
      "properties": {
        "address": { "$ref": "#/definitions/address" }
      }
    }
  }
}
```

POST data:

```json
{ "data": {
   "address": "seattle"
  }
}
```

This will crash the server with a 500 error and the following message:

```shell
"Unresolvable JSON pointer: %r" % fragment
jsonschema.exceptions.RefResolutionError: Unresolvable JSON pointer: 'definitions/address' 
```

I can work on this, I was just wondering what you all felt the approach should be.

It can either check when creating the schema or upon each invocation of it (i.e., POST data). The only slight trouble I can see with checking and not allowing unresolvable schema is that the schema can be a url, so if it is not resolving at the time of creation, it will fail.

That seems fair to me; does that seem like a reasonable fix?

stloma:
Fixes #1685 

This fixes unhandled errors when evaluating data against schema with pointers.

I originally proposed that a possible way to handle this was to, upon initial creation of the schema, verify whether it fully resolves or not and reject it if it does not.

Unfortunately, there was no good way that I found to do this. According to `jsonschema`, this lazy lookup for `$ref` is intended behavior for now (though there is discussion to include this check as default behavior in v2).

This leaves the option to either write an independent implementation in Kinto, or add a library on top of `jsonschema` to do this lookup when creating schema.

My initial thought is I'm not a big fan of either of those, particularly when there is discussion of adding the functionality to `jsonschema` in the future.

My solution was to follow the `jsonschema` thought process and lazy lookup `$ref` references. So, accept any schema and provide a meaningful error if a reference fails when a JSON document is evaluated against it.

If there is a desire to add functionality to fail upon creation of the schema, I can give more options as well, but I think this is a good solution everything considered.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,10 @@ This document describes changes between each past release.
 9.3.0 (unreleased)
 ------------------
 
+** Bug fixes**
+
+- Fixed bug where unresolved JSON pointers would crash server (fixes #1685)
+
 **Internal changes**
 
 - Update the Dockerfile with the new kinto --cache-backend option. (#1686)
--- a/kinto/schema_validation.py
+++ b/kinto/schema_validation.py
@@ -1,5 +1,5 @@
 import colander
-from jsonschema import Draft4Validator, ValidationError, SchemaError, validate
+from jsonschema import Draft4Validator, ValidationError, SchemaError, RefResolutionError, validate
 from pyramid.settings import asbool
 
 from kinto.core import utils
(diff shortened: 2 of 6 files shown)


glasserc:
Thanks!

PR #1735, opened 2018-08-14
Raise error if accounts is misconfigured (fixes #1734)

### LINKED ISSUE #1734: Kinto onboarding experience (part 2)
These is a followup from #1733 with random feedback with onboarding when trying to use the [accounts plugin](http://docs.kinto-storage.org/en/stable/api/1.x/accounts.html).

Started with updating my `config/kinto.ini` with:

```
kinto.includes = kinto.plugins.default_bucket
                 kinto.plugins.admin
                 kinto.plugins.accounts
```

Restarting the server goes smoothly. The admin loads fine and renders a new entry for *Kinto Account Auth*. I never created any Account just yet, though out of curiosity I try to log in using silly:silly:

Tadaa:

Wait, what?

Oh. It seems it actually used the Basic Auth strategy instead of the account one for login. This is odd and confusing as fsck.

Actually, I didn't went further with toying around with the admin as it looks broken to me. This is a little sad.

leplatrem:
Fixes #1734

### DIFF
--- a/kinto/plugins/accounts/__init__.py
+++ b/kinto/plugins/accounts/__init__.py
@@ -27,7 +27,7 @@ def includeme(config):
 
     settings = config.get_settings()
 
-    # Check that the account policy is mentionned in docs if included.
+    # Check that the account policy is mentioned in config if included.
     accountClass = 'AccountsAuthenticationPolicy'
     policy = None
     for k, v in settings.items():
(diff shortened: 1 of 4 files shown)


kinto/plugins/accounts/__init__.py
```diff
@@ -21,13 +25,28 @@ def includeme(config):
         'read': {'account': ['write', 'read']}
     }
 
-    # Add some safety to avoid weird behaviour with basicauth default policy.
     settings = config.get_settings()
+
+    # Check that the account policy is mentionned in docs if included.
```
glasserc: `mentioned`. Also, shouldn't `in docs` be `in config`?

Issue #1733, opened 2018-08-14
Kinto onboarding experience (part 1)


n1k0:
These is random feedback with onboarding when setting up Kinto using docker-compose following the official docs at http://docs.kinto-storage.org/en/stable/tutorials/install.html#using-docker-compose.

You may want to spawn several issues out of all this, your call.

```
$ docker --version
Docker version 18.06.0-ce, build 0ffa825
$ docker-compose --version
docker-compose version 1.18.0, build 8dd22a9
$ wget https://raw.githubusercontent.com/Kinto/kinto/master/docker-compose.yml
$ docker-compose up
```

> Note I don't use sudo as it's not mandatory using docker when properly configured.

To retrieve the `kinto.ini` configuration file, I'm doing:

```
$ mkdir config
$ docker-compose exec web cat /etc/kinto/kinto.ini > config/kinto.ini
```

I enabled the admin by updating the related section:

```
kinto.includes = kinto.plugins.default_bucket
                 kinto.plugins.admin
```

And updated the `docker-compose.yml` file to mount config to a local volume:

```
db:
  image: postgres
  environment:
    POSTGRES_USER: postgres
    POSTGRES_PASSWORD: postgres
cache:
  image: library/memcached
web:
  image: kinto/kinto-server
  links:
  - db
  - cache
  ports:
  - "8888:8888"
  volumes:
  - ./config:/etc/kinto
  environment:
    KINTO_CACHE_BACKEND: kinto.core.cache.memcached
    KINTO_CACHE_HOSTS: cache:11211 cache:11212
    KINTO_STORAGE_BACKEND: kinto.core.storage.postgresql
    KINTO_STORAGE_URL: postgres://postgres:postgres@db/postgres
    KINTO_PERMISSION_BACKEND: kinto.core.permission.postgresql
    KINTO_PERMISSION_URL: postgres://postgres:postgres@db/postgres
```

Then restarted the server.

The root endpoint loads fine at `http://127.0.0.1:8888`, thoug I noticed errors in the docker console:

```
web_1    | "GET   /?" 307 (3 ms)  agent=Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36 errno=0 lang=fr-FR,fr;q=0.9,en-US;q=0.8,en;q=0.7 time=2018-08-14T12:10:26.610000
[...]

Issue #1494, opened 2018-02-13
Revamp authentication documentation


leplatrem:
- Remove mentions like `there is no account`
- Remove basic auth as first class
- Move basic auth token sorcery out of API docs
- Introduce OpenID Connect
- Remove mentions of FXA (not opened to public anyway)
- ...

As discussed in #1425

leplatrem:
Fixed in #1736

PR #1736, opened 2018-08-14
Get rid of basicauth

leplatrem:
Alright, let's do this.

Ref #1733 #1494 

The `basicauth` authentication policy is one the most confusing feature of Kinto. And it's used by default. 
Let's switch everything to `accounts` by default. 

* [x] Remove default value for policies
* [x] Rewrite tutorials
* [ ] Remove basicauth for tests (use specific *always true* policy)
* [x] Rewrite functional tests
* [x] Get green tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN \
     apt-get install -y nodejs; \
     cd kinto/plugins/admin; npm install; npm run build; \
     pip3 install -e /app[postgresql,memcached,monitoring] -c /app/requirements.txt; \
-    pip3 install kinto-pusher kinto-attachment kinto-emailer kinto-elasticsearch; \
+    pip3 install kinto-attachment kinto-emailer kinto-elasticsearch kinto-portier kinto-redis; \
     kinto init --ini $KINTO_INI --host 0.0.0.0 --backend=memory --cache-backend=memory; \
     apt-get purge -y -qq gcc libssl-dev libffi-dev libpq-dev curl nodejs; \
     apt-get autoremove -y -qq; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN \
     apt-get install -y nodejs; \
     cd kinto/plugins/admin; npm install; npm run build; \
     pip3 install -e /app[postgresql,memcached,monitoring] -c /app/requirements.txt; \
-    pip3 install kinto-pusher kinto-attachment ; \
+    pip3 install kinto-pusher kinto-attachment kinto-emailer kinto-elasticsearch; \
     kinto init --ini $KINTO_INI --host 0.0.0.0 --backend=memory --cache-backend=memory; \
     apt-get purge -y -qq gcc libssl-dev libffi-dev libpq-dev curl nodejs; \
     apt-get autoremove -y -qq; \
--- a/docs/api/1.x/authentication.rst
+++ b/docs/api/1.x/authentication.rst
@@ -14,8 +14,6 @@ But the main methods you are most likely to encounter will read the ``Authorizat
 * :ref:`Kinto Accounts <api-accounts>` expects a valid username and password as *Basic Auth*
 * :ref:`OpenID Connect <api-openid>` will validate a *Bearer Access Token* against an :term:`Identity Provider`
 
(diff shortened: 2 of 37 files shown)


n1k0:
Thank you for tackling this, I think it will benefit everyone eventually :+1:

leplatrem:
Some progress made here. I'm glad I found the motivation and time to tackle this :smile:

leplatrem:
Tests are green \o/

r+ anyone ?

Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN \
     apt-get install -y nodejs; \
     cd kinto/plugins/admin; npm install; npm run build; \
     pip3 install -e /app[postgresql,memcached,monitoring] -c /app/requirements.txt; \
-    pip3 install kinto-pusher kinto-attachment ; \
+    pip3 install kinto-attachment kinto-emailer kinto-elasticsearch kinto-portier kinto-redis; \
```
Natim: Are those official plugins that we want to ship with Kinto? Why would we want to install them all by default? Should we plan to move them to kinto.plugins?

kinto/config/kinto.tpl
```diff
@@ -92,21 +92,13 @@ multiauth.policies = basicauth
...
-#
-# Kinto-portier authentication
-# https://github.com/Kinto/kinto-portier
-# Set `multiauth.policies` to `portier`
-# multiauth.policy.portier.use = kinto_portier.authentication.PortierOAuthAuthenticationPolicy
-# kinto.portier.broker_url = https://broker.portier.io
-# kinto.portier.webapp.authorized_domains = *.github.io
-# kinto.portier.cache_ttl_seconds = 300
-# kinto.portier.state.ttl_seconds = 3600
+kinto.account_write_principals = account:admin
```
Natim: Should we enforce admin here as the default root user? I believe it would help people but it seems a bit brittle regarding security.

Dockerfile
leplatrem: I took the ones that were not coupled to any brand or commercial service

kinto/config/kinto.tpl
Natim: `account:admin` suggests that the admin account should exists. Also I would rather encourage people to use an email address as username because it simplifies the use of the sharing capabilities.

leplatrem: > account:admin suggests that the admin account should exists.

Yes maybe there are some place in the getting started docs that don't mention creating the admin. I added mentions in the install docs

> Also I would rather encourage people to use an email address as username because it simplifies the use of the sharing capabilities (although for the later we might need to verify the address)

We could have a regex for account ids, whose default would be an email adress for example. But please open another issue, I don't want to let this rot ;)

Natim: > But please open another issue, I don't want to let this rot

I wasn't suggesting that we should make any changes or enforce the email address there, I was just suggesting to use account:``your.email@host.tld`` rather than ``account:admin``

docs/api/1.x/authentication.rst
```diff
@@ -4,385 +4,121 @@ Authentication
...
+Since authentication is entirely pluggable in Kinto (see :ref:`configuration <configuration-authentication>`), the HTTP method used to authenticate requests may vary.
+But the main methods you are most likely to encounter will read the ``Authorization`` header. For example:
 
-How is that possible?
----------------------
+* :ref:`Kinto Accounts <api-accounts>` expects a valid username and password as *Basic Auth*
+* :ref:`OpenID Connect <api-openid>` will validate a *Bearer Access Token* against an :term:`Identity Provider`
 
-Kinto uses the request headers to authenticate the current user.
+If the authentication fails, the server will return a |status-401| error response. It can also return a |status-403| error response, which would mean that the operation performed on the resource is not allowed to unauthenticated users :)
```
glasserc: I'm a bit confused because here it seems like failed authentication should be obvious (401/403) but elsewhere the documentation says to request `/v1/` and try to infer that auth has failed because of no principals. Is that specific to `/v1/`? Does it depend on the auth mechanism?

leplatrem: You are right I'll rephrase this!

kinto/config/kinto.tpl
leplatrem: For the sake of simplicity I think I prefer `account:admin` :/

PR #1752, opened 2018-09-04
#1647: Deleting a collection doesn't delete access_control_entrries for its children

### LINKED ISSUE #1647: Deleting a collection doesn't delete access_control_entries for its children
`buckets.py` has an event listener that ensures that when a bucket is deleted, everything underneath it is recursively deleted. `collections.py` has one too but it doesn't appear to be as robust -- it doesn't have a wildcard to match objects more than one level below it (which might be OK, since only records are below collections now), and `delete_object_permissions` is only called on the collection rather than its children.

alexryabkov:
Fixes #1647

- [ ] Add documentation.
- [x] Add tests.
- [ ] Add a changelog entry.


tests/test_views_collections.py
```diff
@@ -158,6 +158,10 @@ def test_can_be_created_after_deletion_with_if_none_match_star(self):
         self.app.put_json(self.collection_url, MINIMALIST_COLLECTION,
                           headers=headers, status=201)
 
+    def test_records_permissions_are_removed_after_collection_deleted(self):
+        self.assertDictEqual(self.permission.get_object_permissions(
```
Natim: To test this, you would need to actually create someting and delete it from the test function itself.

alexryabkov: @Natim I thought the creation is covered in _setUp_ method:

```python
    def setUp(self):
        super().setUp()
        bucket = {**MINIMALIST_BUCKET,
                  'permissions': {'collection:create': ['system.Everyone'],
                                  'read': ['system.Everyone']}}
        self.app.put_json('/buckets/beers', bucket,
                          headers=self.headers)
        self.app.put_json(self.collection_url, MINIMALIST_COLLECTION,
                          headers=self.headers)
        resp = self.app.post_json(self.collection_url + '/records',
                                  MINIMALIST_RECORD,
                                  headers=self.headers)
        self.previous_ts = resp.headers['ETag']
        record_id = resp.json['data']['id']
        self.record_url = self.collection_url + '/records/{}'.format(record_id)
        self.app.delete(self.collection_url, headers=self.headers)
```
Do I still need to implement "create/delete" explicitly?

Natim: I would say that it is good enough for me then. Can you also make sure that the collection permissions is empty?

alexryabkov: Sure, will update PR later today

alexryabkov:
Sorry, I forgot to update my local fork to the latest upstream... I will update CHANGELOG.rst once again.

alexryabkov:
@Natim @leplatrem Guys, I'm really sorry for such a big number of commits, I should be move attentive...

glasserc:
Thanks for your contribution!

alexryabkov:
@glasserc @Natim @leplatrem Thank you, too!

PR #1758, opened 2018-09-10
Fix DeprecationWarning for unrecognized backslash escapes

glasserc:
This became a warning as of Python 3.6 -- see
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ python-memcached==1.59
 raven==6.9.0
 repoze.lru==0.7
 requests==2.19.1
-simplejson==3.16.1
+simplejson==3.16.0
 six==1.11.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -28,7 +28,7 @@ python-memcached==1.59
 raven==6.9.0
 repoze.lru==0.7
 requests==2.19.1
-simplejson==3.16.1
+simplejson==3.16.0
```
Natim: Why did we need to rollback?

Natim: https://github.com/simplejson/simplejson/issues/237

glasserc: Yes, as noted in the commit message.

PR #1770, opened 2018-09-19
Use timestamp to communicate collection_timestamp of the parent

### LINKED ISSUE #1769: What should `event.payload["timestamp"]` represent?
Refs #945 and #1731. The documentation describes the timestamp field as representing "the time of the event". However, as far as I can tell, the `timestamp` field in an event represents a collection timestamp calculated relatively early in the request, before the action that triggered event happened. It doesn't represent the time that the event happens, nor the current last_modified / collection_timestamp of anything reflected in the event.

I think I'd like to change the `timestamp` key in the payload to correspond to the updated collection timestamp of the `(parent_id, collection_id)` for which the event was fired.

glasserc:
Don't re-use old self.timestamp when notifying of event. Instead,
re-compute it fresh.

Fixes #1769.

- [x] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/docs/core/notifications.rst
+++ b/docs/core/notifications.rst
@@ -110,7 +110,8 @@ Payload
 The :class:`kinto.core.events.ResourceChanged` and :class:`kinto.core.events.AfterResourceChanged`
 events contain a ``payload`` attribute with the following information:
 
-- **timestamp**: the time of the event
+- **timestamp**: the collection timestamp of the ``(parent_id, resource_name)`` pair for
+  which this event was sent
 - **action**: what happened. 'create', 'update' or 'delete'
 - **uri**: the uri of the impacted resource
 - **user_id**: the authenticated user id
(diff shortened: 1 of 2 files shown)


glasserc:
Another approach would be to turn the `@reify`'d `timestamp` property into a `@property`. As far as I can tell, the `@reify` is a performance optimization -- fewer database round trips. Taking it out doesn't seem to break any tests (except for a few that rely on the ability to set the `timestamp` of a resource). 

Running the test suite with `@reify` on my laptop finishes in 225-235 seconds. Running with `@property` finishes in 240-255. So I guess we should probably keep the `@reify`.

@leplatrem suggested that I add a comment to this change explaining it, which I'll do.

glasserc:
I just added one test (for `put`) in a section where tests are also present for other verbs. What do you think about this, @leplatrem ? Is this test even in the right place?

PR #1787, opened 2018-10-02
Raises 400 invalid request when field parameters contain a NUL character

### LINKED ISSUE #1704: Crash in Postgres backend when querystring contains NUL character (0x00)
screenshot from 2018-07-02 10-40-31

:laughing: 

screenshot_2018-07-02 valueerror a string literal cannot contain nul 0x00 characters

Cnidarias:
Fixes #1704 

- [x] Add tests.
- [ ] Add a changelog entry.

As suggested I discussed I implemented a fix for the null char.

I did however not include this in the 

``` native_value() ``` 
function as this function is never called on the query parameters, and including some exception to catch null bytes would break the tests for the function.

I am open to some better way of fixing the issue.

For the testing I did not use the already existing 400 test as the failure only shows when using the postgres backend.

### DIFF
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -1022,6 +1022,9 @@ class UserResource:
                 if has_invalid_value:
                     raise_invalid(self.request, **error_details)
 
+            if '\x00' in field:
+                raise_invalid(self.request, **error_details)
+
             if field == self.model.modified_field and value == '':
                 raise_invalid(self.request, **error_details)
 
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -1022,7 +1022,7 @@ class UserResource:
                 if has_invalid_value:
                     raise_invalid(self.request, **error_details)
 
-            if '\x00' in field:
+            if '\x00' in field or '\x00' in str(value):
                 raise_invalid(self.request, **error_details)
 
             if field == self.model.modified_field and value == '':
--- a/tests/core/test_views_errors.py
+++ b/tests/core/test_views_errors.py
@@ -7,7 +7,7 @@ from kinto.core.errors import ERRORS, http_error
 from kinto.core.testing import FormattedErrorMixin
 from kinto.core.storage import exceptions as storage_exceptions
 
-from .support import BaseWebTest, authorize
+from .support import BaseWebTest, authorize, PostgreSQLTest
 
 
 class ErrorViewTest(FormattedErrorMixin, BaseWebTest, unittest.TestCase):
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: kinto/core/resource/__init__.py [hunk 1]
```diff
@@ -1022,6 +1022,9 @@ def _extract_filters(self):
                 if has_invalid_value:
                     raise_invalid(self.request, **error_details)
 
+            if '\x00' in field:
```
leplatrem: what about `value` ?

tests/core/test_views_errors.py
```diff
@@ -197,6 +197,16 @@ def test_500_passes_exception_as_exc_info(self):
         self.assertEqual(mocked_logger.error.call_args[1]['exc_info'], custom_error)
 
 
+class ErrorPostgresViewTest(FormattedErrorMixin, PostgreSQLTest, unittest.TestCase):
```
leplatrem: Since the fix happens at the resource level I would prefer not to involve PostgreSQL here.

Could you reproduce the original issue with that test?

tests/core/test_views_errors.py
```diff
@@ -197,6 +197,16 @@ def test_500_passes_exception_as_exc_info(self):
...
+    sample_url = "/mushrooms"
+
+    def test_400_with_invalid_query_field(self):
+        response = self.app.get(self.sample_url + '?field\x00="2"',
```
leplatrem: would be great with `?field="\x00"` too ;)

tests/core/test_views_errors.py
Cnidarias: Yes, when testing against Postgre I get the original error that you posted in the issue. If I just use the ```BaseWebTest``` then I was unable to reproduce the error.

tests/core/test_views_errors.py
```diff
@@ -197,6 +197,23 @@ def test_500_passes_exception_as_exc_info(self):
         self.assertEqual(mocked_logger.error.call_args[1]['exc_info'], custom_error)
 
 
+class ErrorPostgresViewTest(FormattedErrorMixin, PostgreSQLTest, unittest.TestCase):
+    sample_url = "/mushrooms"
+
+    def test_400_with_invalid_query_field(self):
```
leplatrem: instead of having one test method that tests two things, it's better to have two methods:

- `test_400_when_query_field_contains_nul_character`
- `test_400_when_query_value_contains_nul_character`

tests/core/test_views_errors.py
leplatrem: Ok I see.

The thing is that now we have a safety check / fix (and some tests that assert the proper behavior) at the resource level. We don't have to imply PostgreSQL anymore. `BaseWebTest` should be enough :wink:

leplatrem:
Thanks!

If you are interesting in more easy-picks, don't hesitate to look throughout the https://github.com/Kinto org ;)

PR #1797, opened 2018-10-07
Start on type annotations

Cnidarias:
So this is a start on adding type annotations

I wanted to get some feedback before I continue on this - I just used monkeytype to generate the annotations - this does yield pretty good results but with some of these functions the function definitions is _very long_  

Should we just continue on this road? Any other options/improvements because this isn't really pretty.
(it will also fail flake8 right now)

I guess one option would be to let something like black run over the code afterwards to get some decent formatting but I am quite unsure.

So some feedback would be great!

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -12,6 +12,10 @@ from kinto.core.storage.postgresql.migrator import MigratorMixin
 from kinto.core.utils import COMPARISON
 
 
+from kinto.core.storage import Filter, Sort
+from kinto.core.storage.postgresql.client import PostgreSQLClient
+from pyramid.config import Configurator
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 logger = logging.getLogger(__name__)
 HERE = os.path.dirname(__file__)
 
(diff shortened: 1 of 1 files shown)


leplatrem:
Thanks for starting this :) 

@Natim would be happy if we would enable Black here ;) 

I have one feedback: I don't think having that kind of annotations are helpful: `List[Dict[str, Union[int, str]]]:`. If we could have something like `List[StoredObject]` and somewhere have `StoredObject` defined to be a dict with at least an id, last_modified...
Is it possible?

Cnidarias:
Sure,

yes this is very possible. I know some larger projects basically have a type def file where they define such horrible types to shorten them.

I will let black run over it and refactor some of the larger type annotations tonight when I get home from work, then we can see if the direction we are heading is any good or not

Natim:
In the meantime black work has been merged.

leplatrem:
@Cnidarias should we keep this PR open? 
Now that we have black, and that the core is a bit more polished (#1827  #1932), we could get back to this :)

leplatrem:
I'll take this as a «no» :) 

Thanks for starting this, don't hesitate to shout out if you'd like to come back to this!

PR #1799, opened 2018-10-08
Blackify code base

Cnidarias:
Letting black run over everything

- [ ] Add documentation.
- [ ] Add a changelog entry.

So in light of my other pull request I suggested that we run black over everything and @leplatrem seemed to be in favor so here we are.

I figured it would be best if we did this before adding all the type annotations to not have too many changes at once.

I let black run over everything, adding a pre-commit (https://pre-commit.com) config file as well as a config file for black.

I also had to add an ignore E203 in order for flake8 to be happy.

If this is okay then I would still have to add some documentation on how to set things up for development but I do think the result is actually quite nice!

Thoughts?

### DIFF
--- a/docs/community.rst
+++ b/docs/community.rst
@@ -151,7 +151,6 @@ Get started!
 
     therapist install
 
-
 9. Commit your changes and push your branch to GitHub::
 
     git add .
--- a/docs/community.rst
+++ b/docs/community.rst
@@ -139,17 +139,26 @@ Get started!
 
    Now you can make your changes locally.
 
-6. When you're done making changes, check that your changes pass the tests::
+6. When you're done making changes, check that your changes pass linting (requires python >= 3.6)::
+
+    tox -e lint
+
+7. Don't forget to check that your changes pass the tests::
 
     make tests
 
-7. Commit your changes and push your branch to GitHub::
+8. (Optional) Install a git hook::
 
-    $ git add .
-    $ git commit -m "Your detailed description of your changes."
-    $ git push origin name-of-your-bugfix-or-feature
+    therapist install
 
-8. Submit a pull request through the GitHub website.
+
+9. Commit your changes and push your branch to GitHub::
+
+    git add .
+    git commit -m "Your detailed description of your changes."
+    git push origin name-of-your-bugfix-or-feature
+
+10. Submit a pull request through the GitHub website.
 
 
 Testing methodology
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,0 +1,2 @@
+therapist==1.5.0
+black==18.9b0
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,2 +1,4 @@
 therapist==1.5.0
 black==18.9b0
+flake8
+setuptools<36
--- a/tox.ini
+++ b/tox.ini
@@ -42,3 +42,4 @@ deps =
 
 [flake8]
 max-line-length = 99
+ignore = E203, E266, E501, W503
--- a/tox.ini
+++ b/tox.ini
@@ -16,11 +16,12 @@ deps =
     statsd
 install_command = pip install {opts} {packages}
 
-[testenv:flake8]
-commands = flake8 kinto tests docs/conf.py
+[testenv:lint]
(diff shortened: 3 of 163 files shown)


leplatrem:
@Natim ?

leplatrem:
Note: @peterbe recommends therapist https://github.com/peterbe/hashin/pull/80/files

leplatrem:
Be careful commit edbf3b8  should not be part of this PR

Cnidarias:
So I basically just added the stuff you linked -- not sure I did it correctly though especially regarding therapist/travis/tox

But it seems like its fine?

Still need to add some documentation though.

Cnidarias:
Hey,

yes the diff is quite massive. 
This does beg another question, when looking at the diffs alot of changes are changing single quotes to double quotes

Black does provide the option to omit normalizing quotes. 

After testing it - the diff would still be massive. When omitting the quite normalization black still touches all but 24 files. 


I have also added some documentation now. 
I would prefer to rebase once we are happy with the documentation and everything else because there will always be merge conflicts as of now.

docs/community.rst
```diff
@@ -139,17 +139,27 @@ Get started!
 
    Now you can make your changes locally.
 
-6. When you're done making changes, check that your changes pass the tests::
+6. When you're done making changes, format your code with `black <https://github.com/ambv/black>`_ (black requires python >= 3.6)::
+
+    python -m pip install black
+    black .
```
leplatrem: I would suggest to install `black` only and run it via `tox -e lint`  (tox and therapist will be installed via dev-requirements)

docs/community.rst
```diff
@@ -139,17 +139,27 @@ Get started!
...
+
+7. Don't forget to check that your changes pass the tests::
 
     make tests
 
-7. Commit your changes and push your branch to GitHub::
+8. (Optional) Install a git hook with `therapist <https://github.com/rehandalal/therapist>`_ ::
+
+    python -m pip install therapist
```
leplatrem: this line can be omitted since already in dev-requirements

leplatrem:
Almost there! :tada: 

* I would like to avoid configuration of black if possible. Double quotes are fine.
* Since python 3.5 is still officially supported, we have to adapt. Suggesting its installation manually in the dev docs is fine IMO

Thanks :pray:

leplatrem:
Let me know if you need help or support ;) 

Actually I would be very interested with this PR, because I started to work on #710 using `xargs sed` like an animal :)

Cnidarias:
Okay - so I went ahead and got rid of my changes because I did not feel like merging 40+ merge conflicts.

This commit now has everything we had earlier with the additional changes that we now have a lint-requirements.txt file as well as adapting the documentation.

The code has not yet been formatted with black so it will be easier for me to rebase it once we are ready to merge.

lint-requirements.txt
```diff
@@ -0,0 +1,2 @@
+therapist==1.5.0
+black==18.9b0
```
leplatrem: Maybe we could add:
```
    flake8
    setuptools<36
```

BTW @Natim do you remember why `setuptools<36`?

docs/community.rst
```diff
@@ -139,17 +139,26 @@ Get started!
...
+
+7. Don't forget to check that your changes pass the tests::
 
     make tests
 
-7. Commit your changes and push your branch to GitHub::
+8. (Optional) Install a git hook::
+
+    therapist install
+
```
leplatrem: looks like an extra line

tox.ini
```diff
@@ -16,11 +16,12 @@ deps =
     statsd
 install_command = pip install {opts} {packages}
 
-[testenv:flake8]
-commands = flake8 kinto tests docs/conf.py
```
leplatrem: shouldn't you keep `flake8` for py35 on Travis?

docs/community.rst
Cnidarias: Don't you have to tell therapist to create the hook in your .git dir?

tox.ini
Cnidarias: Yes I should have removed it in travis.

leplatrem:
:) here we go :) thanks again ;)

PR #1817, opened 2018-10-15
Use dot as bucket/collection separator in cache control settings (fixes #1815)

### LINKED ISSUE #1815: Per collection caching using dot separator
Per collection caching can be configured in settings using `_` as separator:

```ini
kinto.bid_cid_record_cache_expires_seconds = 60
```
Since `_` is valid character in both bucket and collection ids, this could be confusing. I propose that we change it to `.`

leplatrem:
fixes #1815


Natim:
Except for the test not passing I believe the rationale of this change makes total sense.

PR #1822, opened 2018-10-17
Allow dots in records IDs

leplatrem:
What do you think about making such a change?

(*It would serve a very specific use-case that we have: use domain names as records IDs.*)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ This document describes changes between each past release.
 11.1.0 (unreleased)
 -------------------
 
+**API**
+
+- Allow dots (``.``) in records IDs
+
+API is now at version **1.21**. See `API changelog`_.
+
 **New features**
 
 - Add ability to configure the ``project_name`` in settings, shown in the `root URL <https://kinto.readthedocs.io/en/stable/api/1.x/utilities.html#get>`_ (fixes #1809)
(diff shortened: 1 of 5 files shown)


Natim:
I am a bit reluctant to add this for records only, also it is not wishable for buckets and collections because it will conflict with the settings separator between buckets and collections.

I can understand why it could make sense to use a domain name as an ID.

Usually what we did in the past is to md5 encode the domain to build a guessable UUID.

leplatrem:
> Usually what we did in the past is to md5 encode the domain to build a guessable UUID.

Indeed, if they use a script to upload record, they can just do:
```py
record = {"id": domain.replace(".", "-"), "domain": domain}
```

And do the same when looking it up: 
```js
RemoteSettings("tippytop").get({filters: {id: domain.replace(".", "-")}});
```

And we can set a disabled widget in Kinto-Admin so that user can't change the domain value in order to maintain consistence between the id and domain fields

glasserc:
I think the outstanding issue is about how users actually put data in. Usually they use kinto-admin. Do we make them md5 the domain name themselves? Do we add kinto-admin support for this kind of ID generation?

Natim:
As far as I know the kinto admin doesn't let you pick a ID, it is automatically generated when you enter a new record.

PR #1826, opened 2018-10-18
Exclude folders for flake8/black

leplatrem:
screenshot from 2018-10-18 14-43-01

I think it should be faster than this...


leplatrem:
```therapist run --fix kinto tests                                
black ............................................................... [SUCCESS]
flake8 .............................................................. [SUCCESS]

-------------------------------------------------------------------------------
Completed in: 43.26s
```

PR #1829, opened 2018-10-18
Don't native value

glasserc:
Part of https://github.com/Kinto/kinto-http.js/issues/311 is due to the fact that the `http_api_version` (which is currently `1.20`) is understood as `1.2`. This PR adds a test to demonstrate the problem and offers the first thing I thought of as a solution. Unfortunately I'm not sure the solution is worth pursuing, because it means that settings read from .ini files are always strings. This means we need to update places where we access settings to explicitly convert them to numbers or something.

Another alternative might be to just represent the `http_api_version` as `"1.20"`, which is maybe more honest anyhow?

### DIFF
--- a/kinto/core/utils.py
+++ b/kinto/core/utils.py
@@ -125,7 +125,9 @@ def read_env(key, value):
     :returns: the value from environment, coerced to python type
     """
     envkey = key.replace(".", "_").replace("-", "_").upper()
-    return native_value(os.getenv(envkey, value))
+    if envkey in os.environ:
+        return native_value(os.environ[envkey])
+    return value
 
 
 def encode64(content, encoding="utf-8"):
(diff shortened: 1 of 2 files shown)


glasserc:
The problem is all the tests that break because up until here, everything was being native-valued. For example:

```
  File "/home/travis/build/Kinto/kinto/kinto/plugins/quotas/listener.py", line 178, in on_resource_changed
    if bucket_info["storage_size"] > bucket_max_bytes:
TypeError: '>' not supported between instances of 'int' and 'str'
```

Like it or not, I think this would be a breaking change. Again, I wonder if the smart change is just to force `http_api_version` to be a string.

glasserc:
Closing in favor of #1835.

PR #1855, opened 2018-10-31
Build admin plugin in CI

### LINKED ISSUE #1707: Run `npm install` in TravisCI to test kinto/plugins/admin/package.json
Currently we don't test `npm install` of `kinto/plugins/admin/package.json` in CI.

Appart from the release step (see `kinto/plugins/admin/release_hook.py`), we never run it.

I believe we could add a TravisCI task that runs `make build-kinto-admin`

https://github.com/Kinto/kinto/blob/be8c22bf15189f5afa15592edc53535454093c61/Makefile#L64-L65

itaisteinherz:
Fixes #1707.
I tried to add a `make` task which also tests the plugin by running `react-scripts test`, but that failed and I didn't have the time to look into it. This should probably be fixed and tested in CI in the future.

- [x] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_script:
   - sudo /etc/init.d/postgresql restart
   - psql -c "CREATE DATABASE testdb ENCODING 'UTF8' TEMPLATE template0;" -U postgres
   - make version-file
+  - make build-kinto-admin
 script:
   - tox -e $TOX_ENV
 after_success:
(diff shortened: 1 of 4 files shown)


leplatrem:
Thanks for starting this!

This seems to work from what I see in the Travis CI log.

But that's a pity, the command is ran for every environment, even like docs or linting...  I think the admin build should have a dedicated entry in the matrix instead. 
Is that what you say that fails?

itaisteinherz:
I'm close this and open another PR with just the added changes, since this PR got really crowded with commits for some reason.

PR #1857, opened 2018-10-31
Build admin plugin in CI

### LINKED ISSUE #1707: Run `npm install` in TravisCI to test kinto/plugins/admin/package.json

itaisteinherz:
I've closed #1855 as it got messy for some reason, and opened this PR instead. This fixes #1707.

- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_script:
   - sudo /etc/init.d/postgresql restart
   - psql -c "CREATE DATABASE testdb ENCODING 'UTF8' TEMPLATE template0;" -U postgres
   - make version-file
-  - make build-kinto-admin
+  - [[ $TOX_ENV == "py37" ]] && make build-kinto-admin
 script:
   - tox -e $TOX_ENV
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,4 +48,6 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: true
+      env:
+        - ENV=build-kinto-admin
       script: make build-kinto-admin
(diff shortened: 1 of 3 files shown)


itaisteinherz:
@leplatrem Continuing the discussion from https://github.com/Kinto/kinto/pull/1855#issuecomment-434628803:
> I think the admin build should have a dedicated entry in the matrix instead.

I'll add a dedicated entry to the matrix.

> Is that what you say that fails?

No, what I meant is that running (from the project root):
```bash
cd kinto/plugins/admin/; npm run test
```
fails.

itaisteinherz:
@leplatrem I ended up using a simple bash check to build the plugin only on the main matrix entry (Python 3.7). 47c6ef03317277aa1cf3f5fc9ab9625c281e012a

Natim:
@itaisteinherz why did you pick this option rather than adding something to the matrix?

Natim:
Thank you for your work on this @itaisteinherz

itaisteinherz:
> @itaisteinherz why did you pick this option rather than adding something to the matrix?

As I wrote above:

> This keeps the CI build as simple as possible and reduces unnecessary complications.

For every entry in the matrix, Travis has to start a container, install all of the necessary packages and frameworks on it, and finally start Kinto, and only then can it start the tests. Avoiding creating additional matrix entries makes the CI faster.

.travis.yml
```diff
@@ -45,3 +45,9 @@ matrix:
       script:
         - make runkinto & sleep 5
         - make functional
+    - python: 3.7
```
itaisteinherz: Why was this added? @Natim

Natim:
> Avoiding creating additional matrix entries makes the CI faster.

Maybe but it also makes it more harder to understand what failed.

PR #1902, opened 2018-11-30
Add black clean rule

Natim:
- [x] Add a changelog entry.


Natim:
I eventually rebased because the history wasn't relevant anymore.

PR #1931, opened 2018-12-10
no Total-Records header in GET requests. only in HEAD requests.

### LINKED ISSUE #1624: What should we do with `Total-Records`?
Extracted from #1622, #1615, #1507, etc. One of the fundamental issues here is whether we provide a `Total-Records` header in plural endpoint responses. Doing so has a performance cost, because we have to enumerate every single record that matches the request rather than just the first N (where N is given by `limit=N` or the maximum response length in the config).

Here are some proposals for what to do with `Total-Records`, each of which has a different performance cost and accuracy tradeoff.

1. Compute `Total-Records` by counting every record that matches the request's current filter, in a way that is free from race conditions. This is what 9.0.0 does and the behavior I try to preserve in #1622. This is the most expensive option and the most accurate.

2. Compute `Total-Records` but with some amount of latitude for race conditions. This is what is implied by #1615 and @peterbe's suggestions on #1622 of using a cache to store the previously-computed value for a short duration. Different approaches can still fall into this category even if they have different tolerances for race conditions. In general, the greater the tolerance for race conditions, the better the performance gains.

3. Admit defeat with the `Total-Records` header. We can't outright remove it from the response without breaking clients, but we could stick a constant value in there or otherwise provide something that will allow pagination code to continue working but without actually computing `Total-Records`. For example, if there were any records to return, we could put `offset+1` in there, and otherwise `previous_offset`. Or if there were any records, we could put `max_int` and otherwise 0.

Of course, we could cut a new API version where `Total-Records` is not computed by default. There are some other alternatives that could alleviate the performance cost of `Total-Records`, such as allowing clients to tell us (e.g. `Prefer: no-total-records` or `?select=!total_records` or something) not to compute it, but because the current default is to compute it and because we can't change that default without breaking the API, I feel that this is not going to help much in practice.

peterbe:
Fixes #1624

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.
- [ ] ~~If you added a new configuration setting, update the [`kinto.tpl`](https://github.com/Kinto/kinto/blob/master/kinto/config/kinto.tpl) file with it.~~
- [x] A bunch of storage tests use the now deprecated internal API of `storage.get_all` when in fact it should be using `storage.list_all` or `storage.count_all`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,7 @@ This document describes changes between each past release.
 - Ignore admin plugin node_modules folder while running black (#1902)
 - Remove regexp py36 warnings. (#1907)
 - Changed psycopg2 dependency for psycopg2-binary. (#1905)
+- Renamed core notions (ie. record and collection) (#710)
 
 
 11.2.1 (2018-12-09)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,8 @@ This document describes changes between each past release.
   ``ShareableResource`` class may also break.
 - GET requests no longer include the ``Total-Records`` header. To get a count in a collection
   you need to do a HEAD request. And the new header name is ``Total-Objects``. (#1624)
+- Remove the ``UserResource`` class. And ``ShareableResource`` is now deprecated in
+  favor of ``Resource``.
 
 **Bug Fixes**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ This document describes changes between each past release.
   by the generic terms *resource* and *object*. Plugins that subclass the internal
   ``ShareableResource`` class may also break.
 - GET requests no longer include the ``Total-Records`` header. To get a count in a collection
-  you need to do a HEAD request. (#1624)
+  you need to do a HEAD request. And the new header name is ``Total-Objects``. (#1624)
 
 **Bug Fixes**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ This document describes changes between each past release.
 
 - Remove Python 3.5 support and upgrade to Python 3.6. (#1886)
 - Remove `record` from UnicityError class (#1919). This enabled us to fix #1545.
+- Storage backend API has changed, notions of collection and records were replaced
+  by the generic terms *resource* and *object*. Plugins that subclass the internal
+  ``ShareableResource`` class may also break.
 - GET requests no longer include the ``Total-Records`` header. To get a count in a collection
   you need to do a HEAD request. (#1624)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ This document describes changes between each past release.
 
 - Remove Python 3.5 support and upgrade to Python 3.6. (#1886)
 - Remove `record` from UnicityError class (#1919). This enabled us to fix #1545.
+- GET requests no longer include the ``Total-Records`` header. To get a count in a collection
+  you need to do a HEAD request. (#1624)
 
 **Bug Fixes**
 
--- a/docs/api/1.x/_details-get-list.rst
+++ b/docs/api/1.x/_details-get-list.rst
@@ -1,6 +1,3 @@
-A ``Total-Records`` response header indicates the total number of objects
-of the list (not the response, since it can be paginated).
-
 A ``Last-Modified`` response header provides a human-readable (rounded to second)
 of the current collection timestamp.
 
--- a/docs/api/1.x/_details-head-list.rst
+++ b/docs/api/1.x/_details-head-list.rst
@@ -1,0 +1,10 @@
+Unlike GET requests, HEAD requests contain an additional header called ``Total-Objects``
+(and ``Total-Records`` for backwards compatibility) which is a count of all objects
+in the collection with the current filtering.
+
+List of available URL parameters for HEAD
+-----------------------------------------
+
+- ``<prefix?><field name>``: :doc:`filter <filtering>` by value(s)
+- ``_since``, ``_before``: polling changes
+- ``_token``: :doc:`pagination token <pagination>`
--- a/docs/api/1.x/history.rst
+++ b/docs/api/1.x/history.rst
@@ -53,7 +53,7 @@ Retrieve history
   .. sourcecode:: http
 
         HTTP/1.1 200 OK
-        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
+        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, ETag, Pragma, Cache-Control, Backoff, Next-Page
         Cache-Control: no-cache, no-store
         Content-Length: 1906
         Content-Type: application/json; charset=UTF-8
@@ -61,7 +61,6 @@ Retrieve history
         Etag: "1469006098757"
         Last-Modified: Wed, 20 Jul 2016 09:14:58 GMT
         Server: waitress
-        Total-Records: 4
 
         {
             "data": [
--- a/docs/api/1.x/history.rst
+++ b/docs/api/1.x/history.rst
@@ -53,7 +53,7 @@ Retrieve history
   .. sourcecode:: http
 
         HTTP/1.1 200 OK
-        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
+        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Objects, ETag, Pragma, Cache-Control, Backoff, Next-Page
         Cache-Control: no-cache, no-store
         Content-Length: 1906
         Content-Type: application/json; charset=UTF-8
--- a/docs/api/1.x/history.rst
+++ b/docs/api/1.x/history.rst
@@ -53,7 +53,7 @@ Retrieve history
   .. sourcecode:: http
 
         HTTP/1.1 200 OK
-        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, ETag, Pragma, Cache-Control, Backoff, Next-Page
+        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
         Cache-Control: no-cache, no-store
         Content-Length: 1906
         Content-Type: application/json; charset=UTF-8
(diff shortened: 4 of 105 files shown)


peterbe:
Do I need to put this branch in the `Kinto` repo to start the CI tests?

leplatrem:
@peterbe, don't worry about the conflicts, I can do it and try to push a merge on your branch.

leplatrem:
I have one major question though: what do we do about our clients:

https://github.com/Kinto/kinto-http.js/blob/27737b9e62df2aff6adc32523708cd7e66131282/src/base.js#L540 --> `Nan`
https://github.com/Kinto/elm-kinto/blob/d5b5337166f23d130322858ace30c0a80abdca28/src/Kinto.elm#L299-L301 --> `0`

Maybe we should first release of kinto-http.js where the totalRecords is obtained via a HEAD? 
(I don't know, truely asking..)

peterbe:
Haven't dug in yet but the base.js you pointed to is inside a function called `paginatedList` and the doc string...

```
   * Fetch some pages from a paginated list, following the `next-page`
   * header automatically until we have fetched the requested number
   * of pages. Return a response with a `.next()` method that can be
   * called to fetch more results.
```

So, if the idea is that you use...

```javascript 
for (const page in client.paginatedList(args)) {
  ...
}
```
...then we're fairly certain the total records is not needed or used. The pagination is done using `Next-Page`.

Perhaps my brain is a bit end-of-day fried and there's a lot of mind-bending async in that code but it looks like the `totalRecords` isn't actually used.

docs/api/1.x/_details-get-list.rst
```diff
@@ -1,6 +1,3 @@
-A ``Total-Objects`` response header indicates the total number of objects
-of the list (not the response, since it can be paginated).
```
leplatrem: We should probably have a `_details-head-list.rst` where we mention that the header can be obtained with HEAD

kinto/core/resource/__init__.py
```diff
@@ -312,28 +331,47 @@ def plural_get(self):
...
+        # and a list if you pass `count_only=False`
+        objects = self.model.get_objects(
             filters=filters,
             sorting=sorting,
-            limit=limit,
+            limit=limit + 1,  # See bigger explanation above.
             pagination_rules=pagination_rules,
             include_deleted=include_deleted,
+            count_only=head_request,
         )
```
leplatrem: I'm not a big fan of this pattern. I don't think the return type of a method should differ based on a parameter value. What do you think?

docs/api/1.x/selecting_fields.rst
```diff
@@ -31,13 +31,12 @@ Nested objects fields are specified using dots (e.g. ``address.street``).
 
     HTTP/1.1 200 OK
     Access-Control-Allow-Origin: *
-    Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Content-Length, ETag, Next-Page, Total-Objects, Last-Modified
+    Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Content-Length, ETag, Next-Page, Total-Records, Last-Modified
```
leplatrem: to fix (post-merge)

docs/api/1.x/records.rst
```diff
@@ -245,14 +245,13 @@ Retrieving stored records
     .. sourcecode:: http
 
         HTTP/1.1 200 OK
-        Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Next-Page, Total-Objects, Last-Modified, ETag
+        Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Next-Page, Total-Records, Last-Modified, ETag
```
leplatrem: to fix (post-merge)

glasserc:
Yeah, my hypothesis is that nobody actually uses the `totalRecords` field on the `paginatedList` result, but just pages the results. I think automatically including a call to `getTotalRecords` (on each page...) would be a prohibitive performance penalty and would rather not do that.

peterbe:
> This PR badly needs a rebase

Do you mean a squash? Isn't that done with the green button when it's fully approved? 

Or do you mean it needs the latest and greatest of master?

It's hard to remember what different projects have as best practices with active PRs and the landing protocol.

Natim:
Usually in the Kinto org we merge without squashing to keep the history and pace of the pull-request. Here we just need to fix tests I guess

Natim:
No don't worry about that.

glasserc:
> Do you mean a squash?

No, I meant a rebase, meaning moving these comments onto `master` so that we get fewer wacky diffs like this:

screenshot from 2018-12-11 21-29-27
screenshot from 2018-12-11 21-29-48

leplatrem:
The wacky diffs come from some mistakes I made during the painful merge I guess.

leplatrem:
@peterbe, I merged and repaired the merge :) Tests now pass.

I believe what remains is:
- [ ] be extremely clear about the removal of the header in GET in changelog
- [ ] Maybe we could have a server setting to maintain the header if set to True (default: false) for situations where clients rely on this info
- [ ] document that the header can be obtained with HEAD
- [ ] figure out a way to nicely document that we `Total-Records` is now deprecated in favor of `Total-Objects` because of #710
- [ ] improve the models code (eg. two methods instead of count_only param)

peterbe:
@leplatrem It's a bit of a mess but check out the latest couple of commits related to documentation and whether it solves the `figure out a way to nicely document that we Total-Records is now deprecated in favor of Total-Objects because of` checkbox above. 

Also, see the latest commit regarding how I broken up `model.get_objects` so we no longer need this `count_only=boolean` stuff.

docs/api/1.x/records.rst
```diff
@@ -265,6 +264,49 @@ Retrieving stored records
...
+
+    :synopsis: Includes a response header to tell how many records exist in the collection.
+
+    **Requires authentication**
+
+    **Example Request**
+
+    .. sourcecode:: bash
+
+        $ http head http://localhost:8888/v1/buckets/blog/collections/articles/records --auth="bob:p4ssw0rd" --verbose
```
leplatrem: nit: upercase HEAD ?

docs/api/1.x/records.rst
```diff
@@ -265,6 +264,49 @@ Retrieving stored records
             ]
         }
 
+.. _records-head:
+
+Count of stored records
+=======================
```
leplatrem: side note: In theory, this applies to every resources (buckets, collections, groups, etc.) but I'm fine if we only document it for records...

kinto/core/storage/memory.py
```diff
@@ -354,7 +353,29 @@ def get_all(
             pagination_rules=pagination_rules,
             limit=limit,
         )
-        return objects, count
+        return objects
+
+    @deprecate_kwargs({"collection_id": "resource_name"})
```
leplatrem: We are not supposed to have deprecate stuff for methods that were just introduced. I guess there is a path somewhere that makes a call to this `count_all` method with the wrong parameter...  

IMO the caller should be in charge of rebinding the old `collection_id` parameter to `resource_name`, but I guess we could avoid having this here!

kinto/core/storage/testing.py
```diff
@@ -261,102 +263,97 @@ def test_delete_raise_when_unknown(self):
...
         )
         abc2 = self.create_object(
             parent_id="abc2", resource_name="c", obj={"id": "abc", "secret_data": "abc2"}
         )
-        objects, total_objects = self.storage.get_all(
-            parent_id="ab*", resource_name="c", include_deleted=True
-        )
+        objects = self.storage.list_all(parent_id="ab*", resource_name="c", include_deleted=True)
         self.assertEqual(len(objects), 2)
+        total_objects = self.storage.count_all(parent_id="ab*", resource_name="c")
```
leplatrem: :+1:

tests/test_views_permissions.py
```diff
@@ -182,8 +182,11 @@ def test_permissions_fields_can_be_selected(self):
         self.assertNotIn("resource_name", resp.json["data"][0])
 
     def test_permissions_list_can_be_paginated(self):
-        resp = self.app.get("/permissions?_limit=2", headers=self.headers)
+        resp = self.app.head("/permissions", headers=self.headers)
         self.assertEqual(resp.headers["Total-Objects"], "5")
```
leplatrem: :+1:

kinto/core/storage/__init__.py
```diff
@@ -280,7 +281,25 @@ def purge_deleted(
...
+
+        This method exists solely to make the transition easier.
+        """
+        warnings.warn("Use either self.list_all() or self.count_all()", DeprecationWarning)
+        list_ = self.list_all(*args, **kwargs)
+        kwargs.pop("pagination_rules", None)
+        kwargs.pop("limit", None)
+        kwargs.pop("sorting", None)
+        kwargs.pop("include_deleted", None)
+        count = self.count_all(*args, **kwargs)
```
leplatrem: According to this, there should be no call to `.list_all()` and `.count_all()` with the parameter `collection_id`.

docs/api/1.x/history.rst
```diff
@@ -53,15 +53,14 @@ Retrieve history
   .. sourcecode:: http
 
         HTTP/1.1 200 OK
-        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Objects, ETag, Pragma, Cache-Control, Backoff, Next-Page
+        Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
```
glasserc: More merge messiness.

CHANGELOG.rst
```diff
@@ -34,7 +36,6 @@ This document describes changes between each past release.
 -------------------
 
 - Still supports jsonschema 2.6 before 3.0 is released as a production release. (#1923)
-- Renamed core notions (ie. record and collection) (#710)
```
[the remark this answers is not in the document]
peterbe: Either way, I think that line never belonged there, right? It should be part of 12.0.0 and the line is there now. Resolved?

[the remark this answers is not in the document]
glasserc: Resolved by @leplatrem in https://github.com/peterbe/kinto/commit/3128691ae1396c796000e88d4ca8d49cd93840d7

kinto/core/storage/__init__.py
leplatrem: yep

docs/api/1.x/records.rst
leplatrem: I saw what you did, perfect!

docs/api/1.x/_details-head-list.rst
```diff
@@ -0,0 +1,10 @@
+Unlike GET requests, HEAD requests contain an additional header called ``Total-Objects``
+(and ``Total-Records`` for backwards compatibility) which is a count of all objects
+in the collection with the current filtering.
```
leplatrem: in the *resource*?

peterbe:
> I let you rebase/squash/do whatever you want before pressing the merge button

I don't think there are any commits that are worthy to have to be squashed. If I do try to selectively squash some of them, I will have to go through the same merge conflicts you went through, no?

leplatrem:
Yeah, I'm ok with whatever ;)

PR #1932, opened 2018-12-10
Remove UserResource

leplatrem:
- It is unused
- It simplifies several layers of the resource code
- It can still be implemented with a normal resource if needed

- [x] Update code
- [x] Update docs
- [x] Green tests
- [x] ~~Document upgrade path~~ Deprecate stuff

### DIFF
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -105,7 +105,11 @@ class AuthorizationPolicy:
 
     def _get_bound_permissions(self, object_id, permission):
         if self.get_bound_permissions is None:
-            return [(object_id, permission)]
+            # Permission to 'write' gives permission to 'read'.
+            bound = [(object_id, permission)]
+            if permission == "read":
+                bound += [(object_id, "write")]
+            return bound
         return self.get_bound_permissions(object_id, permission)
 
     def principals_allowed_by_permission(self, context, permission):
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -2,7 +2,7 @@ import functools
 import logging
 
 from pyramid.settings import aslist
-from pyramid.security import IAuthorizationPolicy, Authenticated
+from pyramid.security import IAuthorizationPolicy
 from zope.interface import implementer
 
 from kinto.core import utils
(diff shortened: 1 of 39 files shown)


Natim:
I guess the history was the readinglist project where sharing wasn't a feature.
I believe this is the way to go.

kinto/core/authorization.py
```diff
@@ -12,11 +12,9 @@
 logger = logging.getLogger(__name__)
 
 # A permission is called "dynamic" when it's computed at request time.
+# XXX: this is the only one we have, is that necessary?
 DYNAMIC = "dynamic"
 
-# When permission is set to "private", only the current user is allowed.
-PRIVATE = "private"
```
Natim: We do use PRIVATE :dancer:

Natim: `
./kinto_chefclub/views/devices.py:
    6 : from kinto.core.authorization import PRIVATE
   51 : @devices.delete(permission=PRIVATE)
   68 :     permission=PRIVATE, schema=PayloadSchema(), validators=(colander_body_validator,)
`

kinto/core/authorization.py
```diff
@@ -53,15 +51,11 @@ class AuthorizationPolicy:
     permission depend on others."""
 
     def permits(self, context, principals, permission):
-        if permission == PRIVATE:
-            return Authenticated in principals
```
Natim: Why would you remove this?

leplatrem:
> Could you elaborate on the changed behavior of PATCH operations that don't change anything? Is this a changed API behavior? What was the justification for the old behavior?

In user resource, when a PATCH didn't change anything we wouldn't hit storage. In shareable resources (ie. whole kinto), it does. 
If I remember correctly, in ReadingList we had some code that wouldn't change the existing «reading position» if it was inferior to the current one or something like that.

> We do use PRIVATE :dancer: 

I guess that returning `self.request.get_prefixed_userid()` in the `get_parent_id()` method should be enough.

We can keep the PRIVATE permission and bypass the permission backend entirely.

The only different with the old code is that the responses payloads will contain a `permissions` field.

leplatrem:
@glasserc @Natim r?

leplatrem:
> I'm (still....) sad about the loss of the docstrings which (in my mind) communicate a bit about what Resource is "for" and how you might use it in a plugin.

Sorry I didn't get back to that, true, I will then.

I was very surprised by the amount of docs we have for kinto.core though :D

PR #1941, opened 2018-12-19
optimize use of jsonschema.validate()

### LINKED ISSUE #1807: Optimize use of jsonschema.validate()
It could be interesting to investigate if the fastjsonschema lib would give us better perfs :)

I marked this as easy-pick since everything happens in `schema_validation.py`

https://github.com/horejsek/python-fastjsonschema/

peterbe:
Fixes #1807

- ~~[ ] Add documentation.~~
- ~~[ ] Add tests.~~
- [ ] Add a changelog entry.
- ~~[ ] Add your name in the contributors file.~~
- ~~[ ] If you changed the HTTP API, update the [API_VERSION](https://github.com/Kinto/kinto/blob/master/kinto/__init__.py#L15) constant and add an API changelog entry [in the docs](https://github.com/Kinto/kinto/blob/master/docs/api/index.rst)~~
- ~~[ ] If you added a new configuration setting, update the [`kinto.tpl`](https://github.com/Kinto/kinto/blob/master/kinto/config/kinto.tpl) file with it.~~

### DIFF
--- a/kinto/schema_validation.py
+++ b/kinto/schema_validation.py
@@ -1,5 +1,6 @@
 import colander
-from jsonschema import ValidationError, SchemaError, RefResolutionError, validate
+from jsonschema import ValidationError, SchemaError, RefResolutionError
+from jsonschema.validators import validator_for
 
 try:  # pragma: no cover
     from jsonschema import Draft7Validator as DraftValidator
@@ -37,6 +38,34 @@ def check_schema(data):
     except SchemaError as e:
         message = e.path.pop() + e.message
         raise ValidationError(message)
+
+
+# Module level global that stores a version of every possible schema (as a <class 'dict'>)
+# turned into a jsonschema instance (as <class 'jsonschema.validators.Validator'>).
+_schema_cache = {}
+
+
+def validate(data, schema):
+    """Raise a ValidationError or a RefResolutionError if the data doesn't validate
+    with the given schema.
+
+    Note that this function is just a "wrapper" on `jsonschema.validate()` but with
+    some memoization based on the schema for better repeat performance.
+    """
+    # Because the schema is a dict, it can't be used as a hash key so it needs to be
+    # "transformed" to something that is hashable. The quickest solution is to convert
+    # it to a string.
+    # Note that the order of the dict will determine the string it becomes. The solution
+    # to that would a canonical serializer like `json.dumps(..., sort_keys=True)` but it's
+    # overkill since the assumption is that the schema is very unlikely to be exactly
+    # the same but different order.
+    cache_key = str(schema)
+    if cache_key not in _schema_cache:
+        # This is essentially what the `jsonschema.validate()` shortcut function does.
+        cls = validator_for(schema)
+        cls.check_schema(schema)
+        _schema_cache[cache_key] = cls(schema)
+    return _schema_cache[cache_key].validate(data)
 
 
 def validate_schema(data, schema, ignore_fields=[]):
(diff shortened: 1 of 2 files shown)


peterbe:
What do you think? I'm confident this solution is faster but I'm not sure I can see the whole picture to judge if it's worth it. 

**Here's how I benchmarked it**

I put in a `print([data, schema]); raise Exception('break!')` right befor the call to `validate()` and collected all the various data and the schema dicts that get used in the tests. Then I put this into a little [test file](https://gist.github.com/peterbe/dd2146f74da7613d8e23f175d4e084ff) which runs some a bunch of these calls a total of 10,000 times. There are only 4 unique schemas. Anyway, instead of taking 3.46 seconds it now does it in 0.18 seconds. So for this particular scenario of repeating the schema it's a 20x improvement. 

In https://www.peterbe.com/plog/jsonschema-validate-10x-faster-in-python I discovered that if you make the instance and reuse it repeatedly it's a 10x improvement. In that benchmark/blog post the data blob was large and the caching of the schema was *once*. 

**The big question is, are the total number of distinct schemas, during the lifetime of a started Kinto server, so many that that `_schema_cache` module level global would build up too much memory.**

peterbe:
> uwsgi will renew the worker automatically

really?! why would it do that? Is there like a timer that says how long the process can stay alive? Does it depend on the threading option?

Either way, suppose you have 1,000 different schemas in a Kinto instance then you have 1,000 dicts and 1,000 dicts-as-strings. [The Buildhub schema](https://github.com/mozilla-services/buildhub/blob/d2b89d66c614abce9cf370a6402edd77009d8887/jobs/buildhub/initialization.yml#L71-L198) is 2,456 bytes as a string and 368 bytes as a dict and 1,056 as `Validator` instances. If there were 1,000 of these that are all different, it would mean 1,000 * 2,456 + 1,000 * 1,056 ~= 3.5MB in RAM. 

Basically, 1,000 different schemas cached would be 3.5MB *per Python process/Gunicorn worker*. That's extremely little memory for a production system running on state-of-the-art hardware. Also, 1,000 different schemas is probably an exaggeration.

Natim:
> really?!

If the memory goes wild the harakiri plugin can do that yes.

kinto/schema_validation.py
```diff
@@ -39,6 +40,34 @@ def check_schema(data):
...
+    Note that this function is just a "wrapper" on `jsonschema.validate()` but with
+    some memoization based on the schema for better repeat performance.
+    """
+    # Because the schema is a dict, it can't be used as a hash key so it needs to be
+    # "transformed" to something that is hashable. The quickest solution is to convert
+    # it to a string.
+    # Note that the order of the dict will determine the string it becomes. The solution
+    # to that would a canonical serializer like `json.dumps(..., sort_keys=True)` but it's
+    # overkill since the assumption is that the schema is very unlikely to be exactly
+    # the same but different order.
```
leplatrem: In Python 3.6 dict keys are stable right?

peterbe: > In Python 3.6 dict keys are stable right?

Yes but not guaranteed. Unlike 3.7.

peterbe:
@leplatrem 

> and maybe a test?

Do you have some suggestions as to how to do that?
There's no `test_schema_validation.py` or anything like that. The only tests that execute this code are `tests/test_views_schema_*.py` and they don't really care about implementation details within.

peterbe:
Changelog entry added. Mergy merge?

leplatrem:
> ...and they don't really care about implementation details within.

You're right, let's merge!

PR #1951, opened 2019-01-02
Validate record ID (fixes #1942)

### LINKED ISSUE #1942: String pattern in ID field not taken into account
**Steps to reproduce**

- Create a collection with a JSON schema:
```json
"schema": {
  "properties": {
    "id": {
      "type": "string",
      "pattern": "^[0-7]$"
    },
    "title": {
      "title": "Title",
      "type": "string",
      "pattern": "^[0-7]$",
      "description": "Short title"
    }
}
``` 

* Post a record `{"id": "abc", "title": "1"}`

**Expected**

* the record should be rejected because `id` field does not match pattern

**Actual behaviour**

* pattern is verified on the title field, but not on `id`

See:

https://github.com/Kinto/kinto/blob/239deea6a3c486187e7a5dbd264434661c48560b/kinto/views/records.py#L63-L69

leplatrem:
Fixes #1942

### DIFF
--- a/kinto/schema_validation.py
+++ b/kinto/schema_validation.py
@@ -100,7 +100,7 @@ def validate_schema(data, schema, ignore_fields=[]):
         raise e
 
 
-def validate_from_bucket_schema_or_400(data, resource_name, request, ignore_fields=[]):
+def validate_from_bucket_schema_or_400(data, resource_name, request, id_field, ignore_fields=[]):
     """Lookup in the parent objects if a schema was defined for this resource.
 
     If the schema validation feature is enabled, if a schema is/are defined, and if the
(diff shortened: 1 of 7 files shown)


leplatrem:
I had to complexify a bit the original patch, in order to have consistency WRT schemas defined at bucket level (`record:schema`, `collection:schema` ...etc)

May deserve a second look.

kinto/schema_validation.py
```diff
@@ -130,6 +130,8 @@ def validate_from_bucket_schema_or_400(data, resource_name, request, ignore_fiel
 
     # Validate or fail with 400.
     schema = bucket[metadata_field]
+    if id_field not in schema.get("properties", {}):
+        ignore_fields += (id_field,)
```
glasserc: Maybe we can move this logic into `validate_schema` and not have to duplicate it in `kinto/views/records.py`.

PR #1966, opened 2019-01-09
Remove delete collection

### LINKED ISSUE #1959: Remove the delete-collection command
This command was introduced to fix a very specific use-case a long time ago where we had created a collection with an invalid ID. It could thus not be deleted via the HTTP API. 

I think we can remove it.

(from #1953)

Bjdufre1:
Fixes #1959 
Removes the delete collection command

- [X] Add documentation.
- [X] Add tests.
- [X] Add a changelog entry.
- [X] If you changed the HTTP API, update the [API_VERSION](https://github.com/Kinto/kinto/blob/master/kinto/__init__.py#L15) constant and add an API changelog entry [in the docs](https://github.com/Kinto/kinto/blob/master/docs/api/index.rst)
- [X] If you added a new configuration setting, update the [`kinto.tpl`](https://github.com/Kinto/kinto/blob/master/kinto/config/kinto.tpl) file with it.


CHANGELOG.rst
```diff
@@ -46,7 +46,7 @@ This document describes changes between each past release.
 - Changed psycopg2 dependency for psycopg2-binary. (#1905)
 - Renamed core notions (ie. record and collection) (#710)
 - JSON Schema validation is optimized by keeping instances of validator cached. (#1807)
-
+- Remove delete-collection command (#1959)
```
leplatrem: nitpick: I wouldn't list this change in the *internal changes* category. Probably *breaking changes*...

tests/openapi/test_responses_collections.py
```diff
@@ -47,23 +47,9 @@ def test_put_collection_201(self):
...
         op = self.resources["Collections"].get_collections
         schema = self.spec.deref(op.op_spec["responses"]["200"])
         validate_response(schema, op, response)
-
-    def test_delete_collections_200(self):
-        response = self.app.delete("/buckets/b1/collections", headers=self.headers, status=200)
-        response = self.cast_bravado_response(response)
-        op = self.resources["Collections"].delete_collections
-        schema = self.spec.deref(op.op_spec["responses"]["200"])
-        validate_response(schema, op, response)
```
leplatrem: This tests should be kept, since they don't relate to the CLI but the HTTP API

tests/openapi/test_responses_errors.py
```diff
@@ -308,14 +301,6 @@ def test_list_delete_401(self):
...
         validate_response(schema, op, response)
 
-    def test_lidt_delete_403(self):
-        headers = {**self.headers, **testing.get_user_headers("aaa")}
-        response = self.app.delete("/buckets/b1/collections", headers=headers, status=403)
-        response = self.cast_bravado_response(response)
-        op = self.resources["Collections"].delete_collections
-        schema = self.spec.deref(op.op_spec["responses"]["403"])
-        validate_response(schema, op, response)
-
```
leplatrem: Same here, this is about the API, not the CLI

PR #1943, opened 2018-12-20
Add cache flush command (fixes #1785)

Pac23:
Fixes #1785

- [x] Add documentation.
- [x] Add tests.
- [ ] Add a changelog entry.
- [ ] Add your name in the contributors file.

this is supposed to fix #1785 , still in the works, opened for maintainer reference.

PR #1967, opened 2019-01-10
Preparing release 12.0.0

leplatrem:
I think we can now release this big chunk of changes!

(We can wait for #1943, depending how far it is from being merged.)


Pac23:
@leplatrem that has been tested, works documentation is remaining.

PR #1982, opened 2019-01-18
Add account validation capability to the accounts plugin

### LINKED ISSUE #1973: Thoughts about an "accounts registration" plugin
Following some discussions, here's my rough thoughts on creating an "accounts registration" plugin. The use case is having a flow for users to
1/ register an account
2/ receive a "registration code" via email
2/ validate their email using this "registration code"
3/ reset their password if needed

## Register an account

1/ anonymous POST on `/accountsmgmt/register/<email>`
2/ this creates a record `{"id": email, "password": hash, "registration-code": uuid}`
3/ an email is sent with a link to `/accountsmgmt/register/validate/<registration-code>`
4/ GETing `/accountsmgmt/register/validate/<registration-code>` returns the ID which is the username (the email)
5/ POSTing to `/accountsmgmt/register/validate/<registration-code>` creates a kinto account with the same id and password, and updates the current "register" record to mark it as used (or deletes it)

## Resetting a password
1/ anonymous POST on `acountsmgmt/resetpassword/<email>`
2/ this creates a record `{"id": email, "reset-code": uuid}`
3/ an email is sent with a link to `/accountsmgmt/resetpassword/reset/<reset-code>`
4/ GETing `/accountsmgmt/resetpassword/<reset-code>` returns the ID which is the username (the email)
5/ POSTing the new password to `/accountsmgmt/resetpassword/reset/<reset-code>` updates the kinto account with the same ID, and updates the current "register" record to mark it as used (or deletes it)

Using the email (username) as the ID as the advantage of not having several registration or password reset codes laying around for the same user. The drawback is that we're not benefiting from the "kinto resource" management as we're GET/POST-ing on endpoints that are the registration or password reset codes, and not the IDs.

Not sure if that makes sense? Do you have any feedback, ideas or tips?

magopian:
Fixes #1973

**Try it here: https://kinto.github.io/kinto-account-demo/**

- [x] Register a new account
- [x] Send an email on account creation with a link to the activation form with the activation key
- [x] Validate (activate) an account
- [x] Reset a password
- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.
- [x] Only load validate and reset-password endpoints if the email_validation feature is activated [More info](https://github.com/Kinto/kinto-fxa/blob/master/kinto_fxa/__init__.py#L63-L68)
- [x] Update the capabilities to add a boolean info in the accounts feature rather than having multiple capabilities [More info](https://github.com/Kinto/kinto/pull/1982/#discussion_r251736595)
- [ ] Add a heartbeat that would try to connect to the SMTP service to validate that we can send emails, it would default to true in debug mode. [See pyramld_mailer related discussion](https://github.com/Pylons/pyramid_mailer/issues/91)
- [x] Move the mail sending to a listener/dedicated functions

This is a work in progress, may I have some feedback on the current work before pushing forward with the reset password functionality and the documentation?

Thanks!


tests/plugins/test_accounts.py
```diff
@@ -158,6 +180,170 @@ def test_authentication_refresh_the_cache_each_time_we_authenticate(self):
...
+                status=201,
+            )
+        self.app.post_json("/accounts/alice@example.com/validate/" + uuid_string, {}, status=200)
+        resp = self.app.patch_json(
+            "/accounts/alice@example.com",
+            {"data": {"some-other-metadata": "foobar"}},
+            status=200,
+            headers=get_user_headers("alice@example.com", "12éé6"),
+        )
+        print(resp)
```
n1k0: Oh!

magopian: errr... those are not the debug messages you're looking for

n1k0: Ah.

kinto/plugins/accounts/views.py
```diff
@@ -43,13 +49,21 @@ class Account(resource.Resource):
...
         context.is_administrator = (
             len(set(aslist(allowed_from_settings)) & set(request.prefixed_principals)) > 0
         )
         # Shortcut to check if current is anonymous (before get_parent_id()).
         context.is_anonymous = Authenticated not in request.effective_principals
+        # Is the "accounts validation" setting set?
+        context.validation = settings.get("account_validation", False)
+        # Account validation requires the user id to be an email.
+        validation_email_regexp = settings.get(
+            "account_validation.email_regexp", "^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$"
```
almet: Just a question: what happens in case I want to update my prez-account-validation existing account? Will my userid be rejected?

kinto/plugins/accounts/views.py
```diff
@@ -103,16 +130,51 @@ def process_object(self, new, old=None):
...
+            if "activation-form-url" not in new:
+                error_details = {
+                    "name": "data.activation-form-url",
+                    "description": "Account validation is enabled, and requires an "
+                    "`activation-form-url` field to be set",
+                }
+                raise_invalid(self.request, **error_details)
+
+            activation_key = str(uuid.uuid4())
+            new["activation-key"] = activation_key
```
magopian: Haha, I just realized while taking my shower: storing this activation key in the account means the user will get it back as a response to the POST. And will thus be able to POST to the `validate` endpoint without needing the email. Which defeats the whole purpose of validating the email.

So this `activation-key` has to be used straight away while sending the email, but not stored!

Also, I want to be paid overtime.

Natim: Storing it in the cache backend seems to be a good idea in that case.

magopian: Or we could generate the activation-key by hashing the record with the server's secret key, which means it could be re-generated when needed (when checking the validation key on the `validate` endpoint, or if the activation link needs to be sent again).

kinto/plugins/accounts/views.py
magopian: I'll add a test to make sure an account created before enabling the account validation can still authenticate.

kinto/plugins/accounts/views.py
Natim: > Or we could generate the activation-key by hashing the record with the server's secret key

To make this correctly, the server secret key should be rotqted for each user which means you would still have to store the hash in the cache.

The benefit of using the cache is that you can handle expiration seamlessly.

magopian: What i see from the code in `kinto/plugins/accounts/authentication.py` is that we're using the server secret to get the `cache_key`. Could we use the exact same technique with the user record (and not only the username) to generate the activation key?

I don't understand why we would have to rotate the secret key? Is it because the cache key is not accessible to the end user, but the activation key is, and would thus allow attackers to get enough activation keys to infer the secret key?

magopian: I've implemented your suggestion @Natim storing the activation key in the cache, let me know what you think ;)

https://github.com/Kinto/kinto/pull/1982/commits/9dd642ebb038497215dbc0dce7f499aae604d14a

magopian:
> At first, I had trouble to understand what this activation-form-url was made for, but from what I understand, it's purpose is to add a link in the validation email that is sent to the user.

@almet the idea (discussed in #1973) is that modifying the user record should not be done with a GET, but with a POST (or PATCH). So the `activation-form-url` is the link to the WebApp, which should display a proper form asking the user to confirm that they want to enable/activate/validate the account (probably a simple "call to action" button that will POST to the `validate` endpoint).

The WebApp should do this POST asynchronously, in the background (everything kinto related should be done this way in a WebApp I believe), so I'm not sure we need a redirection url.

I'm not arguing against the GET and redirection url though ;)

kinto/plugins/accounts/views.py
Natim: Yep seems good.

magopian:
> Why displaying a form if the user clicks on the link already? The Webapp could do the POST directly, saving an extra click, and staying RESTful. If we decide to go this way, we could have a default webapp shipped direclty with kinto-accounts, which will do the POST on behalf of the user, and letting the ability to the developer to specify a different webapp to use.

If we're comfortable with modifying a record without a user explicit action (auto-POSTing a form), I'd be ok with going all in and changing the `validate` endpoint to accept a simple GET.

If we do that, then there's no more need for an `activation-form-url`.

In any case, the `redirect-url` can already be set by modifying the link in the `email_body_template` setting. We could also accept it as a GET or POST parameter on the `validate` endpoint.

Maybe we should discuss this in another issue/PR?

magopian:
Ok, the current work on the user creation and validation should be done now, and properly documented.

What's left to discuss before this can be merged:
1/ do we want to get rid of the `activation-form-url` in favor of validating a user using a simple GET? (see https://github.com/Kinto/kinto/pull/1982#issuecomment-456316544)
2/ does it make sense to merge this PR as is, without the "reset password" functionality, which could be discussed in another issue, and implemented in another PR?

magopian:
After an out-of-band discussion with @Natim we came to the conclusion that this PR could be slightly modified:

1/ don't require an `activation-form-url` field
2/ allow the user creator to provide some email context when POSTing on the `/accounts/` endpoint (that will be used in addition to the user record and the activation-key to render the email template from the settings)
3/ to validate the user, still require a POST on the `/accounts/<user id>/validate/<activation key>` endpoint, through whatever means the user creator deems fit (it could be through a form, at a URL that they have provided in the email context in 2/). Here also allow the user creator to provide some email context (to render the "user successfully activated" email in 4/)
4/ when the user is validated, send another mail to confirm its activation

almet:
Not sure what you mean with « email context ». You mean a template?

Also, In addition to what you've described, I think that *if* we go with a POST to validate the account, then we should provide a simple WebApp that does the POSTing automatically, in order to have a easy-to-setup account registration mechanism (e.g. as an admin deploying the stack, I want this to be simple enough)

Natim:
> Some softwares have UI that are not browsers

In that case they might not use links to validate the account.

almet:
> > Some softwares have UI that are not browsers
> 
> In that case they might not use links to validate the account.

I think I missed something then. To me, if I register an account with the command line, the workflow is as follows:

1. A request is made from the CLI to the kinto server to create the account
2. An email is sent to the user, to confirm the email belongs to her / him
**3. The user clicks on the link to validate**
4. Account is validated.

Unfortunately, users use links in the emails to validate the accounts, and this happens through HTTP GET requests only.

What other scenario do you have in mind to avoid this webapp?

Natim:
>  1. A request is made from the CLI to the kinto server to create the account
>  2. An email is sent to the user, to confirm the email belongs to her / him
>  3. **The user clicks on the link to validate**
>  4. Account is validated.

This is the case when working with a webapp, but if you are using a software you might want to use the following:

1. A request is made from the CLI to the kinto server to create the account
2. An email is sent to the user, to confirm the email belongs to her / him
3. **The user copy the activation key and use the CLI to validate the account.**
4. Account is validated.

almet:
Ah yes! I haven't thought about this solution. I remember that's what Twitter used to do in the past :-)

That being said, I tend to prefer the « user clicks on the link » solution, because it's simpler. What's the value of having the user copy a verification code over clicking a link in an email, apart not having to implement an UI?

Natim:
I think it is up to the developper to define the flow, I am not against building a little Elm demo webapp if it pleases you. We can even pair on it once this lands.

almet:
> I think it is up to the developper to define the flow, I am not against building a little Elm demo webapp if it pleases you. We can even pair on it once this lands.

I agree, we need to let the possibility for the developer to tweak the flow, **if they want it to be different**. One major selling point of kinto is « don't reinvent the wheel each time », and as such, I believe we should provide the golden path for everything. Something that makes it easy for everyone to use accounts, without having to implement something special.

Natim:
I agree with you in the scope of Kinto which is an API. As soon as you have front-end facing experience, it is not the place of Kinto to solve it.

almet:
I agree, **but** I believe that if we start solving one problem (here having accounts validation), then we need to do it completely, and painlessly for the devs that want to use it.

As a side note, I think we're kinda going in circles with this discussion. I don't really mind disagreeing, but if anyone wants to weight in in order to move forward on this topic that would be beneficial :-) cc @leplatrem @glasserc

glasserc:
I'm not sure you really want to hear my opinion, so I've been trying to stay out of it, but since you asked: I don't think Kinto is a good foundation for the kinds of thing you want to do. I don't like the current Kinto accounts plugin and I would prefer Kinto core not be cluttered with support for all the different possible things some developer might want an accounts system to do.

I think the PostgREST approach here is pretty solid: don't do any user stuff whatsoever -- delegate that to some other service -- but allow people to present JWTs to prove their identities. I recently saw https://github.com/netlify/gotrue which seems like a pretty nice product, but if you wanted to build something open source I would certainly support that too.

Natim:
> I would prefer Kinto core not be cluttered with support for all the different possible things some developer might want an accounts system to do.

I agree with you and I guess it is why we decided to put the account code in a plugin.

As of why plugins ships with Kinto itself, this is another story.

magopian:
Wow, such long thread.

Here's a plan:
1/ I'll continue working on this PR to implement the modifications suggested in https://github.com/Kinto/kinto/pull/1982#issuecomment-456539599
2/ I'll implement the reset password feature

Once this is done, we can decide if it's good enough to land.

Regarding the "sample webapp that kinto should provide": let's discuss this in another issue, I'd like to not delay the landing of this functionality for too long. We can still improve in the future if we want (eg: also allowing a GET to validate an endpoint, providing redirect URL, providing a sample webapp, ...).

magopian:
I believe the PR as is is good to review on the user creation and validation side. I'm now going to implement the "reset password" functionality.

magopian:
I believe the PR as is is good to review on the user creation and validation side. I'm now going to implement the "reset password" functionality.

magopian:
I believe this is now complete (with the password reset) and ready to review.

kinto/plugins/accounts/views.py
```diff
@@ -138,3 +223,192 @@ def on_account_changed(event):
...
     # Delete cache
     cache.delete(cache_key)
+
+
+# Account validation (enable in the settings).
+validation = Service(
+    name="account-validation",
+    path="/accounts/{user_id}/validate/{activation_key}",
+    description="Validate an account",
+)
```
magopian: Quick note here: the `validation` and the `reset-password` endpoint (services) are available, even if the account validation option isn't enabled in the settings. However, they would not work properly. I'm not sure how to conditionally add a Service?

Natim: Yep that would be great: https://github.com/Kinto/kinto-fxa/blob/master/kinto_fxa/__init__.py#L65-L68

Natim:
Second issue is that the accounts doesn't seems to have CORS activated?

Natim:
I made a little UI that you can try here: https://kinto.github.io/kinto-account-demo/index.html

magopian:
> Ok the creation doesn't have a CORS issue but the validation has got one.

I'm not sure why this would happen, do you have an idea? Is there something specific to do when adding the service?

> My first issue was that I had a smtplib connection error.

I added a commit with some documentation on the configuration of `pyramid_mailer`.

> Is there a way to add it to the heartbeat part to make sure emailing is correctly configured?

I'll have a look at how to do that

almet:
I've tried with the demo @natim put together and it worked for me (wrt. CORS). Any way to reproduce the error?

kinto/plugins/accounts/__init__.py
```diff
@@ -29,6 +41,20 @@ def includeme(config):
 
     settings = config.get_settings()
 
+    if settings.get("account_validation", False):
+        config.add_api_capability(
+            "account-validation",
```
Natim: @leplatrem what do you think about this?

I my opinion, we should restrict to one capability per plugin and instead add other info in the capability of the plugin.

```
accounts": {

    "description": "Manage user accounts.",
    "url": "https://kinto.readthedocs.io/en/latest/api/1.x/accounts.html",
    "id-validation": "^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$",
    "reset-password-enable": True

}
```

kinto/plugins/accounts/__init__.py
```diff
@@ -29,6 +41,20 @@ def includeme(config):
```
Natim: Please add conditional view loading for reset password and validation.

Natim:
I add the reset-password and change-password screen on the kinto-account-demo UI.
It seems that the CORS OPTIONS request for the `/account/<user-id>` returns a 400.

kinto/plugins/accounts/views.py
```diff
@@ -138,3 +223,192 @@ def on_account_changed(event):
...
+            parent_id=parent_id, resource_name="account", object_id=user_id
+        )
+    except storage_exceptions.ObjectNotFoundError:
+        # Don't give information on the existence of a user id: return a generic error message.
+        error_details = {"message": "Account ID and activation key do not match"}
+        raise http_error(httpexceptions.HTTPForbidden(), **error_details)
+
+    if is_validated(user):
+        error_details = {"message": f"Account {user_id} has already been validated"}
+        raise http_error(httpexceptions.HTTPForbidden(), **error_details)
```
leplatrem: Maybe we shouldn't care about this. It allows me to know that user X is validated or not

kinto/plugins/accounts/views.py
```diff
@@ -138,3 +223,192 @@ def on_account_changed(event):
...
+
+    if not check_validation_key(activation_key, user_id, request.registry):
+        error_details = {"message": "Account ID and activation key do not match"}
+        raise http_error(httpexceptions.HTTPForbidden(), **error_details)
+
+    # User is now validated.
+    user["validated"] = True
+
+    request.registry.storage.update(
+        parent_id=parent_id, resource_name="account", object_id=user_id, record=user
```
leplatrem: I'm pretty sure the kwarg is `object` and not `record`

kinto/plugins/accounts/views.py
```diff
@@ -138,3 +223,192 @@ def on_account_changed(event):
...
+    )
+    formatted_body = formatter.format(email_confirmation_body_template, **user, **email_context)
+
+    message = Message(
+        subject=formatted_subject,
+        sender=email_sender,
+        recipients=[user_email],
+        body=formatted_body,
+    )
+    mailer.send(message)
```
leplatrem: I would suggest to send an event and move all this code to a listener instead (or at least a function)

kinto/plugins/accounts/views.py
```diff
@@ -138,3 +223,192 @@ def on_account_changed(event):
...
+
+
+def cache_reset_password(reset_password, username, registry):
+    """Store a reset-password in the cache."""
+    settings = registry.settings
+    hmac_secret = settings["userid_hmac_secret"]
+    cache_key = utils.hmac_digest(hmac_secret, ACCOUNT_RESET_PASSWORD_CACHE_KEY.format(username))
+    # Store a reset password for 7 days by default.
+    cache_ttl = int(
+        settings.get("account_validation.reset_password_cache_ttl_seconds", 7 * 24 * 60 * 60)
```
leplatrem: ditto

kinto/plugins/accounts/views.py
```diff
@@ -138,3 +223,192 @@ def on_account_changed(event):
...
+        )
+    except storage_exceptions.ObjectNotFoundError:
+        # Don't give information on the existence of a user id: return a generic message.
+        return {"message": "A temporary reset password has been sent by mail"}
+
+    settings = request.registry.settings
+
+    user_email = user["id"]
+    email_regexp = settings.get(
+        "account_validation.email_regexp", "^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$"
```
leplatrem: constant...

kinto/plugins/accounts/views.py
```diff
@@ -138,3 +223,192 @@ def on_account_changed(event):
...
+        email_reset_password_body_template, **user, **extra_data, **email_context
+    )
+
+    message = Message(
+        subject=formatted_subject,
+        sender=email_sender,
+        recipients=[user_email],
+        body=formatted_body,
+    )
+    mailer.send(message)
```
leplatrem: I would prefer the email stuff to be confined in one place instead of being spread out in views...

kinto/plugins/accounts/views.py
almet: +1 for having a listener for this :-)

docs/api/1.x/accounts.rst
```diff
@@ -294,3 +294,166 @@ Or delete some account:
...
+==============================
+
+If the ``account validation`` option in :ref:`the settings
+<settings-accounts>` has been enabled, a temporary reset password may be
+requested through the endpoint is available at
+`/accounts/(user id)/reset-password`.
+
+.. sourcecode:: bash
+
+    $ http POST http://localhost:8888/v1/accounts/bob@example.com/reset-password --verbose
```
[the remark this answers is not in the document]
magopian: Indeed you can. What do you recommend? Adding a timer to only resend a mail (and overwrite the reset password) every 30 minutes or something?

docs/api/1.x/accounts.rst
```diff
@@ -294,3 +294,166 @@ Or delete some account:
...
+addresses: they need to match the regexp in the
+``account_validation.email_regexp`` setting.
+
+.. sourcecode:: bash
+
+    $ echo '{"data": {"id": "bob@example.com", "password": "azerty123"}}' | http POST http://localhost:8888/v1/accounts --verbose
+
+If the user was created, an email will be sent to the user with a link to a
+form with the activation key. This form should POST this activation key to the
+``validate`` endpoint.
```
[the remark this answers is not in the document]
Natim: To be restful, we want this step to be a POST, this would also prevent developer to point their user to a JSON API endpoint rather than to an interface.

kinto/plugins/accounts/__init__.py
magopian: I'll add a boolean for `validation-enabled` (I don't think it's relevant to have a `reset-password-enabled` boolean as it's tied to the account validation).

docs/api/1.x/accounts.rst
```diff
@@ -294,3 +294,166 @@ Or delete some account:
...
+If the ``account validation`` option in :ref:`the settings
+<settings-accounts>` has been enabled, account IDs need to be valid email
+addresses: they need to match the regexp in the
+``account_validation.email_regexp`` setting.
+
+.. sourcecode:: bash
+
+    $ echo '{"data": {"id": "bob@example.com", "password": "azerty123"}}' | http POST http://localhost:8888/v1/accounts --verbose
+
+If the user was created, an email will be sent to the user with a link to a
```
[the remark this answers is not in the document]
magopian: Isn't it already linked a few lines above, at the top of the section ? (line 306 in this diff)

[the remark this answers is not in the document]
magopian: There's a link to the settings section a few lines above (line 306 in this div). In the settings section, I moved the "email configuration" closer to the beginning of the section instead of the very bottom of it.

kinto/plugins/accounts/authentication.py
```diff
@@ -34,13 +41,43 @@ def account_check(username, password, request):
...
         cache.set(cache_key, hashed_password, ttl=cache_ttl)
         return True
 
+    # Last chance: is this a "reset password" flow?
+    cached_password = get_cached_reset_password(username, request.registry)
+    if not cached_password:
+        return None
+
+    # The temporary reset password is only available for changing a user's password.
+    if request.method.lower() not in ["post", "put", "patch"]:
```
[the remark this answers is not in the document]
magopian: @leplatrem where can I find this `resource_name`? Should I somehow use the "private" utility `kinto.authorization._resource_endpoint`?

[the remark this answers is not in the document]
leplatrem: You can try this maybe (?)
```py
resource_name, _ = core_utils.view_lookup(request, request.path)
```
https://github.com/Kinto/kinto/blob/master/kinto/core/utils.py#L441

kinto/plugins/accounts/authentication.py
```diff
@@ -34,13 +41,43 @@ def account_check(username, password, request):
...
+        return None
+
+    try:
+        data = request.json["data"]
+    except ValueError:
+        return None
+
+    if not data or "password" not in data or [key for key in data.keys() if key != "password"]:
```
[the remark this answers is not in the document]
magopian: I wanted to prevent modifying anything else than the password, and also we don't use the content of the password field in this piece of the code.

kinto/plugins/accounts/views.py
magopian: It seems it's `obj` :P

magopian:
Ok folks, I hope I have addressed all your comments and reviews, but one thing: the heartbeat request. I've seen [@natim PR](https://github.com/Pylons/pyramid_mailer/issues/91) and would love that they merge it. In any case, I'd love this PR to land as is, and maybe in the future add the heartbeat.

What do you think @natim @leplatrem @almet? Anything else I'm missing?

kinto/plugins/accounts/__init__.py
magopian: It's done now ;)

kinto/plugins/accounts/__init__.py
```diff
@@ -4,30 +4,51 @@
...
         "read": {"account": ["write", "read"]},
     }
 
-    settings = config.get_settings()
+    if validation_enabled:
+        # Valid mailers other than the default are `debug` and `testing`
+        # according to
+        # https://docs.pylonsproject.org/projects/pyramid_mailer/en/latest/#debugging
+        mailer = settings.get("mail.mailer", "")
+        config.include("pyramid_mailer" + ("." + mailer if mailer else ""))
```
Natim:
```suggestion
        config.include("pyramid_mailer" + (f".{mailer}" if mailer else ""))
```

magopian:
Quick note: it seems the `+` character (which can be present in an email address) is [not allowed in the ID](https://github.com/Kinto/kinto/blob/master/kinto/plugins/accounts/views.py#L33). Maybe we should fix that?

Natim:
> Quick note: it seems the `+` character (which can be present in an email address) is [not allowed in the ID](https://github.com/Kinto/kinto/blob/master/kinto/plugins/accounts/views.py#L33). Maybe we should fix that?

Feel free to add a test for it if you want.

docs/configuration/settings.rst
```diff
@@ -569,6 +569,172 @@ You can set ``account_create_principals`` if you want to limit account creation
...
+    `pyramid_mailer's configuration <https://docs.pylonsproject.org/projects/pyramid_mailer/en/latest/#configuration>`_.
+
+You can restrict the email addresses allowed using the
+``account_validation.email_regexp`` setting, and the delay for which the
+activation key will be valid:
+
+.. code-block:: ini
+
+    # Set the regular expression used to validate a proper email address.
+    kinto.account_validation.email_regexp = "^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+$"
```
Natim: Here you accept `+` character

magopian: that's an "additional restriction", it doesn't take over/replace what's declared in https://github.com/Kinto/kinto/blob/master/kinto/plugins/accounts/views.py#L33 (if that's the root cause of this issue, not sure yet, I'm investigating)

Natim: Maybe shall we use the same regex for both?

magopian: I don't think it's the same use case: in one case (no account validation) you want to allow a few set of characters, in the other case you might want to restrict emails to a given sub-domain.

magopian: I pushed a fix

Natim: > I don't think it's the same use case: in one case (no account validation) you want to allow a few set of characters, in the other case you might want to restrict emails to a given sub-domain.

Yeah makes sense :+1:

docs/api/1.x/accounts.rst
```diff
@@ -294,3 +294,115 @@ Or delete some account:
...
+
+.. _accounts-reset-password:
+
+Resetting a forgotten password
+==============================
+
+If the ``account_validation`` option in :ref:`the settings
+<settings-account-validation>` has been enabled, a temporary reset password may
+be requested through the endpoint available at `/accounts/(user
+id)/reset-password`.
```
leplatrem: This is cool and worth mentioning in the CHANGELOG!

leplatrem: Also double back ticks in RST

docs/api/1.x/accounts.rst
```diff
@@ -294,3 +294,115 @@ Or delete some account:
...
+
+.. http:post:: /accounts/(user_id)/reset-password
+
+    :synopsis: Require a temporary reset password for an account with the ``account_validation`` option enabled.
```
leplatrem: *Require*?  *Request*?

docs/api/index.rst
```diff
@@ -11,6 +11,15 @@ API
 Changelog
 ---------
 
+1.22 (unreleased)
+'''''''''''''''''
+
+- new `/accounts/{user id}/validate/{validation key}` endpoint when the
```
leplatrem: *New* (capitalize)

docs/api/index.rst
```diff
@@ -11,6 +11,15 @@ API
...
+  `account validation` option is enabled for the accounts plugin. See
+  :ref:`account validation <accounts-validate>` (#1973)
+- new `/accounts/{user id}/reset-password` endpoint to request a temporary
```
leplatrem: *New* (capitalize)

docs/configuration/settings.rst
```diff
@@ -569,6 +569,162 @@ You can set ``account_create_principals`` if you want to limit account creation
...
+
+    kinto.account_validation = true
+    # Mail configuration:
+    # Set the sender for the validation email.
+    kinto.account_validation.email_sender = "admin@example.com"
+
+.. note::
+
+    Both the account validation and password reset need a properly configured
+    smtp server.
```
leplatrem: SMTP (upercase)

kinto/plugins/accounts/authentication.py
```diff
@@ -34,13 +41,62 @@ def account_check(username, password, request):
...
+
     hashed = existing["password"].encode(encoding="utf-8")
     pwd_str = password.encode(encoding="utf-8")
     # Check if password is valid (it is a very expensive computation)
     if bcrypt.checkpw(pwd_str, hashed):
         cache.set(cache_key, hashed_password, ttl=cache_ttl)
         return True
 
+    # Last chance, is this a "reset password" flow?
+    reset_password_flow(username, password, request)
```
leplatrem: Why no return value here? Deserves a comment IMO

kinto/plugins/accounts/authentication.py
```diff
@@ -34,13 +41,62 @@ def account_check(username, password, request):
...
+
+    cached_password_str = cached_password.encode(encoding="utf-8")
+    if bcrypt.checkpw(pwd_str, cached_password_str):
+        # Remove the temporary reset password from the cache.
+        reset_password_cache_key = utils.hmac_digest(
+            hmac_secret, ACCOUNT_RESET_PASSWORD_CACHE_KEY.format(username)
+        )
+        cache.delete(reset_password_cache_key)
+        cache.set(cache_key, hashed_password, ttl=cache_ttl)
+        return True
```
leplatrem: This return value seems to never be used. How is that?

kinto/plugins/accounts/authentication.py
```diff
@@ -34,13 +41,62 @@ def account_check(username, password, request):
...
+
+    # Only allow modifying a user account, no other resource.
+    uri = utils.strip_uri_prefix(request.path)
+    resource_name, _ = utils.view_lookup(request, uri)
+    if resource_name != "account":
+        return None
+
+    try:
+        data = request.json["data"]
+    except ValueError:
```
leplatrem: Never KeyError? Deserves a comment IMO

kinto/plugins/accounts/authentication.py
```diff
@@ -34,13 +41,62 @@ def account_check(username, password, request):
...
+        return None
+
+    cached_password_str = cached_password.encode(encoding="utf-8")
+    if bcrypt.checkpw(pwd_str, cached_password_str):
+        # Remove the temporary reset password from the cache.
+        reset_password_cache_key = utils.hmac_digest(
+            hmac_secret, ACCOUNT_RESET_PASSWORD_CACHE_KEY.format(username)
+        )
+        cache.delete(reset_password_cache_key)
+        cache.set(cache_key, hashed_password, ttl=cache_ttl)
```
leplatrem: The same way we have `get_cached_reset_password()`  we should have `delete_reset_password()` so that both uses of `ACCOUNT_RESET_PASSWORD_CACHE_KEY` are close to each other

docs/configuration/settings.rst
```diff
@@ -569,6 +569,162 @@ You can set ``account_create_principals`` if you want to limit account creation
...
+    kinto.account_validation.email_body_template = "Hello {id},\n you can now activate your account using the following link:\n {form-url}{activation-key}"
+
+... and they will be ``String.format``-ted with the content of the user, an
+optional additional ``email-context`` provided alongside the user object, and
+the ``activation-key``:
+
+.. sourcecode:: bash
+
+    $ echo '{"data": {"id": "bob@example.com", "password": "azerty123", "email-context": {"name": "Bob Smith", "form-url": "https://example.com/validate/"}}}' | http POST http://localhost:8888/v1/accounts --verbose
+
```
leplatrem: This docs about `email-context` belong to API docs, *settings* docs are about configuration not API usage :wink:

kinto/plugins/accounts/views/__init__.py
```diff
@@ -27,10 +39,27 @@ def _extract_posted_body_id(request):
         raise http_error(httpexceptions.HTTPUnauthorized(), error=error_msg)
 
 
+def cache_validation_key(activation_key, username, registry):
```
leplatrem: Move to `utils`

kinto/plugins/accounts/views/__init__.py
```diff
@@ -140,3 +197,26 @@ def on_account_changed(event):
...
+        return
+
+    cache = request.registry.cache
+    hmac_secret = settings["userid_hmac_secret"]
+
+    for impacted_object in event.impacted_objects:
+        account = impacted_object["new"]
+        user_email = account["id"]
+        cache_key = utils.hmac_digest(hmac_secret, ACCOUNT_VALIDATION_CACHE_KEY.format(user_email))
+        activation_key = cache.get(cache_key)
```
leplatrem: Move to `utils`? (along other cache manipulations?)

kinto/plugins/accounts/views/validation.py
```diff
@@ -0,0 +1,154 @@
...
+)
+
+
+def check_validation_key(activation_key, username, registry):
+    """Given a username, compare the activation-key provided with the one from the cache."""
+    hmac_secret = registry.settings["userid_hmac_secret"]
+    cache_key = utils.hmac_digest(hmac_secret, ACCOUNT_VALIDATION_CACHE_KEY.format(username))
+
+    cache = registry.cache
+    cache_result = cache.get(cache_key)
```
leplatrem: Ditto about utils and cache read/write/delete :)

kinto/plugins/accounts/views/validation.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Password reset.
+reset_password = Service(
+    name="reset-password",
+    path="/accounts/{user_id}/reset-password",
+    description="Send a temporary reset password by mail for an account",
+)
+
+
+def cache_reset_password(reset_password, username, registry):
```
leplatrem: move to `utils`

tests/test_config.py
```diff
@@ -66,6 +66,7 @@ def test_hmac_secret_is_text(self, mocked_render_template):
 
     @mock.patch("kinto.config.render_template")
     def test_init_postgresql_values(self, mocked_render_template):
+        self.maxDiff = None
```
leplatrem: why this here?

tests/plugins/test_accounts.py
```diff
@@ -157,6 +220,344 @@ def test_authentication_refresh_the_cache_each_time_we_authenticate(self):
...
+        )
+        # Ask for a reset password.
+        self.app.post_json("/accounts/alice@example.com/reset-password", status=404)
+
+
+class AccountValidationCreationTest(AccountsValidationWebTest):
+    def get_account_validation_cache(self, username):
+        hmac_secret = self.app.app.registry.settings["userid_hmac_secret"]
+        cache_key = utils.hmac_digest(hmac_secret, ACCOUNT_VALIDATION_CACHE_KEY.format(username))
+        return self.app.app.registry.cache.get(cache_key)
```
leplatrem: why not use `accounts.utils`?

tests/test_config.py
magopian: I'll remove it, @Natim added it to debug some test failure that needed some more context.

kinto/plugins/accounts/authentication.py
magopian: I believe I forgot to add a `return` value here, good catch ;)

kinto/plugins/accounts/authentication.py
magopian: I have no idea how/why this worked... I'll investigate

magopian: Ok so https://docs.pylonsproject.org/projects/pyramid/en/latest/api/authentication.html#pyramid.authentication.BasicAuthAuthenticationPolicy says that the function is expected to return either `None` or a list of principals. In our case (even before this PR) the function was only return `None` or `True`, and it was working as expected... because it seems everything is based on the cached password (which must be checked elsewhere).

In this PR the `reset_password_flow` value wasn't used/returned, and so the `account_check` function was returning `None`, which worked "correctly" because what really matter is the cache.

What do you think should be done here? I'm going to "fix" the problem by returning the `reset_password_flow` function value even though it's useless in our case, but I guess this should be properly fixed (or not? maybe just commented?) in another issue/PR.

kinto/plugins/accounts/authentication.py
magopian: I'll add the `KeyError`

docs/configuration/settings.rst
magopian: Oh wait, I understood the opposite from one of your previous review, sorry about that will move it back to API

leplatrem: I may have said something confusing. 

I think we can consider: API docs are for web developers who interact with the API, and Settings docs are for admins who install and run the service

tests/test_config.py
Natim: It's not a left over, in case this test fail we will be happy to have it to debug.

kinto/plugins/accounts/authentication.py
leplatrem: Good catch! Don't change the whole thing, you're right it's better to track that in a separate issue

tests/test_config.py
magopian: ok, leaving

magopian:
I believe I've addressed all your comments @leplatrem please let me know if I missed anything (in particular, are you happy with the way the docs are organized now?)

PR #1985, opened 2019-01-21
Fix OpenID connect plugin.

Natim:
Using Google, the `?` is removed which makes it unable to log in.

image

Fixes #

- [ ] Add documentation.
- [x] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/kinto/plugins/openid/views.py
+++ b/kinto/plugins/openid/views.py
@@ -115,7 +115,7 @@ def get_login(request):
 
     # Redirect the client to the Identity Provider that will eventually redirect
     # to the OpenID token endpoint.
-    token_uri = request.route_url("openid_token", provider=provider) + "?"
+    token_uri = request.route_url("openid_token", provider=provider)
     params = dict(
         client_id=client_id, response_type="code", scope=scope, redirect_uri=token_uri, state=state
     )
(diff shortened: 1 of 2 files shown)


kinto/plugins/openid/views.py
```diff
@@ -115,7 +115,7 @@ def get_login(request):
 
     # Redirect the client to the Identity Provider that will eventually redirect
     # to the OpenID token endpoint.
-    token_uri = request.route_url("openid_token", provider=provider) + "?"
+    token_uri = request.route_url("openid_token", provider=provider)
```
leplatrem: Weird, I'm pretty it's here for a reason...  I'm afraid that if we remove it we break Auth0 or something...

Isn't there something you could configure in the Google settings?

Natim: > Isn't there something you could configure in the Google settings?

Unfortunately nope.

Natim: I can also add a settings, `add_trailing_question_mark` or detect that it is Google and remove it :confused:

Natim: I am pretty sure Auth0 should add it automatically too.

Natim: Could you try it on your end?

leplatrem: > Could you try it on your end?

Ok I will on Kinto DEV

I know that I got this plugin working with Google ID :)

Natim: > I know that I got this plugin working with Google ID :)

I do remember that, but it seems to me that they made some changes on their interface :confused:

PR #1988, opened 2019-01-22
Encode token in base64 for url safety.

Natim:
Refs: https://github.com/Kinto/kinto-admin/pull/715

Encode the openid payload as base64 in order to make it URL-safe

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,10 +3,13 @@ Changelog
 
 This document describes changes between each past release.
 
-12.1.0 (unreleased)
+13.0.0 (unreleased)
 -------------------
 
+**Breaking changes**
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -6,7 +6,7 @@ This document describes changes between each past release.
 12.1.0 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Update Kinto OpenID plugin to redirect with a base64 JSON encoded token. (#1988)
```
leplatrem: We should add a note that this is a breaking change IMO. At least mention the minimum required version of kinto-admin :)

PR #1993, opened 2019-01-23
Expose OpenID user profile.

### LINKED ISSUE #1989: OpenID authentication plugin is missing a profile view
When validating the token, the plugn is fetching the profile, but there are no endpoint available to fetch this info: https://github.com/Kinto/kinto/blob/master/kinto/plugins/openid/__init__.py#L74

I would be in favor of adding a `/v1/openid/google/profile` view in order to grab the username and the profile_pic of the user.

This information might also be added in the `user` field of the `/v1/` endpoint. What do you think?

Natim:
Fixes #1989

- [X] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/kinto/core/__init__.py
+++ b/kinto/core/__init__.py
@@ -143,7 +143,10 @@ class JsonLogFormatter(dockerflow_logging.JsonLogFormatter):
 
 def get_user_info(request):
     # Default user info (shown in hello view for example).
-    return {"id": request.prefixed_userid, "principals": request.prefixed_principals}
+    user_info = {"id": request.prefixed_userid, "principals": request.prefixed_principals}
+    if hasattr(request, "get_user_profile"):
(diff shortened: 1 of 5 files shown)


kinto/core/__init__.py
```diff
@@ -143,7 +143,10 @@ def __init__(self, fmt=None, datefmt=None, style="%"):
 
 def get_user_info(request):
     # Default user info (shown in hello view for example).
-    return {"id": request.prefixed_userid, "principals": request.prefixed_principals}
+    user_info = {"id": request.prefixed_userid, "principals": request.prefixed_principals}
+    if hasattr(request, "get_user_profile"):
```
leplatrem: Maybe add a comment that this comes from the openid plugin

Natim: I don't think it is only coming from the openid plugin, it is something any plugin can define.

PR #1996, opened 2019-01-24
Get rid of pytest-cover in favor of pytest-cov

leplatrem:
pytest-cov was already in dev-requiremnents...


glasserc:
Refs https://github.com/Kinto/kinto-signer/pull/262.

PR #1999, opened 2019-01-25
[security] Fix pagination leaking between collections

leplatrem:
Note: kinto < 12 is not affected

Bug introduced in 2fd5cbc15 (#1931)

The SQL generated was missing parenthesis:

screenshot from 2019-01-25 16-54-43

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -420,7 +420,7 @@ class Storage(StorageBase, MigratorMixin):
              WHERE id = :object_id
                AND parent_id = :parent_id
                AND resource_name = :resource_name
-               AND deleted = FALSE
+               AND NOT deleted
             RETURNING as_epoch(last_modified) AS last_modified;
             """
         else:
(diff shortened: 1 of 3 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -420,7 +420,7 @@ def delete(
              WHERE id = :object_id
                AND parent_id = :parent_id
                AND resource_name = :resource_name
-               AND deleted = FALSE
+               AND NOT deleted
```
peterbe: Is this just cleaning up whilst-you're-at-it?

Also, why is it not using `core.storage.DEFAULT_DELETED_FIELD`?

leplatrem: > Is this just cleaning up whilst-you're-at-it?

yes

> Also, why is it not using core.storage.DEFAULT_DELETED_FIELD?

hmm, good point, but the column is hard coded in the schema.sql file anyway

glasserc:
Ouch. Thanks for jumping on this.

PR #2004, opened 2019-01-28
Loosen up CSP in admin plugin (fixes #2000)

### LINKED ISSUE #2000: CSP error while trying to load kinto-admin 1.23 from /v1/admin/
https://catalog.chefclub.tools/v1/admin/

image

image

leplatrem:
Fixes #2000 

This is how I fixed it.

There might be better options using nounces or hashes, but since we use `react-scripts` I could not find anything obvious.
If we consider this as a temporary solution, we should open another issue..

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,9 @@ This document describes changes between each past release.
 13.0.1 (unreleased)
 -------------------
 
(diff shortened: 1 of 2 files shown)


Natim:
Tested and I was able to load the kinto-admin.

Issue #1921, opened 2018-12-06
Merge setup.py, setup.cfg and pyproject.toml ? 


leplatrem:
I don't think we should have all of these.

I would be in favor of `setup.cfg` only :)

glasserc:
I haven't really paid attention to packaging stuff, but I'm in favor of moving to the future (which is I guess `setup.cfg`?).

leplatrem:
There is Flit as well... https://flit.readthedocs.io/en/latest/ which says *no need for setup.py and setup.cfg* ...

leplatrem:
Done in https://github.com/Kinto/kinto/pull/3359

PR #2013, opened 2019-01-30
Use setup.cfg for package metadata (ref #1921)

leplatrem:
Ref #1921

IMO it makes more sense to have package metadata in a configuration file rather than an executed python file.


setup.cfg
```diff
@@ -1,3 +1,81 @@
+[metadata]
+name = kinto
+version = 13.1.0.dev0
```
Natim: Do you know if zest.releaser handle this?

setup.cfg
```diff
@@ -1,3 +1,81 @@
...
+description = Kinto Web Service - Store, Sync, Share, and Self-Host.
+long_description = file: README.rst, CHANGELOG.rst, CONTRIBUTORS.rst
```
Natim: :1st_place_medal:

setup.cfg
Natim: :heart_decoration:

setup.cfg
```diff
@@ -1,3 +1,81 @@
...
+license = Apache License (2.0)
+author = Mozilla Services
+author_email = storage-team@mozilla.com
+url = https://github.com/Kinto/kinto
+download_url = https://github.com/dj-stripe/dj-stripe/tarball/master
```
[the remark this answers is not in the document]
leplatrem: Thanks, fixed in 80c4a7d18

PR #2024, opened 2019-02-05
Fix tox environment dependencies

leplatrem:
I tested it on the commit range where colander was < 1.6 and it pulled the proper version

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -35,6 +35,7 @@ deps =
     pytest-sugar
     webtest
     werkzeug
+install_command = pip install {opts} {packages} -c{toxinidir}/requirements.txt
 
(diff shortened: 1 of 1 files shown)


glasserc:
This might be a dumb question, but what is the value of a `requirements.txt` file? Do we expect our users to retrieve this file and pass it when doing `pip install kinto`?

leplatrem:
> This might be a dumb question, but what is the value of a `requirements.txt` file?

Pin the known good set of dependencies.

> Do we expect our users to retrieve this file and pass it when doing `pip install kinto`?

I don't think so, although they could.

At least we want to make sure that the proper dependencies are pulled when we build the Docker container for example.

PR #2084, opened 2019-03-22
upgrade therapist

peterbe:
[Therapist 1.6.0](https://github.com/rehandalal/therapist/releases/tag/v1.6.0) fixes a [bug related to git copied files](https://github.com/rehandalal/therapist/commit/f5658d1532912717b656004b3f6503980619adcc).

### DIFF
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,4 +1,6 @@
-therapist==1.5.0
+therapist==1.6.0 \
+    --hash=sha256:ba7bebe4df1ee7c78b420b5894824ce8150868eab8819e8a4a510feefc0e2884 \
+    --hash=sha256:e29c8c07676215093318e7b8aadc6c59784a1d9877eff77ffce20c9d63770a24
 black==19.3b0
 flake8==3.7.7
 setuptools==40.8.0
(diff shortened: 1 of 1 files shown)


peterbe:
Oops. Sorry for the first commit.

peterbe:
The ultimate goal is to codify how linting checks are done in a predictable way. If all the devs just channel all their linting checks through one config and that's done on CI too then it's uniformly predictable. 
E.g. One dev might manually run `black -l 80 src` and `flake8 .` but in CI you run `black -l 88 .` and `flake8 src/ tests/` then it might be too different.

The added bonus is that therapist can easily be installed, by everyone, as a pre-commit hook. That way, what is exactly going to potentially block CI is blocking someone from making the PR and wasting CI time. 

At least that's how I understand it.

glasserc:
Thanks for the explanation. For the record, while I agree with most of what you say above, I disagree with the requirement that linting be configured by one config file. But I guess that isn't really relevant to this PR.

peterbe:
> linting be configured by one config file

perhaps semantics but the linting *config* is, in a sense, done elsewhere. E.g. `.flake8`. The `.therapist.yml` is a lint *runner config*. The alternative to typing in the README "Please always run the following linting commands ... Because that's what we will run, for the sake of checking, in CI"

PR #2108, opened 2019-04-08
Fix memcached heartbeat test (fixes #2107)

### LINKED ISSUE #2107: memcached cache backend does not really test memcached
@autrilla reported that when he moved Kinto to GCP, the heartbeat reported success for all backends, including cache, even though there is no `memcached`. I think the problem is here:

https://github.com/Kinto/kinto/blob/61494caae2bb8fe342f32f6a5d89f40ed2b62dff/kinto/core/cache/__init__.py#L87-L91

The heartbeat for the cache backend just sends two messages which don't expect responses. For something like memcached which communicates over UDP, these messages can "succeed" by silently going nowhere.

leplatrem:
Fixes #2107

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,8 +6,9 @@ This document describes changes between each past release.
 13.2.0 (unreleased)
 -------------------
 
-- Nothing changed yet.
+**Bug fixes**
 
+- Fix cache heartbeat test (fixes #2107)
 
 13.1.0 (2019-03-20)
 -------------------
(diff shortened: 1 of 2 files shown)


leplatrem:
> This is kind of racy

Because of UDP? How would you do? sleep for an arbitrary amount of time?

glasserc:
No, I meant it's racy between multiple calls to the heartbeat endpoint. I can't think of any way to make a heartbeat like this that isn't racy. Eventually you have to `get()` some value, and you want to confirm that it's a known value, which means you must have `set()` it earlier, but that means another thread could have `set()` or `delete()`d it in the meantime.

There's also the unfortunate situation here that if memcached is down, I think `get()` returns `None`, which means that the heartbeat will still succeed `_HEARTBEAT_DELETE_RATE` fraction of the time..

leplatrem:
I see. Well, let's take this as an improvement. 
Thanks for the detailed comment!

PR #2149, opened 2019-05-16
pyramid-mailer dependency is missing from requirements

glasserc:


glasserc:
The tests failed... I guess that proves my point, but not in the way I expected.

Natim:
I guess it isn't enough to put it in requirements.txt we should probably also add it in setup.py.

glasserc:
It looks like it was added in https://github.com/Kinto/kinto/pull/1982.

Natim:
I see what you mean.

PR #2151, opened 2019-05-23
Fix support of sqlalchemy.pool.NullPool

leplatrem:
/cc @autrilla

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,11 @@ This document describes changes between each past release.
 **Bug fixes**
 
 - Fix cache heartbeat test (fixes #2107)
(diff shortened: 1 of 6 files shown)


autrilla:
Thanks a lot for the quick turnaround!!

leplatrem:
Tests fail because kinto-redis requires `pool_size`. But, kinto-redis should not be required to run the core tests. Removing this in #2153

glasserc:
The problem reported by @autrilla is that `NullPool` doesn't support the `pool_size` argument, which is added to everything because of the `DEFAULT_SETTINGS`. The logic to process these arguments is backend-specific; see https://github.com/Kinto/kinto/blob/master/kinto/core/storage/postgresql/client.py#L74.

Issue #2169, opened 2019-06-09
Question:  Docstring Format


scwicker:
For a school project I am looking for potential improvements to help contribute to open source projects. This is one I think would benefit the code style and usability of this project. 
In some of the Python files, such as: kinto/kinto/schema_validation.py, there are functions with 
docstrings that could be formatted better to include parameters.  For example, in the following code: 

----------------------------------------------------------------------------------------------
def validate(data, schema):
    """Raise a ValidationError or a RefResolutionError if the data doesn't validate
    with the given schema.
    Note that this function is just a "wrapper" on `jsonschema.validate()` but with
    some memoization based on the schema for better repeat performance.
    """
    # Because the schema is a dict, it can't be used as a hash key so it needs to be
    # "transformed" to something that is hashable. The quickest solution is to convert
    # it to a string.
    # Note that the order of the dict will determine the string it becomes. The solution
    # to that would a canonical serializer like `json.dumps(..., sort_keys=True)` but it's
    # overkill since the assumption is that the schema is very unlikely to be exactly
    # the same but different order.
    cache_key = str(schema)
    if cache_key not in _schema_cache:
        # This is essentially what the `jsonschema.validate()` shortcut function does.
        cls = validator_for(schema)
        cls.check_schema(schema)
        _schema_cache[cache_key] = cls(schema)
    return _schema_cache[cache_key].validate(data)
----------------------------------------------------------------------------------------------
The above function could be re-written in the format below: 

def some_function(argument1):
    """Summary or Description of the Function

    Parameters:
    argument1 (int): Description of arg1

    Returns:
    int:Returning value

   """

    return argument1

----https://www.datacamp.com/community/tutorials/docstrings-python

leplatrem:
There are tools to automate docstrings validation, we could consider adding them to the CI

PR #2170, opened 2019-06-09
Docstring update to schema_validation.py 

scwicker:
This would be the general format that I think would improve the code.  
This wouldn't be ready to merge as is, because the parameter types and descriptions are incomple.

Fixes #

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/kinto/schema_validation.py
+++ b/kinto/schema_validation.py
@@ -48,6 +48,15 @@ _schema_cache = {}
 def validate(data, schema):
     """Raise a ValidationError or a RefResolutionError if the data doesn't validate
     with the given schema.
+    
+    Parameters:
+    data (int): Description of arg1 #needs completed
+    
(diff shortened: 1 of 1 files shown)


glasserc:
Refs #2169. Thanks for your contribution. We actually do try to document parameters and return values in some places, especially in interfaces that will be implemented by other developers. See for example: https://github.com/Kinto/kinto/blob/2a08ddf908f3b0bb357399d67662e123e8aa26ed/kinto/core/permission/__init__.py#L72-L79

Not all of our functions are documented this way, the same way that not all functions have docstrings. All things being equal, it would be nice for every function to have parameter/returns documented, but I don't generally consider it a hard requirement. If we are going to add such annotations, I think it would be good to conform to the style we already have (Sphinx/ReST style).

PR #2195, opened 2019-07-04
Add assertDatetimeAlmostEqual helper for comparing datetime objects.

### LINKED ISSUE #2129: ConfigTest is a little bit intermittent
In https://github.com/Kinto/kinto/pull/2125, ConfigTest failed because the timestamp it generated in the test was different from the timestamp generated in the `kinto.config.init` function. (I restarted the test and merged it when it passed.) This is possible when the test runs just before the end of a second, causing the two calls to `strftime` to refer to two different seconds.

tamarbuta:
Fixes #2129

- [n/a] Add documentation.
- [n/a] Add tests.
- [x] Add a changelog entry.
- [n/a] If you changed the HTTP API, update the [API_VERSION](https://github.com/Kinto/kinto/blob/master/kinto/__init__.py#L15) constant and add an API changelog entry [in the docs](https://github.com/Kinto/kinto/blob/master/docs/api/index.rst)
- [n/a] If you added a new configuration setting, update the [`kinto.tpl`](https://github.com/Kinto/kinto/blob/master/kinto/config/kinto.tpl) file with it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,12 @@ Changelog
 
 This document describes changes between each past release.
 
+13.3.0 (unreleased)
+-------------------
+
+- Nothing changed yet.
+
+
 13.2.0 (2019-06-18)
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,6 @@ This document describes changes between each past release.
 - Upgrade kinto-admin to v1.24.1
 
 
-
 13.2.0 (2019-06-18)
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ This document describes changes between each past release.
 13.3.0 (unreleased)
 -------------------
 
-- Nothing changed yet.
+-  Fix intermittent failure when comparing timestamps in ConfigTest (fixes #2129)
 
 
 13.2.0 (2019-06-18)
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -11,13 +11,27 @@ from kinto import __version__
 
 
 class ConfigTest(unittest.TestCase):
-    def _assertDatetimeAlmostEqual(self, s1, s2):
-        """Assert that two date strings are almost equal."""
+    def _assertTimestampStringsAlmostEqual(
+            self, s1, s2, delta=timedelta(seconds=1)):
+        """Assert that two timestamp strings are almost equal, within a
+        specified timedelta.
+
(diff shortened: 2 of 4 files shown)


tests/test_config.py
```diff
@@ -10,6 +11,16 @@
 
 
 class ConfigTest(unittest.TestCase):
+
+    def _assertDatetimeAlmostEqual(self, s1, s2):
+        """Assert that two date strings are almost equal."""
+        s1 = datetime.strptime(s1, "%a, %d %b %Y %H:%M:%S %z")
+        s2 = datetime.strptime(s2, "%a, %d %b %Y %H:%M:%S %z")
+        return self.assertTrue(
```
leplatrem: I think there's a `self.assertLess()`

tests/test_config.py
```diff
@@ -10,6 +11,16 @@
...
+            abs(s1 - s2) < timedelta(seconds=1),
```
leplatrem: This delta value could be a third optional parameter

CHANGELOG.rst
```diff
@@ -19,6 +20,7 @@ This document describes changes between each past release.
 - Upgrade kinto-admin to v1.24.1
 
 
+
```
leplatrem: nitpick: Not sure this was necessary ;)

leplatrem:
You should run `black` on `test_config.py` also :)

tests/test_config.py
```diff
@@ -10,6 +11,15 @@
 
 
 class ConfigTest(unittest.TestCase):
+    def _assertDatetimeAlmostEqual(self, s1, s2):
```
glasserc: Considering that both of these are strings, could we rename this method? Something like `_assertTimestampStringsAlmostEqual`. Or alternately, since the callers of this method do `strftime`, could we maybe make this method take `datetimes` directly?

tamarbuta: Excellent suggestion. I'll update the method name. 

If we wanted to keep the name the same and pass in datetime objects instead, we'd need to make more invasive changes to kinto/config/__init__.py or convert the stringified time back to datetime before passing to this method. I'll go with the less invasive approach to just change the method name and add a docstring for clarification.

CHANGELOG.rst
tamarbuta: Good catch. This was an accidental space. Removed.

tests/test_config.py
tamarbuta: Good suggestion. Implemented.

tests/test_config.py
tamarbuta: updated to use `self.assertLessEqual`

PR #2283, opened 2019-10-06
Update tests to pass on Windows

dstaley:
This PR makes some small modifications to the tests so that they pass when run on Windows.

## Windows Path Separators 

On Windows, the path separator is the `\` character, as opposed to `/` on *nix/macOS. I modified the tests that were explicitly using the `/` character to use `os.path.join` instead.

## Manually close temp file from `mkstemp`

According to the Python documentation, `tempfile.mkstemp` [returns an OS-level handle to an open file](https://docs.python.org/3.7/library/tempfile.html#tempfile.mkstemp). On Windows, when the `setUp` function was called to remove the temporary file between tests, it was throwing an error that the file couldn't be removed since it was in use by another process. This PR adds a call to `os.fdopen().close()` to explicitly close the open file handler before beginning the tests. It also adjusts the exception to allow only `FileNotFoundError` as opposed to allowing any `OSError`.

After these changes I'm pleased to say that all the tests pass on Windows! :tada:

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -12,14 +12,15 @@ from io import StringIO
 from kinto import __version__ as kinto_version
 from kinto.__main__ import main, DEFAULT_LOG_FORMAT
 
-_, TEMP_KINTO_INI = tempfile.mkstemp(prefix="kinto_config", suffix=".ini")
+fd, TEMP_KINTO_INI = tempfile.mkstemp(prefix="kinto_config", suffix=".ini")
+os.fdopen(fd).close()
 
 
 class TestMain(unittest.TestCase):
     def setUp(self):
         try:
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -12,14 +12,15 @@
 from kinto import __version__ as kinto_version
 from kinto.__main__ import main, DEFAULT_LOG_FORMAT
 
-_, TEMP_KINTO_INI = tempfile.mkstemp(prefix="kinto_config", suffix=".ini")
+fd, TEMP_KINTO_INI = tempfile.mkstemp(prefix="kinto_config", suffix=".ini")
+os.fdopen(fd).close()
```
leplatrem: Can you add a comment to explain why you had to do that please?

PR #2297, opened 2019-10-17
Add Python 3.8 to CI

leplatrem:
* [x]  Wait for https://github.com/mozilla-services/python-dockerflow/pull/38


leplatrem:
I really don't understand the coverage failure with Py38.

When I run the tests locally with py38 and add a breakpoint, I can see that the lines are all executed as expected. @Natim what should we do? Add a `# no cover` :rofl: ??

Natim:
I am pretty sure we are missing a test or the if is wrong in the test with py38.

leplatrem:
> I am pretty sure we are missing a test or the if is wrong in the test with py38.

The line is executed in my local py38...

Natim:
I am pretty sure this is something like bytes vs str variables. Could we log the password and confirm_password values that we have in travis?

Natim:
I just tried to run the test with tox and py38 locally and I have error in the memory cache backend.

Natim:
I confirm 100% even with removing .coverage

PR #2302, opened 2019-10-21
Remove unsolicited log.

Natim:


leplatrem:
What does it mean? Can you elaborate please?

Natim:
It seems that we log really often this. For some reason it appears as an error log in our development environment even if it looks like a info log here.

It looks like a left over from the past, I would gladly remove it.

leplatrem:
> It seems that we log really often this. For some reason it appears as an error log in our development 

Do you have custom config / formatter?

This is supposed to log the batch subrequests. I would keep it, add a setting maybe?

leplatrem:
```
kinto init --ini=/tmp/conf.ini
kinto start --ini=/tmp/conf.ini
```
The logs make sense if I do that: 
```
echo '{"requests":[{"path":"/accounts"}]}' | http POST :8888/v1/batch
```
Look:
Screenshot from 2019-10-22 17-44-00

PR #2324, opened 2019-10-30
PIP requires an update when docker-compose is building images

kaloneh:
Modifying `docker-compose` due to enhancing readability and modularity has been met.
`.env` folder includes three different files containing the relevant configuration of three services declared on `docker-compose.yml`.
Also, adding volumes to persist data storage while docker container is stopped or removed after all. It helps data which accumulated during tests in the development environment still remain on volume and it's clear it can be pruned and disposed at any time.

**PS**: In `Dockerfile`, `pip` requires an update and it's done by `pip install --upgrade pip` that easily along with `httpie` (also _important to debug_) can be chained within `RUN` segment.


Dockerfile
```diff
@@ -17,10 +17,12 @@ RUN \
     curl -sL https://deb.nodesource.com/setup_10.x | bash -; \
     apt-get install -y nodejs; \
     cd kinto/plugins/admin; npm install; npm run build; \
+#    pip3 install --upgrade pip \
+#    apt-get install -y httpie; \
```
leplatrem: Maybe move this `httpie` install on the same line as `nodejs`

Dockerfile
```diff
@@ -17,10 +17,12 @@ RUN \
     curl -sL https://deb.nodesource.com/setup_10.x | bash -; \
     apt-get install -y nodejs; \
     cd kinto/plugins/admin; npm install; npm run build; \
+#    pip3 install --upgrade pip \
```
leplatrem: So, we upgrade or not?

docker-compose.yml
```diff
@@ -1,21 +1,31 @@
...
-    KINTO_STORAGE_URL: postgresql://postgres:postgres@db/postgres
-    KINTO_PERMISSION_BACKEND: kinto.core.permission.postgresql
-    KINTO_PERMISSION_URL: postgresql://postgres:postgres@db/postgres
+version: '3'
+
+services:
+  db:
+    image: postgres
+    env_file:
+      - './.env/db'
```
leplatrem: I don't necessarily find it more readable to have env files. It may be completely valid on big projects, but here it's pretty straightforward don't you think? 

I'm not a super fan of having a level of indirection here, plus another folder on the repo :) But I may be wrong :) 

@Natim @glasserc what do you think?

Dockerfile
kaloneh: It can be shifted to the upper line, however to the sake of readability it's just separated.
If you find it useful in the same manner, yeah, suit yourself :)

Dockerfile
kaloneh: Yeah, buddy, why not?! Actually, I should have said that obviously, it has to be uncommented.

docker-compose.yml
kaloneh: Maybe you right, if you underestimate yourself. It's clear, this project rocks and goes forward so that the future can justify it and look back to this checkpoint!
I think you are developing; so, what is the problem with another folder when it's **grown-up** days and days.
Think about it, you wanna create multiple instances at the same time and you have to change lots of different environment variables; it's not obvious that you _modify three lines_ of code in which define `.env` files _instead of altering all environment variables_? It's messy, don't you think? :D

Natim: I concur with @leplatrem especially since our docker-compose file purpose is to help dev environments.

kaloneh: I'm not insisting on merging all lines, especially, those in `docker-compose` which make hitches. It was an offer to increase modularity that clearly helps the environment to be more flexible. For instance, you _don't have to set a password_ and authorize all users to access the database because it's the main purpose of _dockerization_ in the realm of test and `dev-mode` as a straightforward approach; aren't you fan of simplicity in `dev-mode`? Besides, you can have _multiple users_ with _different roles_**`*`**. Don't you think one specific `docker-compose` associated with several (fully modular) `.env` files can be more **user-friendly** instead of _multiple bloody lines of ENV...ENV_ in `docker-compose` that prunes to make a headache?!
Since you're _in-depth_ and know all about angles, you might prefer a sort of simplicity that you mentioned, I respect it, meanwhile, I suggest look at it one more time _out of the box_ and decide whatever is better to interact with the community and _increasing population_ of your users.

**`*`**_**PS**_: Cleary, you can handle all roles while the access of each user is gonna be grunted via web protocols, what if some different roles can physically access the console? Thanks to docker, here's the purpose of `.env` files to verify your **security level** safe and sound.

Dockerfile
```diff
@@ -17,10 +17,11 @@ RUN \
     curl -sL https://deb.nodesource.com/setup_10.x | bash -; \
     apt-get install -y nodejs; \
     cd kinto/plugins/admin; npm install; npm run build; \
+    pip3 install --upgrade pip; pip3 install httpie; \
     pip3 install -e /app[postgresql,memcached,monitoring] -c /app/requirements.txt; \
     pip3 install kinto-attachment kinto-emailer kinto-elasticsearch kinto-portier kinto-redis; \
     kinto init --ini $KINTO_INI --host 0.0.0.0 --backend=memory --cache-backend=memory; \
-    apt-get purge -y -qq gcc libssl-dev libffi-dev libpq-dev curl nodejs; \
+    apt-get purge -y -qq gcc libssl-dev libffi-dev libpq-dev curl nodejs httpie; \
```
Natim: You didn't install httpie from apt, so this won't work.

docs/tutorials/install.rst
```diff
@@ -210,6 +210,32 @@ Assuming `brew <http://brew.sh/>`_ is installed:
     pip install virtualenv
 
 
+
+Microsoft Windows
```
Natim: Can we put this in another pull-request and keep this one about pip update?

Dockerfile
kaloneh: I respect the simplicity so that to avoid complexity I revert all changes to the first place and the only changes still remain in `Dockerfile`, contain two segments: `httpie` right after `nodejs` at line 24 and the other upgrading `pip` by 

> `pip3 install pip --upgrade`

 before installing packages by `pip3`. It's necessary to upgrade `pip3` in which it reports a warning during the installation process.
All suggestions are postponing to next pull request that I'm gonna update the document to aim Microsoft Windows users and add a `Powershell` script to run the application under the hood.

Natim: Thanks

Natim: Can you elaborate the change on httpie?

docs/tutorials/install.rst
kaloneh: Of course, you got it. I just was about to remove it for know because it should be edited anyway. I'll make an update in another request :)

Dockerfile
kaloneh: > You didn't install httpie from apt, so this won't work.

It doesn't matter because of the `docker` environment. You're monitoring part of the `HTTP` traffic from where it's not necessarily internal and doesn't depend on the version of `Python`. The system installs the `httpie` with respect to `Python2` environment and `pip3` does it correspond to `Python3` core. If you're gonna import `httpie` as a module and write some tests, obviously, `pip3` is middling otherwise the only thing that command line cares about is consistency, bottom line.

kaloneh: Furthermore, the main idea to append the `httpie` into Dockerfile is the nature of *Kinto* to communicate over http(s) protocol. So, it seems attaching `httpie` to the container helps developers to be more equipped in order to test the application omnidirectionally.
Besides, it actually doesn't matter to be installed by either of `pip3` and `apt-get` because it's a handy tool to make get/post/put/delete/patch requests more easily. As I mentioned, if the purpose of adding `httpie` has to cover all aspects, that is, test stuff throughout the command line and pure python scripts, yeah, both `apt-get` and `pip3` should have been involved unless it would be enough to choose one amongst

> pip3 install pip --upgrade && apt-get httpie

where the latter preferably will be installed based on `Python2` core if it's already installed. If `pip3` after `apt-get` installs the `httpie` package and `httpie` has been installed and attached to `PATH` it does not just override the executable file but also install the eggs and it's ready to be imported in python program. Also, if you choose to install it by means of `pip3` it should be done in the following fashion why the upgrade has to be completed before the `pip3` starts to install `httpie`

> pip3 install pip --upgrade
> pip3 install httpie

Any `Python` interpreter has its own specific version of `pip` whatever the major and minor version of all `Python` interpreters are the same so that `pip` needs to be upgraded to the last stable version otherwise causes reporting to warn you to upgrade the `pip`; this was the why I offered you to append both segments including `pip3 install --upgrade pip` and `pip3 install httpie`, however, according to discussion `httpie` installation has no unique way.
I hope that is persuasive and we can finally move on and you merge this PR and go forward to the next PR.

kaloneh: > You didn't install httpie from apt, so this won't work.

Yeah, you right. It was awkwardly had been missed. BTW, it's done.

PR #2333, opened 2019-11-06
Update doc/tutorial/install.rst & add MS Windows instruction.

kaloneh:
Since MS Windows users need some extra tools to run the application properly, the documentation has been updated due to give a guideline in order to satisfy such users.

Fixes #

- [x] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.


dstaley:
I don't think that the steps described in this PR are required to get Kinto running on Windows. You do, however, need to install the Visual C++ Build Tools in order to compile some of the dependencies that use native modules.

The Windows section should essentially be:

```rst
Windows
'''''''

In order to install Kinto, you'll need to `install Python <https://www.python.org/downloads/windows/>`_ and the
`Visual C++ Build tools <https://wiki.python.org/moin/WindowsCompilers>`_.
```

Just to confirm this I spun up a new Windows 10 VM, installed Python, installed the Visual C++ Build Tools, and was able to successfully install and run Kinto.

kaloneh:
> I don't think that the steps described in this PR are required to get Kinto running on Windows...

Yeah, it's true. However, the point is by following the instruction via [Visual C++ Build Tools](https://wiki.python.org/moin/WindowsCompilers) you finally reach the link that represented in the PR as `[Required]` part and rest of it as `[Optionnal]` can help users who looking for a tool like [scoop](https://scoop.sh) to make the process more easily. I just tried to be more specific so that I find it practical to be dealt with.
Since _winpython_ has a bunch of static libraries that make it portable and less dependant on _MS-Windows_ version is [preferable](https://winpython.github.io/#overview), even though, _python3.6+_ with standard core library, of course, is the one who everything depends on.

kaloneh:
The main reason that I made this [PR](https://github.com/Kinto/kinto/pull/2333) was a gap in the manual, so if you agree this is important to dedicate a specific part for _MS-Windows_ users.
Also, there is a motivation for _MS-Windows_ user to be dealt with a flexible command-line tool like [scoop](https://scoop.sh) so that manage and install tools and applications as much as _Unix-Based_ user does (e.g., _Linux_, _Mac_, etc.)
There are some alternatives like _Cygwin_, _MSYS_, _MingW_, etc., however, I recommend the native terminal such as __x86/x64 Native Tool Command Prompt__ to taste a real development environment. Also, the older versions of _MS-Windows_ such as _XP_, _Vista_, _Win7_, _Win8_, _Win8.1_ depends on their **.Net** versions, sometimes, behave unpredictably and always native tools and packages (e.g., _python_ vs _winpython_) are more capable of acting expectable.

leplatrem:
Most of the team members here are not Windows users, so we'll have to trust you guys here.

From an outside standpoint, I wish we could limit the details in the documentation as much as we can, and point to existing ones instead (for example, git, virtualenv, libcffi may have some install docs for Windows somewhere?).

For example, Kinto docs do not explain how to run a PostgreSQL server. We do have a Wiki page though https://github.com/Kinto/kinto/wiki/How-to-run-a-PostgreSQL-server%3F 

The main reason is that these instructions are very likely to become outdated and very few contributors available to update them.

Maybe we should move the non-Kinto specific instructions to a wiki page (git, virtualenv, libcffi ...)

What do you think?

kaloneh:
> For example, Kinto docs do not explain how to run a PostgreSQL server. We do have a Wiki page though https://github.com/Kinto/kinto/wiki/How-to-run-a-PostgreSQL-server%3F

* I'm on it! It's a really good idea to update the Wiki page and introduce some tools that used to help the developer with _MS-Windows_ background. So, I'll try to put useful information over there.

> The main reason is that these instructions are very likely to become outdated and very few contributors available to update them.
> ...
> What do you think?

* Still, there is a gap in the documentation in which doesn't consider all users omnidirectionally so that it might be helpful if respect kinda _MS-Windows_ users with at least few lines of installation instruction. I can trim the instruction if it's required.
Besides, I think it's a need to add a `.bin` (or `cli`) folder under _kinto_ package to write some executable scripts depend on specific platforms (e.g., `kinto.cmd`, `kinto.ps1`, `kinto.sh`, etc.,) each of which checks _(pre)installation requirements_ and produces proper reports.

**PS**: I've already written a Powershell script to satisfy installation process. By the way, as I mentioned, I respect simplicity to avoid stuck in a trap.

leplatrem:
Thanks for taking care of this!

> Besides, I think it's a need to add a .bin (or cli) folder under kinto package to write some executable scripts depend on specific platforms (e.g., kinto.cmd, kinto.ps1, kinto.sh, etc.,) each of which checks (pre)installation requirements and produces proper reports.

I see what you mean. It would be a different topic to discuss, could you open an issue please?

PR #2336, opened 2019-11-06
History entries datetimes now carry timezone information

leplatrem:
Thanks Mike for suggesting this!

### DIFF
--- a/tests/plugins/test_history.py
+++ b/tests/plugins/test_history.py
@@ -10,6 +10,8 @@ from kinto.core.testing import get_user_headers, skip_if_no_statsd
 
 from .. import support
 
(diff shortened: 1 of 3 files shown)


tests/plugins/test_history.py
```diff
@@ -78,7 +78,7 @@ def test_tracks_user_and_date(self):
         resp = self.app.get(self.history_uri, headers=self.headers)
         entry = resp.json["data"][-1]
         assert entry["user_id"] == self.principal
-        assert re.match("^\\d{4}\\-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d{6}", entry["date"])
+        assert re.match("^....-..-..T..:..:..\\.......+..:..$", entry["date"])
```
mythmon: I think I preferred it the other way, but both work fine. You probably could also use raw strings so you don't have to double escape things.

PR #2442, opened 2020-03-04
Replace IRC and Slack badges with Gitter

dstaley:
Replacing the README Slack and IRC badges with Gitter.

It looks like this is the only repo with a chat-related badge, but let me know if you'd like me to add them to other READMEs!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,17 +1,14 @@
 Kinto
 =====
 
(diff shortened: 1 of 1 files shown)


leplatrem:
> It looks like this is the only repo with a chat-related badge, but let me know if you'd like me to add them to other READMEs!

I think that a badge on the main repo is enough...but :man_shrugging:

PR #2460, opened 2020-03-19
Remove `auth` parameter from postgresql/memory storage backends

slav0nic:
Removed `auth` param from postgresq/memory backends that was already droped from base class.
kinto_redis required this changes too


slav0nic:
pytest-sugar broken with latest pytest https://github.com/Teemu/pytest-sugar/issues/187

PR #2468, opened 2020-03-24
Fix permission check on POST on plural (fixes #2459)

### LINKED ISSUE #2459: Possible security flaw?
Here's are the steps to recreate the issue, tested with `kinto/kinto-server:13.6.3` Docker container, with `KINTO_BUCKET_CREATE_PRINCIPALS=system.Authenticated`

Create two users, `alice` and `bob`

```shell
docker exec -it $CID kinto create-user -u alice -p x
docker exec -it $CID kinto create-user -u bob -p x
```

Let `alice` create a bucket:

```shell
echo '{"data":{"id":"alices","secret":"metadata"}}' | http POST http://localhost:8888/v1/buckets --auth "alice:x"
# => HTTP/1.1 200 OK
```

Confirm that `bob` cannot access that bucket:

```shell
http GET http://localhost:8888/v1/buckets/alices --auth "bob:x"
# => HTTP/1.1 403 Forbidden
```

But, if `bob` tries to create a bucket with the same name, they *are* able to see it:

```shell
echo '{"data":{"id":"alices"}}' | http POST http://localhost:8888/v1/buckets --auth "bob:x"
# => HTTP/1.1 200 OK
# {
#     "data": {
#         "id": "alices",
#         "last_modified": 1584544849592,
#         "secret": "metadata"
#     },
#     "permissions": {}
# }
```

When trying to create objects with existing IDs Kinto returns the existing object. In this case `account:bob` has the `bucket:create` permissions but they don't have explicit `read` for a bucket created by `alice`.

leplatrem:
Fixes #2459

### DIFF
--- a/kinto/core/authorization.py
+++ b/kinto/core/authorization.py
@@ -84,7 +84,11 @@ class AuthorizationPolicy:
         # If not allowed on this plural endpoint, but some objects are shared with
         # the current user, then authorize.
         # The :class:`kinto.core.resource.Resource` class will take care of the filtering.
-        is_list_operation = context.on_plural_endpoint and not permission.endswith("create")
+        is_list_operation = (
+            context.on_plural_endpoint
+            and not permission.endswith("create")
+            and context.current_object is None
+        )
         if not allowed and is_list_operation:
             allowed = bool(
                 context.fetch_shared_objects(permission, principals, self.get_bound_permissions)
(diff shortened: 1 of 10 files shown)


kinto/core/authorization.py
```diff
@@ -151,22 +155,32 @@ def __init__(self, request):
...
-                resource = service.resource(request=request, context=self)
-                try:
-                    # Save a reference, to avoid refetching from storage in resource.
-                    self.current_object = resource.model.get_object(resource.object_id)
-                except storage_exceptions.ObjectNotFoundError:
-                    pass
+            )
+            is_post_on_plural = self.on_plural_endpoint and request.method.lower() == "post"
+            if is_write_on_object or is_post_on_plural:
+                # We instantiate the resource to determine the object targetted by the request.
```
glasserc:
```suggestion
                # We instantiate the resource to determine the object targeted by the request.
```

Sorry... I don't make the rules :slightly_smiling_face:

kinto/core/authorization.py
```diff
@@ -283,4 +297,13 @@ def _find_required_permission(self, request, service):
...
                     required_permission = "write"
 
+        # In the case of a "POST" on a plural endpoint, if an "id" was
+        # specified, then the object is returned. The required permission
+        # is thus "read" on this object.
+        if request.method.lower() == "post" and self.current_object is not None:
+            permission_object_id = self.get_permission_object_id(
+                request, object_id=self._resource.object_id
+            )
+            required_permission = "read"
```
glasserc: Wait a sec. Won't the POST in the linked issue replace the data? Shouldn't it therefore be `write`? I guess if they can't do the POST then we want `read` to show them the existing record?

leplatrem: No, if `id` is specified and object exists, then we return the existing object:
https://docs.kinto-storage.org/en/stable/api/1.x/records.html#uploading-a-record

glasserc: I see. OK, sure.

PR #2469, opened 2020-03-25
Add PostgreSQLClient method to set connection options

leplatrem:


leplatrem:
This will set the options at the client level, which is not thread safe

PR #2474, opened 2020-03-30
Fix resource timestamp unicity (fixes #2472, #602)

### LINKED ISSUE #2472: Postgres' bump_timestamp function should round timestamps to epoch precision
Under heavy concurrent load, two records may end in Postgres with effectively the same epoch. Here's an example:

```
> SELECT * FROM objects ORDER BY last_modified;

                id                    | parent_id | resource_name |       last_modified        | data | deleted 
--------------------------------------+-----------+---------------+----------------------------+------+---------
 4fd08eed-4532-462e-a4b3-deb120317bde | /foo      | object        | 2020-03-27 16:09:52.388    | {}   | f
 b3d05073-e345-438d-9992-79b75a3fb8c7 | /foo      | object        | 2020-03-27 16:09:52.389    | {}   | f
 03db7f27-bf88-4848-9ef2-292a7dcf8cc0 | /foo      | object        | 2020-03-27 16:09:52.389808 | {}   | f
 a524e6ad-eac2-4abd-a83e-ade58a7c58d0 | /foo      | object        | 2020-03-27 16:09:52.389928 | {}   | f
 cb791368-80e7-4dae-9daa-d267a3d04a87 | /foo      | object        | 2020-03-27 16:09:52.39     | {}   | f
 b25378de-e6bf-46a1-a2de-94190ca32e33 | /foo      | object        | 2020-03-27 16:09:52.391    | {}   | f
```

Here's my fix:

```diff
--- schema.sql
+++ schema.sql
@@ -64,8 +64,8 @@ DROP TRIGGER IF EXISTS tgr_objects_last_modified ON objects;
 CREATE OR REPLACE FUNCTION bump_timestamp()
 RETURNS trigger AS $$
 DECLARE
-    previous TIMESTAMP;
-    current TIMESTAMP;
+    previous BIGINT;
+    current BIGINT;
 BEGIN
     previous := NULL;
     WITH existing_timestamps AS (
@@ -87,7 +87,7 @@ BEGIN
           AND resource_name = NEW.resource_name
       )
     )
-    SELECT MAX(last_modified) INTO previous
+    SELECT as_epoch(MAX(last_modified)) INTO previous
       FROM existing_timestamps;
 
     --
@@ -99,16 +99,16 @@ BEGIN
     -- an error (operation is cancelled).
     -- See https://github.com/mozilla-services/cliquet/issues/25
     --
-    current := clock_timestamp();
+    current := as_epoch(clock_timestamp()::TIMESTAMP);
     IF previous IS NOT NULL AND previous >= current THEN
-        current := previous + INTERVAL '1 milliseconds';
+        current := previous + 1;
     END IF;
 
     IF NEW.last_modified IS NULL OR
-       (previous IS NOT NULL AND as_epoch(NEW.last_modified) = as_epoch(previous)) THEN
+       (previous IS NOT NULL AND as_epoch(NEW.last_modified) = previous) THEN
         -- If record does not carry last-modified, or if the one specified
         -- is equal to previous, assign it to current (i.e. bump it).
-        NEW.last_modified := current;
+        NEW.last_modified := from_epoch(current);
     END IF;
 
     RETURN NEW;
```

leplatrem:
- [x] Import @dim's code from #2472 (huge thanks!)
- [ ] Design a reliable unit test

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -77,7 +77,7 @@ class Storage(StorageBase, MigratorMixin):
 
     # MigratorMixin attributes.
     name = "storage"
-    schema_version = 21
+    schema_version = 22
     schema_file = os.path.join(HERE, "schema.sql")
     migrations_directory = os.path.join(HERE, "migrations")
 
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -201,7 +201,7 @@ class Storage(StorageBase, MigratorMixin):
             FROM objects
             WHERE parent_id = :parent_id
               AND resource_name = :resource_name
-            ORDER BY last_modified DESC
+            ORDER BY as_epoch(last_modified) DESC
             LIMIT 1
           )
           -- Timestamp of empty resource.
--- a/kinto/core/storage/postgresql/migrations/migration_021_022.sql
+++ b/kinto/core/storage/postgresql/migrations/migration_021_022.sql
@@ -12,7 +12,7 @@ BEGIN
         FROM objects
         WHERE parent_id = NEW.parent_id
           AND resource_name = NEW.resource_name
-        ORDER BY last_modified DESC
+        ORDER BY as_epoch(last_modified) DESC
         LIMIT 1
       )
       -- Timestamp when resource was empty.
--- a/kinto/core/storage/postgresql/migrations/migration_021_022.sql
+++ b/kinto/core/storage/postgresql/migrations/migration_021_022.sql
@@ -1,3 +1,5 @@
+DROP TRIGGER IF EXISTS tgr_objects_last_modified ON objects;
+
 CREATE OR REPLACE FUNCTION bump_timestamp()
 RETURNS trigger AS $$
 DECLARE
@@ -51,6 +53,10 @@ BEGIN
     RETURN NEW;
 END;
 $$ LANGUAGE plpgsql;
+
+CREATE TRIGGER tgr_objects_last_modified
+BEFORE INSERT OR UPDATE OF data ON objects
+FOR EACH ROW EXECUTE PROCEDURE bump_timestamp();
 
 -- Bump storage schema version.
 INSERT INTO metadata (name, value) VALUES ('storage_schema_version', '22');
--- a/kinto/core/storage/postgresql/migrations/migration_021_022.sql
+++ b/kinto/core/storage/postgresql/migrations/migration_021_022.sql
@@ -1,0 +1,56 @@
+CREATE OR REPLACE FUNCTION bump_timestamp()
+RETURNS trigger AS $$
+DECLARE
+    previous BIGINT;
+    current BIGINT;
+BEGIN
+    previous := NULL;
+    WITH existing_timestamps AS (
+      -- Timestamp of latest record.
+      (
+        SELECT last_modified
+        FROM objects
+        WHERE parent_id = NEW.parent_id
+          AND resource_name = NEW.resource_name
+        ORDER BY last_modified DESC
+        LIMIT 1
+      )
+      -- Timestamp when resource was empty.
+      UNION
+      (
+        SELECT last_modified
+        FROM timestamps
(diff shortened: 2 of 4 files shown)


dim:
@leplatrem it's unrelated to this issue but since you are adding a migration I thought it may be worth addressing it in the same PR.

In your `WITH existing_timestamps AS` block, in the function as well as in the https://github.com/Kinto/kinto/blob/master/kinto/core/storage/postgresql/__init__.py, you are using:

```
SELECT last_modified
FROM objects
...
ORDER BY last_modified DESC
LIMIT 1
```

The problem is that the `objects` table only has an index for `"idx_objects_last_modified_epoch" btree (as_epoch(last_modified))`. Therefore:

```
# EXPLAIN SELECT last_modified FROM objects ORDER BY last_modified DESC LIMIT 1;
                                QUERY PLAN                                 
---------------------------------------------------------------------------
 Limit  (cost=3275.67..3275.67 rows=1 width=8)
   ->  Sort  (cost=3275.67..3601.94 rows=130511 width=8)
         Sort Key: last_modified DESC
         ->  Seq Scan on objects  (cost=0.00..2623.11 rows=130511 width=8)
```

vs:

```
# EXPLAIN SELECT last_modified FROM objects ORDER BY as_epoch(last_modified) DESC LIMIT                                                        QUERY PLAN                                                       
------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.29..0.59 rows=1 width=16)
   ->  Index Scan Backward using idx_objects_last_modified_epoch on objects  (cost=0.29..38145.73 rows=131105 width=16)
```

leplatrem:
Awesome! thanks! It's been quite a while since no one reviewed thoroughly the low-level SQL parts! Niice!

glasserc:
> The problem is that the objects table only has an index for "idx_objects_last_modified_epoch" btree (as_epoch(last_modified)). Therefore:

Which query are we talking about here? The one in `resource_timestamp`? Shouldn't this query instead be using `idx_objects_parent_id_resource_name_last_modified ON objects(parent_id, resource_name, last_modified DESC);` ?

dim:
@glasserc good point. my apologies, I have removed the condition on `parent_id` and `resource_name` from my example to make it more readable but it's obviously quite essential in this context.

dim:
to confirm: 
```
EXPLAIN SELECT last_modified FROM objects WHERE parent_id = '/foo' AND resource_name = 'object' ORDER BY last_modified DESC LIMIT 1;
                                                            QUERY PLAN                                                             
-----------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.29..0.34 rows=1 width=8)
   ->  Index Only Scan using idx_objects_parent_id_resource_name_last_modified on objects  (cost=0.29..1099.80 rows=19935 width=8)
         Index Cond: ((parent_id = '/foo'::text) AND (resource_name = 'object'::text))
```

leplatrem:
We have a unit test that asserts about the timestamp unicity:
https://github.com/Kinto/kinto/blob/bb6b3e053fbb4006035f92e3db1ecf27b481f32f/kinto/core/storage/testing.py#L786-L787

It is disabled on Travis because it fails there (it passes on my machine)

But I cannot find the PostgreSQL parameter that would cause the transactions not to be isolated. I thought it was `fsync = off`
https://github.com/travis-ci/travis-cookbooks/blob/b5e4fc1f3cd153f9a8485998eed63789e7c146b1/cookbooks/travis_postgresql/templates/ubuntu/9.5/postgresql.conf.erb#L181

But the docs do not say anything about isolation: https://www.postgresqlco.nf/en/doc/param/fsync/
:man_shrugging:

revolunet:
HI ! any news about this useful fix ? we're facing some uniqueness issues on heavy load too

leplatrem:
Hi @revolunet! I had no idea you were using Kinto ;) 

So, I wanted to take advantage of this PR to re-enable the timestamp unicity test on TravisCI, but couldn't figure out the root cause. 

I'm happy to merge and release, and if you could give us some feedback that would be awesome!

revolunet:
Hi there, yes we're using kinto in a bunch of projects and thank you for the work to make this tool available and this fix that should make it more reliable :)

PR #2475, opened 2020-03-30
Simplify Postgres get_object_permissions implementation

glasserc:
@smarnach turned up these queries while trying to investigate why the permissions backend was using so much memory. He observed that in the most common case (a single object_id), the query basically devolves into a WHERE clause. This inspired me to simplify this query a bit. We don't need Postgres to do the ordering since we do it ourselves on the way out using the OrderedDict. That means we can get rid of the join.

- (n/a) Add documentation.
- (n/a) Add tests.
- [ ] Add a changelog entry.


glasserc:
I also toyed with the idea of optimizing this query for the single-object-id case but once I got to this form, it didn't seem worth it.

smarnach:
@glasserc Can you cut a new release once this is merged, so I can deploy it to kintowe?

PR #2596, opened 2020-09-01
Do not return 400 for ?_since=null (fixes #2595)

### LINKED ISSUE #2595: Cannot fetch records list with _since=null anymore
Since https://github.com/Kinto/kinto/pull/2525 the server returns 400 when a plural endpoint is called with `_since=null`. This is not intended and breaks the Admin UI in certain situations.

leplatrem:
Fixes #2595

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.


Natim (approved):
:kissing_heart:

PR #2617, opened 2020-10-01
Update pyproject.toml

vainisha:
Fixes #

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.


leplatrem:
Can you explain this change? I'm not against following recent good practices but at least justify a little bit why we should change this

Natim (changes requested):
The pep8 state that we should use 99 nowadays.

Issue #893, opened 2016-10-28
Add write permission to objects only if necessary


leplatrem:
Currently, when an object is edited, the `write` permission is given to the creator/editor.

This way, there's always someone with `write` permission.

Problem, revoking permission on children is hard  (#144).

One idea could be to add the `write` permission only if the user has no `write` permission on the parent.

Examples:
- `write` on bucket → create sub collection → no write on sub collection
- `write` on bucket → update sub collection → no write on sub collection
- `read` and `collection:create` on bucket → create collection → write added on collection
- `read` and `record:create` on collection → create record → write added on record

Permissions specified via settings should be taken into account with this approach. Related https://github.com/Kinto/kinto/issues/892

PR #2618, opened 2020-10-01
Add ability to disable explicit permissions at object level (ref #893)

leplatrem:
Ref #893 

I'm looking at our DB in prod, and the `history` plugin combined with our explicit permissions system adds around 5K rows every month. And most of them are for permissions on the history entries (99.17% of 1M+ rows)

``` 
[kintoprod] # SELECT COUNT(*) FROM access_control_entries WHERE object_id LIKE '/buckets/%/history/%';
  count  
---------
 1165553
(1 row)

Time: 663,572 ms
[kintoprod] # SELECT COUNT(*) FROM access_control_entries WHERE object_id NOT LIKE '/buckets/%/history/%';
 count 
-------
  9715
(1 row)


``` 

This is not ideal, and it makes the [permission endpoint]() super slow of course (used in Kinto Admin to list accessible/writable objects on the server).

Introducing a new setting to disable explicit permissions is a cheap and simple way to overcome this issue. @Natim what do you think of the approach? (I will add tests of course)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,10 @@ This document describes changes between each past release.
 14.0.2 (unreleased)
 -------------------
 
+**New feature**
+
+- Add ability to disable explicit permissions at object level (ref #893). Use ``kinto.explicit_permissions = false`` to only rely on inherited permissions (see settings docs)
+
 **Internal Changes**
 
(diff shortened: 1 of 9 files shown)


Natim:
Génial !!

Natim (approved):

Issue #1535, opened 2018-03-13
Add a way to change the default_bucket_id generator


Natim:
Today the ``default_bucket ID`` is calculated using the ``user ID`` and the ``user_hmac_secret``

In my case I would like to be able to give it a more meaningful and stable ID.

One way of doing that is [overriding the `request.default_bucket_id` value using a new plugin](https://github.com/Kinto/kinto/blob/f5167e86d7975cbeebfa2c1c549978e2686f8b5f/kinto/plugins/default_bucket/__init__.py#L207), but I wish I could configure the default_bucket plugin to give it the path to that function.

Issue #881, opened 2016-10-26
Add a specific salt key for default_bucket ID generation


Natim:
Here https://github.com/Kinto/kinto/blob/master/kinto/plugins/default_bucket/__init__.py#L172

We are using the same secret as the BasicAuth plugin.
We should fallback to the basic auth secret but we should use another prefered key that doesn't contains secret but salt in its name.

    default_bucket_ID_salt

mansimarkaur:
@Natim How many characters and base salt do you think should be used?

Natim:
As much as security people want's to use.

Right now they want to use this command to generate the salt:

``cat /dev/urandom | tr -dc _A-Z-a-z-0-9 | head -c${1:-32}; echo``

sahildua2305:
@Natim how do we plan to take input for the salt (or the pattern) from the user?

Natim:
The same way we do for the userHmacSecret I guess. Using the command ``kinto init`` to generate it and then we will have something different for kinto-heroku and in the doc we can explain to people how to change it.

leplatrem:
Related #691

suraj1074:
@Natim So basically we need to make this change in the file [https://github.com/Kinto/kinto/blob/master/kinto/plugins/default_bucket/__init__.py#L172](url)

```
secret = settings['default_bucket_ID_salt'] 
if secret == None:
    secret = settings['userid_hmac_secret']
```

Natim:
Yes something like that as well as adding some tests and updating the kinto command.

```
secret = settings.get('default_bucket_ID_salt', settings['userid_hmac_secret'])
```

chinchaun:
Hello there i was looking at this issue and trying to fix it, i have a question related testing the changes,
in the test files the logic is been mimic of the function default_bucket_id
``` python
def test_default_bucket_exists_and_has_user_id(self):
        bucket = self.app.get(self.bucket_url, headers=self.headers)
        result = bucket.json
        settings = self.app.app.registry.settings
        hmac_secret = settings['userid_hmac_secret']
        bucket_id = hmac_digest(hmac_secret, self.principal)[:32]

        self.assertEqual(result['data']['id'], str(UUID(bucket_id)))
        self.assertEqual(result['permissions']['write'], [self.principal])
````

but the method is not been invoked, i don't have to much experience in python so i don't know how to invoke the method that is need to be tested of the new behaviour implementation or maybe i don't missing something.

Kind regards

glasserc:
[...]

PR #2647, opened 2020-10-31
Default bucket salt

feiming:
Fixes https://github.com/Kinto/kinto/issues/1535 and #881

- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -820,6 +820,7 @@ Legacy Basic Auth
 In the first versions of Kinto, we had a built-in ``basicauth`` policy enabled by default.
 
 Basically it generates a unique :term:`user identifier` from any username/password combination using a HMAC secret.
+Bucket IDs are generated using salt declared in `default_bucket_id_salt`.
 
 Even if it was convenient to get started, we decided to get rid of it because it was very confusing. But you can enable it with the follow configuration:
 
@@ -828,6 +829,7 @@ Even if it was convenient to get started, we decided to get rid of it because it
     multiauth.policies = basicauth
 
     kinto.userid_hmac_secret = have-you-seen-the-new-carioca
+    kinto.default_bucket_id_salt = bucket-id-random-salt-garam
     multiauth.policy.basicauth.use = kinto.core.authentication.BasicAuthAuthenticationPolicy
 
 .. _settings-kinto-auth-plugins:
--- a/docs/configuration/settings.rst
+++ b/docs/configuration/settings.rst
@@ -820,7 +820,7 @@ Legacy Basic Auth
 In the first versions of Kinto, we had a built-in ``basicauth`` policy enabled by default.
 
 Basically it generates a unique :term:`user identifier` from any username/password combination using a HMAC secret.
-Bucket IDs are generated using salt declared in `default_bucket_id_salt`.
+Bucket IDs are generated using salt declared in `default_bucket_hmac_secret`.
 
 Even if it was convenient to get started, we decided to get rid of it because it was very confusing. But you can enable it with the follow configuration:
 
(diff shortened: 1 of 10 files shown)


feiming:
please add `hacktoberfest-accepted` label to this PR

Natim:
Thank you for you work on this !

tests/test_configuration/test.ini
```diff
@@ -83,6 +83,10 @@ kinto.permission_url = permission_url
 #
 kinto.userid_hmac_secret = secret
 multiauth.policies = account
+
+# Bucket ID salt
+kinto.default_bucket_ID_salt = bucket_id_salt
```
Natim, line 88:
```suggestion
kinto.default_bucket_id_salt = bucket_id_salt
```

Natim (changes requested):
Do you mind adding a line in the documentation next to `userid_hmac_secret` (see docs/configuration/settings.rst)
I wonder if to be consistent we should call the config `default_bucket_hmac_secret` rather than `default_bucket_id_salt`

feiming:
@Natim I've renamed all default_bucket_ID_salt to default_bucket_id_salt

kinto/config/__init__.py
```diff
@@ -67,6 +67,7 @@ def init(config_file, backend, cache_backend, host="127.0.0.1"):
 
     values["host"] = host
     values["secret"] = core_utils.random_bytes_hex(32)
+    values["bucket_id_salt"] = core_utils.random_urlsafe(32)
```
Natim, line 70: Let's use the same function.

```suggestion
    values["bucket_id_salt"] = core_utils.random_bytes_hex(32)
```

kinto/core/utils.py
```diff
@@ -103,6 +104,15 @@ def random_bytes_hex(bytes_length):
...
 
+def random_urlsafe(length):
+    """Return a urlsafe secret
+
+    :param int length: number of random string.
+    :rtype: str
+    """
+    return secrets.token_urlsafe(length)
+
+
```
Natim, line 115: If we use the existing method, we don't need this one.

```suggestion
```

Natim:
I would add a test to see that when the default_bucket_hmac secret is not defined it uses the user_hmac one and that when it is defined it doesn't.

feiming:
> I would add a test to see that when the default_bucket_hmac secret is not defined it uses the user_hmac one and that when it is defined it doesn't.

like this? https://github.com/Kinto/kinto/pull/2647/files#diff-79caf24359b02ce1c1bed59ce1514942e785f4498d3474758287796bb5dd5af7R38

docs/configuration/settings.rst
```diff
@@ -820,6 +820,7 @@ Legacy Basic Auth
 In the first versions of Kinto, we had a built-in ``basicauth`` policy enabled by default.
 
 Basically it generates a unique :term:`user identifier` from any username/password combination using a HMAC secret.
+Bucket IDs are generated using salt declared in `default_bucket_hmac_secret`.
```
leplatrem, line 823: nit: use double back ticks in RST

leplatrem (approved):

feiming:
> I would add a test to see that when the default_bucket_hmac secret is not defined it uses the user_hmac one and that when it is defined it doesn't.

I'm not too sure how to test it. Does something like this work?

```
    def test_default_bucket_hmac_secret_define(self):
        settings = {"default_bucket_hmac_secret":"bucket_id_salt","userid_hmac_secret":"secret"}
        secret = settings.get("default_bucket_hmac_secret", settings["userid_hmac_secret"])

        self.assertEqual(secret,"bucket_id_salt")

    def test_default_bucket_hmac_secret_not_define_fallback_to_userid_hmac_secret(self):
        settings = {"userid_hmac_secret":"super_secret"}
        secret = settings.get("default_bucket_hmac_secret", settings["userid_hmac_secret"])

        self.assertEqual(secret,"super_secret")
```

leplatrem:
Maybe you could test that `hmac_digest()` is called with a value or another?

feiming:
I had to mock `hmac_secret` because the default `hmac_secret` is an empty string. I also had to delete `default_bucket_hmac_secret` to test if it's not defined.

I hope this satisfied the test case.

feiming:
@Natim I've added the test 

> I would add a test to see that when the default_bucket_hmac secret is not defined it uses the user_hmac one and that when it is defined it doesn't.

@Natim I've added the test

Issue #1544, opened 2018-03-21
Kinto plugins settings can't be loaded using ENV variable


Natim:
During the initialization phase, we are checking for each Kinto default settings if any env variable are defined.

https://github.com/Kinto/kinto/blob/master/kinto/core/initialization.py#L512-L514

However it only works for kinto default settings, I believe it doesn't for plugin's settings.

leplatrem:
We iterate on the settings defined in the .ini + the defaults:

https://github.com/Kinto/kinto/blob/86a9943b428b8e7fab53190bf840222f5c833225/kinto/core/initialization.py#L480-L497

Natim:
Does it means that if we want to override a value with ENV variables we need to define it with a wrong value in the ini file?

leplatrem:
According to my understanding of the code, it seems like a workaround yes

leplatrem:
Unless we add a helper like `request.get_setting("key", 128)` and use it instead of `request.registry.settings.get()` I don't see how we could fix this.

leplatrem:
Ok, we can add an additional initialization step that does that.

But I think the input list should be settings names because I'm not sure we can reverse (`SIGNER_CID_PATH` should be `signer.cid.path` or `signer_cid.path`?)

leplatrem:
I like the idea of a helper like `request.config()` with option similar to `python-decouple`

leplatrem:
```
config.kinto_setting("mail.host", default="localhost")
config.kinto_setting("mail.port", cast=int, default=25)
config.kinto_setting("mail.password", env="PASSWD")
```
See [an example of how to add a config method](https://github.com/Kinto/kinto/blob/9ffd27856802ffbd251b8e5b548b5b6fa0e01812/kinto/core/__init__.py#L161-L171).

See [loading of default settings](https://github.com/Kinto/kinto/blob/9ffd27856802ffbd251b8e5b548b5b6fa0e01812/kinto/core/initialization.py#L484-L525)

leplatrem:
In `core/__init__.py`, add a `kinto_setting` directive with something like that:

```py
def kinto_setting(config, name, default=None, cast=None, env=None):
    config.registry.kinto_settings[name] = {
        "default": default,
        "cast": cast,
        "env": env or name.upper(),
    }

config.add_directive("kinto_setting", kinto_setting)
```

[...]

PR #2662, opened 2020-11-30
feat: allow env-var based email config

revolunet:
Fixes https://github.com/Kinto/kinto-emailer/issues/233

Hi, i'm not sure this is the way to fix it, but this allow to use environment variables to configure pyramid email settings

I can now use these variables without any ini file : 

```sh
MAIL_DEFAULT_SENDER=toto@toto.com
MAIL_HOST=smtp.orange.fr
MAIL_PORT=25
```

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Add a changelog entry.

### DIFF
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -36,6 +36,11 @@ DEFAULT_SETTINGS = {
     "group_id_generator": "kinto.views.NameGenerator",
     "record_id_generator": "kinto.views.RelaxedUUID",
     "project_name": "kinto",
+    "mail.host": "localhost",
+    "mail.port": 25,
+    "mail.default_sender": None,
+    "mail.username": None,
+    "mail.password": None,
 }
 
 


kinto/__init__.py
```diff
@@ -36,6 +36,11 @@
     "group_id_generator": "kinto.views.NameGenerator",
     "record_id_generator": "kinto.views.RelaxedUUID",
     "project_name": "kinto",
+    "mail.host": "localhost",
+    "mail.port": 25,
+    "mail.default_sender": None,
+    "mail.username": None,
+    "mail.password": None,
```
leplatrem, line 43: This is one way of solving it, but I think it would be better to introduce a sort of registry to be used by plugins. 

kinto-emailer could register expected settings:

```py
config.kinto_setting("mail.host", default="localhost")
config.kinto_setting("mail.port", cast=int, default=25)
config.kinto_setting("mail.password", env="PASSWD")
```

See https://github.com/Kinto/kinto/issues/1544

revolunet: i have no idea how to do this :) is there any plugin as example ?

leplatrem: I wrote some details here https://github.com/Kinto/kinto/issues/1544#issuecomment-743241112

leplatrem: Let's discuss there ;)

leplatrem:
Please reopen if you still would like to see this implemented :)

Issue #2643, opened 2020-10-26
Run tests in GitHub Actions


dstaley:
I think it would be wise to have at least one set of CI validation coming from GitHub Actions, so that we have two sources of code quality, especially for times like now where open source builds on Travis are delayed for several hours.

I don't think we need to do the multiple Python version testing, but at least testing on the latest Python in GHA would be nice.

dstaley:
On Monday, Travis introduced a [new billing plan](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) for OSS projects that's based on a credits system. Since then, we've consumed almost half of our monthly allotment in just under 72 hours. They do have a [process in place](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing#building-on-a-public-repositories-only) where we can request additional credits, but it sounds like this would be something we'd have to continually do as we ran out of credits.

I think it might be a wise idea to move as much CI testing as we can off of Travis and over to GitHub Actions. It looks like the JavaScript repos (`kinto-admin`, `kinto-http.js`, and `kinto.js`) trigger quite often, and are probably the lowest hanging fruit, so I'll work on migrating those over first. My hope is that we can get enough tests off of Travis so that we remain comfortably within their 1000 build minutes a month allotment.

@leplatrem @Natim just pinging you so you're aware!

dstaley:
I reached out to Travis support and we now have an additional 70,000 credits (7000 minutes) to use. This should (hopefully) give us about a month or so of runway to reduce our usage.

dstaley:
At least one user has been notified by Travis that the open source credit system is now on hold. [Source](https://twitter.com/james_hilliard/status/1336081776691843072). I'm going to try and get GHA to parity with Travis on all JavaScript repos by the end of this week.

leplatrem:
I cannot promise to work on this in the short term, but I will reduce the frequency of `dependabot` right now, it may help a little bit

dstaley:
Well, it finally happened: we ran out of OSS credits. I'm going to ask Travis and see if we can get a few more just to get us through the transition.

leplatrem:
I think we can just ditch TravisCI, it's fine.

PR #2677, opened 2020-12-14
Add GitHub Actions workflows

dstaley:
As discussed in #2643, this PR mirrors the CI steps we're currently performing in Travis to GitHub Actions.

This PR makes the following changes:
- pip now installs using the [legacy resolver](https://pip.pypa.io/en/stable/user_guide/#resolver-changes-2020), which matches the installation behavior it used prior to the 20.3 upgrade released earlier this month. There are some incompatibilities in our dependencies, so I'm hoping that those get resolved by the time the legacy resolver is deprecated in January.
- Sphinx has been downgraded to the last known working version, 3.0.3.
- The latest versions of pip and virtualenv are installed into the environment prior to tox being run

In addition to the above changes, this PR adds GitHub Actions workflows mirroring the things we're currently doing in Travis.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,10 +4,10 @@ on:
       - master
   pull_request:
 
-name: Functional Testing
+name: Unit Testing
 jobs:
   chore:
-    name: Functional
+    name: Unit Tests
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,74 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+name: Functional Testing
+jobs:
+  chore:
+    name: Functional
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        toxenv: [py36, py36-raw, py37, py38, py39]
+        include:
+          - toxenv: py36
+            python-version: "3.6"
+          - toxenv: py36-raw
+            python-version: "3.6"
+          - toxenv: py37
+            python-version: "3.7"
+          - toxenv: py38
+            python-version: "3.8"
+          - toxenv: py39
+            python-version: "3.9"
+
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install virtualenv
+        run: |
+          pip install virtualenv
+          virtualenv --python=python3 .venv
+
+      - name: Print environment
+        run: |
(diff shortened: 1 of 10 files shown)


leplatrem (commented):
Nice start!!

.github/workflows/test.yml
```diff
@@ -0,0 +1,83 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+name: Functional Testing
```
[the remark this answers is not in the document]
dstaley: Ha, yeah, I copy and pasted and haven't gotten around to fixing that yet.

Natim:
@dstaley why the calls to which? Is it debugs? Shall we get rid of it?

dstaley:
> @dstaley why the calls to `which`? Is it debugs? Shall we get rid of it?

@Natim I don't see `which` in the diff. Could you link me to the line you're talking about?

Natim:
My bad I probably looked at a commit diff rather than the PR diff.

dstaley:
@leplatrem @Natim if one of you could review this so I could merge it that would be great. It includes fixes to get Travis building again, so I can clear out some of the outstanding dependency update PRs.

Natim:
I think it is good to go for me !

tox.ini
```diff
@@ -16,7 +20,9 @@ deps =
 install_command = pip install {opts} {packages} -c{toxinidir}/requirements.txt
 
 [testenv:py36-raw]
-passenv = TRAVIS
+passenv = TRAVIS CI
+setenv =
+    PIP_USE_DEPRECATED = legacy-resolver
```
leplatrem, line 25: Should we create an issue to remove that?

kinto/core/testing.py
```diff
@@ -12,7 +12,7 @@
 from kinto.core.storage import generators
 from kinto.core.utils import encode64, follow_subrequest, memcache, sqlalchemy
 
-skip_if_travis = unittest.skipIf("TRAVIS" in os.environ, "travis")
+skip_if_travis = unittest.skipIf("CI" in os.environ, "travis")
```
leplatrem, line 15: Should it be `skip_if_ci`?

.travis.yml
```diff
@@ -10,7 +10,9 @@ env:
   - TOX_ENV=py36
   - TOX_ENV=py36-raw
 install:
+  - pip install -U pip
   - pip install tox
+  - pip install -U virtualenv
 before_script:
   - echo "Pull request ${TRAVIS_PULL_REQUEST}"
   - sudo sed -i "s/fsync/#fsync/" /etc/postgresql/9.5/main/postgresql.conf
```
leplatrem, line 18: Aren't we getting rid of travis?

leplatrem (approved):
Thanks Dylan!

This looks good to me :)

tox.ini
dstaley: This is a problem with a dependency, but I'll still create one with details so we can keep track of when it's no longer needed.

kinto/core/testing.py
dstaley: Yeah, I'll change that!

.travis.yml
dstaley: That's the goal, but I want a few PRs with both to compare and make sure they're doing the same stuff.

PR #2722, opened 2021-02-17
Restore turn Sphinx warnings into errors

Natim:
Fixes Pylons/pyramid#3655


Natim:
Should be fixed in the next Pyramid release: https://github.com/Pylons/pyramid/pull/3657

PR #2737, opened 2021-03-04
Replace ujson

### LINKED ISSUE #2736: Decoding record with > 64-bit integer fails
If a record is added to Kinto that includes a number greater than 64 bits, one cannot view the record via the
 GET /buckets/(bucket_id)/collections/(collection_id)/records/(record_id)
endpoint.

For instance if the record contains {"weight": 211230200427052203004205123} , the numeric value is stripped, rendering the resultant JSON as invalid so that it can't be decoded.

From a quick test, I saw that the issue is related to ujson usage; the same error does not occur with the native json module, nor with pysimdjson which like ujson is advertised as being extremely performant.

damilgra:
Fixes #2736


kinto/core/storage/memory.py
```diff
@@ -13,7 +13,8 @@
     StorageBase,
     exceptions,
 )
-from kinto.core.utils import COMPARISON, find_nested_value, json
+from kinto.core.utils import COMPARISON, find_nested_value
+import json
```
leplatrem, line 17: Couldn't you import `json` from `kinto.core.utils` here too?

damilgra: Right now, kinto.core.utils redefines json as simdjson, so no. Currently I have:
simdjson for PostgreSQL, and native json for Memory 

Not sure whether or not a more performant (than native) json library is needed but that's the rationale. Alternatively it's possible simdjson could be used for both storages but the way records are added via memory storage would need to be modified which is a far more invasive change.

leplatrem: The usage of `json`  in the memory backend is rather straightforward. What issues would we have if we would use `simdjson` there too?

damilgra: The problem is that the unit test I introduced will raise an exception here (found in core/storage/memory.py):

> obj = json.loads(json.dumps(obj))

The reason is that simdjson cannot deserialize strings representing numbers > 64 bits, as per https://github.com/simdjson/simdjson/issues/167. On the other hand, it *can* serialize a > 64-bit number.

leplatrem: Ok, gotcha.

And then I would suggest that once we merge this, you create an issue to keep track of this so that we can use simdjson everywhere once this is fixed upstream.

(we could also do something like `
Then what do you think of something like `obj = simdjson.loads(json.dumps(obj))` but I don't think it brings a lot of value

damilgra: Sure thing. I'll create an issue once it's merged and I've subscribed to the simdjson issue to track progress there.

TkTech:
> The reason is that simdjson cannot deserialize strings representing numbers > 64 bits, as per simdjson/simdjson#167. On the other hand, it can serialize a > 64-bit number.

Funny enough, this is because pysimdjson's `dump()` and `dumps()` are just aliases to Python's built in `json` module. The underlying simdjson library doesn't do serialization at all (yet).

damilgra:
> 
> Funny enough, this is because pysimdjson's `dump()` and `dumps()` are just aliases to Python's built in `json` module. The underlying simdjson library doesn't do serialization at all (yet).

Thanks very much, good to know. I'm going to investigate other options.

kinto/core/utils.py
```diff
@@ -32,8 +32,18 @@
     memcache = None
 
 
-def json_serializer(v, **kw):
-    return json.dumps(v, escape_forward_slashes=False)
+class json:
+    def dumps(v, **kw):
+        return rapidjson.dumps(v, bytes_mode=rapidjson.BM_NONE)
```
leplatrem, line 37: Dropping the `kw` will introduce breaking changes... See https://github.com/Kinto/kinto/blob/b9b4eb047ff0edb1d70a95a14c8afc9630c30d7d/kinto/plugins/quotas/utils.py#L6

damilgra: Perhaps I'm misunderstanding but the file you reference imports stdlib json so it shouldn't be affected. I will make a change however to respect passed in kw, if any; grepping the source I don't see any instances where this happens right now.

import json

def record_size(record):
    # We cannot use ultrajson here, since the `separator` option is not available.
    canonical_json = json.dumps(record, sort_keys=True, separators=(",", ":"))
    return len(canonical_json)

leplatrem:
@damilgra Are you still motivated to push this to the finish line?

kinto/core/utils.py
```diff
@@ -32,8 +32,24 @@
     memcache = None
 
 
-def json_serializer(v, **kw):
-    return json.dumps(v, escape_forward_slashes=False)
+class json:
+    def dumps(v, **kw):
+        if "bytes_mode" not in kw:
+            kw["bytes_mode"] = rapidjson.BM_NONE
```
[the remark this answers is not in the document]
damilgra: Thanks. I've made that change.

damilgra:
> @damilgra Are you still motivated to push this to the finish line?

Sure, anything else I need to do? I'm trying to get a tox environment set up, but if memory serves there were some failing tests unrelated to changes I made. Please let me know if you have any other changes you'd like me to make.

leplatrem (approved):

PR #2883, opened 2021-10-15
Update Python versions (fixes #2881)

### LINKED ISSUE #2881: Make a plan to drop Python 3.6 compatability after it reaches EOL
Python 3.6 [is scheduled](https://devguide.python.org/#status-of-python-branches) to reach end-of-life (EOL) status later this year on 2021-12-23. We should consider if/when we want to follow suit and drop 3.6 compatibility from Kinto.

leplatrem:
Fixes #2881

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        toxenv: [py36, py37, py38, py38-raw, py39]
+        toxenv: [py37, py38, py38-raw, py39]
         include:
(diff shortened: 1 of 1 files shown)


leplatrem:
Shall we remove all mentions of 3.6? or shall we leave it until it blocks us?

Natim:
> Shall we remove all mentions of 3.6? or shall we leave it until it blocks us?

That's a good question. I believe we should remove all mention of it. If people are stuck on 3.6 they will be able to use it at their own risk. However sending the message that we don't support it helps for them to find a plan forward I think.

PR #2921, opened 2021-11-16
coveralls.io support

slav0nic:
Added test coverage to  github-actions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,7 +69,7 @@ jobs:
           source .venv/bin/activate
           make install-dev
           make install-postgres
-          pip install tox
+          pip install tox coveralls
 
       - name: Create database
         env:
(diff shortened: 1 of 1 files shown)


slav0nic:
what's the point of using `virtualenv` in github-actions?

leplatrem:
> what's the point of using virtualenv in github-actions?

Not sure. None? Maybe we could get rid of it (another PR tho)

.github/workflows/test.yml
```diff
@@ -81,3 +81,27 @@ jobs:
         run: |
           source .venv/bin/activate
           tox -e ${{ matrix.toxenv }}
+      
+      - name: Coveralls Parallel
```
leplatrem, line 85: If that's possible, Instead of `Parallel`, I would use something like `Coveralls for ${{ matrix.toxenv }}`

leplatrem (approved):

.github/workflows/test.yml
slav0nic: oke, done

PR #2922, opened 2021-11-16
Remove duplicated dependencies

leplatrem:
It seems better to only list dependencies that are used in `kinto/plugins/admins/src/index.js`


grahamalama (approved):
✂️

PR #2923, opened 2021-11-16
Remove virtualenv step from github-actions

slav0nic:
.


slav0nic:
oops, current `make` commands based on virtualenv, so it's necessary

PR #2939, opened 2022-01-13
Removal of contributor

DrFaustie:
Would like to remove name from contributor list


Natim:
Why is that? Is it your name?

PR #2966, opened 2022-03-18
Build Kinto Admin from source

grahamalama:
Instead of using the `<KintoAdmin />` React component and mounting that in a `create-react-app` project, when we build `kinto`, we download the source code of [`kinto-admin`](https://github.com/Kinto/kinto-admin), build it, and move the build folder to the plugin.


scripts/build-kinto-admin.sh
```diff
@@ -0,0 +1,22 @@
...
+# download and unzip release
+# TODO: error handling (404 error, can't unzip tarball, etc)
+cd kinto/plugins/admin/
+curl -L https://github.com/Kinto/kinto-admin/archive/refs/tags/v${TAG}.tar.gz --output $TARBALL_NAME
+tar -xf $TARBALL_NAME && rm $TARBALL_NAME
+
+# build kinto-admin bundle
+cd $SRC_DIR
+npm ci
+export SINGLE_SERVER=1
```
grahamalama, line 16: Instead of setting the server here, I was thinking that this could just be a flag so that in kinto-admin, we have something like:
```js
export const SINGLE_SERVER = process.env.SINGLE_SERVER
  ? document.location.toString().split("/admin/")[0]
  : null;
```

leplatrem: Yes I don't think we have a way to know the final URL in advance anyway

leplatrem (commented):
Great :100: 

You cannot measure the joy of having one `package-lock.json` less in my life :heart_eyes_cat:

scripts/build-kinto-admin.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+TAG=$(cat kinto/plugins/admin/VERSION)
```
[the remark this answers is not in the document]
grahamalama: Oddly, though the URL for the release contains the `v`, the unzipped tarball doesn't. Here's what I'll do:
```bash
#!/bin/bash
VERSION=$(cat kinto/plugins/admin/VERSION)
TAG="v${VERSION}"
TARBALL_NAME="${TAG}.tar.gz"
SRC_DIR="kinto-admin-${VERSION}"

cd kinto/plugins/admin/
curl -OL https://github.com/Kinto/kinto-admin/archive/refs/tags/${TAG}.tar.gz
tar -xf $TARBALL_NAME && rm $TARBALL_NAME

cd $SRC_DIR
...
```

leplatrem (approved):
:partying_face:

PR #2981, opened 2022-04-13
Prevents merging of `ResourceChanged` events if they are emitted in cascade

leplatrem:
This wlil change the way cascading events are grouped together. 

Before, we would group all cascading events together, as long as they were for the same (resource name, parent id, action). When grouping events, some information is lost (we keep only payload and merge the list of impacted objects), including the `user_id` that triggered the event. This caused us issues because we were unable to distinguish events triggered by the plugin itself, from events triggered by users. See mozilla/remote-settings#203

Now, we prevent grouping of cascading events. Events triggered from events listeners won't be grouped with upstream events. To accomplish this, we increment a cascade_level state variable when we start consuming the list of events to be triggered, and use it in the grouping key.

### DIFF
--- a/kinto/core/events.py
+++ b/kinto/core/events.py
@@ -94,7 +94,7 @@ class EventCollector(object):
         self.event_dict = OrderedDict()
         """The events as collected so far.
 
-        The key of the event_dict is a triple (cascade_level, resource_name,
+        The key of the event_dict is a quadruple (cascade_level, resource_name,
         parent_id, action). The value is a triple (impacted, request,
         payload). If the same (cascade_level, resource_name, parent_id, action) is
         encountered, we just extend the existing impacted with the new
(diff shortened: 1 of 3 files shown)


kinto/core/events.py
```diff
@@ -84,13 +84,19 @@ class EventCollector(object):
...
+        """Current level of event cascade. When we start consuming the
+        gathered events, we increment it. This way, events emitted from
+        events listeners (cascade) are not merged with upstream ones.
+        """
+
         self.event_dict = OrderedDict()
         """The events as collected so far.
 
-        The key of the event_dict is a triple (resource_name,
+        The key of the event_dict is a triple (cascade_level, resource_name,
```
grahamalama, line 97: > The key of the event_dict is a triple (cascade_level, resource_name,

Would it be a `quad` now?

grahamalama (approved):
🚢

PR #3051, opened 2022-10-06
Add Cache-Control on anonymous responses of root URL

leplatrem:
- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ This document describes changes between each past release.
 
 **New features**
 
-- Add ``Cache-Control`` response header in root URL endpoint (``/v1/``) when no authentication request header is provided.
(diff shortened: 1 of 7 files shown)


smarnach (approved):
Looks good to me. Some Selenium UI test is failing, but that looks unrelated.

leplatrem:
> Some Selenium UI test is failing, but that looks unrelated.

It was actually related. 

If the first request to the server was anonymous, the client would cache it and would not fetch a fresh response once authenticated.

I decided to change the criteria and put the condition on whether the instance was configured as readonly

smarnach (approved):
Good catch! Restricting this to read-only instances sounds good to me. Alternatively we could could add `Vary: Authorization`, but that has the potential to interfere with cacheability in CDN, and also result in a small amount of additional data, so I think your solution is better.

PR #3262, opened 2023-09-05
Add explicit Sentry logging integration

leplatrem:
Since https://github.com/Kinto/kinto/pull/3055 the logging messages do not seem to be sent to Sentry anymore.

This adds the logging integration explicitly, and adds settings to adjust levels

### DIFF
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -29,6 +29,7 @@ except ImportError:  # pragma: no cover
     ProfilerMiddleware = False
 try:
     import sentry_sdk
+    from sentry_sdk.integrations.logging import LoggingIntegration
     from sentry_sdk.integrations.pyramid import PyramidIntegration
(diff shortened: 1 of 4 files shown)


kinto/core/initialization.py
```diff
@@ -284,11 +285,19 @@ def setup_sentry(config):
...
         sentry_sdk.init(
             dsn,
             integrations=[
                 PyramidIntegration(),
                 SqlalchemyIntegration(),
+                LoggingIntegration(
+                    # Capture debug and above as breadcrumbs
+                    level=breadcrumbs_level,
+                    # Send warnings and above as events
+                    event_level=events_level,
```
grahamalama, line 299: Nit, but these comments are only true if the default settings are overridden. Should we instead explain what `level` and `event_level` do, and then say "defaults to `<whatever>`"?

grahamalama (approved):

PR #3263, opened 2023-09-08
Keep warning level for server-side logging

leplatrem:
We shouldn't use warnings for concerns that apply to the request. 

Warnings should be used to warn developers/admins of potential issues on the service, not on malformed requests

### DIFF
--- a/kinto/core/errors.py
+++ b/kinto/core/errors.py
@@ -205,7 +205,7 @@ def request_GET(request):
     except UnicodeDecodeError:
         querystring = request.environ.get("QUERY_STRING", "")
         logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


kinto/core/errors.py
```diff
@@ -205,7 +205,7 @@ def request_GET(request):
     except UnicodeDecodeError:
         querystring = request.environ.get("QUERY_STRING", "")
         logger = logging.getLogger(__name__)
-        logger.warning("Error decoding QUERY_STRING: %s" % request.environ)
+        logger.info("Error decoding QUERY_STRING: %s" % request.environ)
```
leplatrem, line 208: This one should even be debug maybe?

grahamalama (approved):

PR #3308, opened 2023-12-04
Group all Dependabot pull-requests

leplatrem:
* [x] Removed obsolete ignore entries
* [x] Removed obsolete NPM entries for the Admin UI  (See #2966 )
* [x] Grouped all kind of Dependabot pull-requests
* [ ] ~~Split devs dependencies~~ will be done in future PR when modernizing repo
* [ ] ~~Split docs dependencies~~ will be done in future PR when modernizing repo

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,52 +5,14 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 99
-  ignore:
-  - dependency-name: sqlalchemy
-    versions:
-    - 1.3.23
-    - 1.4.2
-    - 1.4.3
-    - 1.4.5
-    - 1.4.7
-  - dependency-name: cornice
-    versions:
-    - 5.2.0
-  - dependency-name: docutils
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,52 +5,14 @@ updates:
...
-    - 4.0.0
-  - dependency-name: react
-    versions:
-    - 17.x
-  - dependency-name: react-dom
-    versions:
-    - 17.x
+  groups:
+    all-dependencies:
+      update-types: ["major", "minor", "patch"]
```
grahamalama, line 10: What's our rationale for grouping all dependencies together, vs say keeping major releases out of the groups so we have 1 PR per major release?

I think a major release would "spoil" an update group if it had breaking changes and caused CI to fail, for example

leplatrem: I'm happy separating majors yes, that makes sense. Dev and docs deps can have different rules (I'll look at this once migrated to pyproject)

PR #3343, opened 2024-01-11
Add a new ``kinto.admin_assets_path`` to specify the location of the Admin UI assets

leplatrem:
- [x] Add documentation.
- [x] Add tests.
- [x] Add a changelog entry.


alexcottner:
I think we need to adjust this logic as well: https://github.com/Kinto/kinto/blob/introduce-admin-ui-location-setting/kinto/plugins/admin/__init__.py#L26-L28

leplatrem:
Nice catch! I wouldn't have missed it

PR #3353, opened 2024-01-22
Generate changelog from Github releases

leplatrem:
With this change, we won't need to update the `CHANGELOG.rst` file before each release, and will allow us to tag new releases from the Github releases page (with one-click 😎 )

Screenshot 2024-01-22 at 14 19 10


Natim (approved):
Well done!

PR #3372, opened 2024-01-29
Fix Docker image build frozen step

leplatrem:
* Use an intermediary folder instead of build on container's root folder `/`
* Do not install as editable within the container
* Explicitly list non `.py` as package data
* Fix setuptools find package to include subpackages (`kinto.core` etc.)


alexcottner (approved):
This fixes the docker build issue for me

PR #3374, opened 2024-01-30
Playwright

### LINKED ISSUE #3341: Investigate selenium test errors
We've had some recent inconsistent test runs with browser-tests where selenium will fail due to stale DOM elements. Initial investigations couldn't find a good solution so we have a time.sleep(1) line in the browser tests now. Locate and fix this the root cause.

alexcottner:
Fixes #3341 - Replacing selenium browser-test with playwright

This replaces the existing browser test with a very simple test that verifies we can login to the site and see the expected home page. We will add more detailed testing to the kinto-admin project.

- [x] Add documentation.
- [x] Add tests.

I propose we make the same change in remote-settings, then start working on a comprehensive test suite in the kinto-admin project. Created https://github.com/Kinto/kinto-admin/issues/3147 as followup.

This was done in node in an attempt to keep our UI testing code consistent. However, if we prefer to do that in python (since this is a python project) I can change it pretty easily.


grahamalama:
Nice! A step in the right direction.

Was there a reason you opted for the Javascript version of playwright instead of [the Python version](https://playwright.dev/python/)? It'd be nice to stick to one language in this repo

alexcottner:
> Was there a reason you opted for the Javascript version of playwright instead of [the Python version](https://playwright.dev/python/)? It'd be nice to stick to one language in this repo

I forgot to mention that in the PR body. I can change this to python pretty easily, my initial thought was keeping the testing framework consistent was slightly more beneficial. But I'm ~51/49 on that particular fence and it's not like the tests in here or remote-settings will be all that complex.

grahamalama:
>  my initial thought was keeping the testing framework/language consistent 

Here do you mean consistent within different repos? Like we'll be using Playwright here (a Python project) and Kinto Admin (a JS project), so it's better to have Playwright tests be consistently written than match the predominant language in a repo?

alexcottner:
> Here do you mean consistent across different repos?

Correct, I thought that having a consistent pattern there was slightly better than switching to fit the language of the project. I will hopefully have a python version up tomorrow for comparison. It'll still be pretty simple, just got sidetracked today.

leplatrem:
Yes, I agree with Graham. I'm sorry for the lost energy, but for the convenience of contributors it would be better to spare the usage of `npm`

PR #3420, opened 2024-06-14
Resolves #3419 : Taking into account the timezone in the as_epoch method.

### LINKED ISSUE #3419: Invalid last_modified value on a timezone other than UTC
Hello,

We're using kinto with a database using the Europe/Paris timezone, and we've noticed that the bump_timestamp function calculates a value in the future for the last_modified property.
The cause is that the timezone is not taken into account in the as_epoch function. An erroneous timestamp is obtained when the result of the as_epoch function is passed as a parameter to the from_epoch function.

Example : 

```
kinto=# SELECT clock_timestamp(), 
              clock_timestamp()::TIMESTAMP as timestamp,
              from_epoch(as_epoch(clock_timestamp()::TIMESTAMP)) as "timestamp 1 conversion",
              from_epoch(as_epoch(from_epoch(as_epoch(clock_timestamp()::TIMESTAMP)))) as "timestamp 2 conversions";

        clock_timestamp        |         timestamp          | timestamp 1 conversion | timestamp 2 conversions 
-------------------------------+----------------------------+------------------------+-------------------------
 2024-06-14 18:53:38.949726+02 | 2024-06-14 18:53:38.949726 | 2024-06-14 20:53:38.95 | 2024-06-14 22:53:38.95
(1 row)
```

Best regards.

slafon51:
Fixes #3419 by taking into account the timezone in the as_epoch method.

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -79,7 +79,7 @@ class Storage(StorageBase, MigratorMixin):
 
     # MigratorMixin attributes.
     name = "storage"
-    schema_version = 22
+    schema_version = 23
     schema_file = os.path.join(HERE, "schema.sql")
     migrations_directory = os.path.join(HERE, "migrations")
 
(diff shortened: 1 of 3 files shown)


leplatrem:
Supporting other timezones than UTC introduces some risks and little benefits. I'll close the PR now, and we can discuss more in https://github.com/Kinto/kinto/issues/3419

PR #3461, opened 2024-10-24
Observe request summary, size, and duration with labels

leplatrem:
Based on #3459 

@ahoneiser Now that Kinto has Prometheus support, we can count things with more details. Do you have any suggestion?

### DIFF
--- a/kinto/__init__.py
+++ b/kinto/__init__.py
@@ -38,6 +38,7 @@ DEFAULT_SETTINGS = {
     "record_id_generator": "kinto.views.RelaxedUUID",
     "project_name": "kinto",
     "admin_assets_path": None,
+    "metrics_matchdict_fields": ["bucket_id", "collection_id", "group_id", "record_id"],
 }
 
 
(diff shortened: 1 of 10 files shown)


ahoneiser:
@leplatrem oh that's super nice, great work! From a perspective of determining that kinto is alive and well, `__heartbeat__` and `__lbheartbeat__` do a good job already. 

Thinking about counting things, counting requests to individual buckets / collections / records would be a nice thing to have, e.g. to identify high volume resources and low volume resources, would help when doing housekeeping work (e.g. cleanups / maintenance), counts of requests / gauges of request sizes could also help us to put price tags to defined workloads (e.g. to be able to better answer questions like "how much would it cost to distribute resource x via kinto", or "how much cost do I produce when I require all clients to re-download resource x" etc.).

leplatrem:
Related https://github.com/mozilla/remote-settings/pull/691

alexcottner (approved):

PR #3502, opened 2025-02-21
Remove obsolete sections from docs (fixes #307, fixes #330, fixes #508)

### LINKED ISSUE #307: scalability/deployement comparisons
There's one thing that's not mentioned in the docs: how Kinto _scales_ compared to other existing solutions / and how it's deployed.

I am not sure how we can add this in the doc, but for instance, Kinto is easier to deploy than CouchDB when you have thousands of users because you can deploy a single Postgres server.

Maybe a page that explains a typical Kinto deployement vs typical deployements with other stacks

### LINKED ISSUE #330: Enhance the comparison to compare products
Currently, it's not really clear if we're comparing servers or complete solutions (client + server). We need to be clearer about what Kinto + Kinto.js can do, compared to other solutions such as Couch+Pouch or Hoodie, parse etc.

### LINKED ISSUE #508: The Github oauth tutorial is not OpenID Connect compliant.

leplatrem:
I tried to remove all documentation pieces that don't reflect the reality.

fixes #307, fixes #330, fixes #508


alexcottner (approved):
Excellent

PR #3505, opened 2025-02-21
Remove accounts emailing features

leplatrem:
Like in #3503, I would like to drop the code that is not crucial and that we don't use in production.
Validation, emailing, password reset features represent a lot of code that we can get rid of it IMO. 
We can drop a line in the changelog showing a migration procedure that would consist in copying the `kinto/plugins/accounts` code from previous kinto version and drop it in a module available in the `PYTHONPATH`


tests/plugins/test_accounts.py
```diff
@@ -217,332 +155,6 @@ def test_authentication_refresh_the_cache_each_time_we_authenticate(self):
...
-            status=401,
-        )
-        # Can't use the reset password to modify other resources than accounts.
-        resp = self.app.post_json(
-            "/buckets/default/collections",
-            {"data": {"id": "some_collection_id"}},
-            headers=get_user_headers("alice@example.com", reset_password),
-            status=401,
-        )
-        assert resp.json["message"] == "Please authenticate yourself to use this endpoint."
```
alexcottner, line 446: Looks like this test was the only only covering https://github.com/Kinto/kinto/blob/drop-accounts-emailing-support/kinto/core/views/errors.py#L24-L31

alexcottner (changes requested):
Lost the test covering https://github.com/Kinto/kinto/blob/drop-accounts-emailing-support/kinto/core/views/errors.py#L24-L31 .

PR #3515, opened 2025-03-27
Add option to disable history for certain users

leplatrem:
If certain users like machine-to-machine jobs generate a lot of history that users never look at, then we can choose to just exclude them with this (cheap to implement) option.

I chose to trim the history by resource_name (ie. type of objects) instead of by URI.
This way, if a user deletes the whole collection content and recreates a lot of new records, the history will be trimmed. Whereas by URI, we would keep 2 entries for each record URI, and never trim anything.

### DIFF
--- a/docs/api/1.x/history.rst
+++ b/docs/api/1.x/history.rst
@@ -202,13 +202,13 @@ Or certain users:
 
     kinto.history.exclude_user_ids = account:quicksuggest
 
-In order to limit the size of the history length of each object:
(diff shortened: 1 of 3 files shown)


leplatrem:
Instead of just disabling history which could be weird when inspecting stuff in Admin UI, I propose that we introduce:

```ini
kinto.history.auto_trim_user_ids = account:experimenter
kinto.history.auto_trim_max_count = 100
```

alexcottner (approved):

PR #3516, opened 2025-03-31
Partitioning objects table and adding a new index

alexcottner:
This PR aims to improve performance on the postgres backend by:
1. Partitioning the objects table. But only declaring the default partition, allowing implementers to make their own partitions.
2. Creating a new index to improve checks against deleted records (or looking specifically for non-deleted records)

A migration script is included to transfer existing data from `objects` over to the new partitioned table before dropping the old table.

With the current data for remote-settings, this cuts the execution cost of our common queries in half.

### DIFF
--- a/kinto/core/storage/postgresql/migrations/migration_022_023.sql
+++ b/kinto/core/storage/postgresql/migrations/migration_022_023.sql
@@ -9,8 +9,7 @@ CREATE TABLE IF NOT EXISTS objects_partitioned
     CONSTRAINT objects_pkey PRIMARY KEY (id, parent_id, resource_name)
 ) PARTITION BY LIST(resource_name);
 
-CREATE TABLE objects_default PARTITION OF objects_partitioned 
-    FOR VALUES IN ('', 'account', 'bucket', 'collection', 'group');
+CREATE TABLE objects_default PARTITION OF objects_partitioned DEFAULT;
 CREATE TABLE objects_record PARTITION OF objects_partitioned 
     FOR VALUES IN ('record');
 CREATE TABLE objects_history PARTITION OF objects_partitioned
--- a/kinto/core/storage/postgresql/migrations/migration_022_023.sql
+++ b/kinto/core/storage/postgresql/migrations/migration_022_023.sql
@@ -10,10 +10,6 @@ CREATE TABLE IF NOT EXISTS objects_partitioned
 ) PARTITION BY LIST(resource_name);
 
 CREATE TABLE objects_default PARTITION OF objects_partitioned DEFAULT;
-CREATE TABLE objects_record PARTITION OF objects_partitioned 
-    FOR VALUES IN ('record');
-CREATE TABLE objects_history PARTITION OF objects_partitioned
-    FOR VALUES IN ('history');
 
 INSERT INTO objects_partitioned(id, parent_id, resource_name, last_modified, data, deleted)
 SELECT id, parent_id, resource_name, last_modified, data, deleted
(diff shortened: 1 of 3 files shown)


kinto/core/storage/postgresql/schema.sql
```diff
@@ -47,7 +47,8 @@ CREATE UNIQUE INDEX IF NOT EXISTS idx_objects_parent_id_resource_name_last_modif
     ON objects(parent_id, resource_name, last_modified DESC);
 CREATE INDEX IF NOT EXISTS idx_objects_last_modified_epoch
     ON objects(as_epoch(last_modified));
-
+CREATE INDEX IF NOT EXISTS idx_objects_resource_name_parent_id_deleted
+    ON objects(resource_name, parent_id, deleted);
```
leplatrem, line 51: You would need to add a DB migration. See example: https://github.com/Kinto/kinto/pull/1436/files

leplatrem (changes requested):

kinto/core/storage/postgresql/migrations/migration_022_023.sql
```diff
@@ -0,0 +1,50 @@
...
+    FOR EACH ROW
+    EXECUTE FUNCTION bump_timestamp();
+
+ALTER INDEX idx_objects_last_modified_epoch RENAME TO idx_objects_last_modified_epoch_old;
+ALTER INDEX idx_objects_parent_id_resource_name_last_modified RENAME TO idx_objects_parent_id_resource_name_last_modified_old;
+ALTER INDEX idx_objects_last_modified_epoch_partitioned RENAME TO idx_objects_last_modified_epoch;
+ALTER INDEX idx_objects_parent_id_resource_name_last_modified_partitioned RENAME TO idx_objects_parent_id_resource_name_last_modified;
+
+CREATE INDEX IF NOT EXISTS idx_objects_resource_name_parent_id_deleted
+    ON objects(resource_name, parent_id, deleted);
```
leplatrem, line 45: Can we add this index in a different PR so that we can add it and decide later whether we do the partitioning?

kinto/core/storage/postgresql/schema.sql
```diff
@@ -22,32 +22,29 @@ IMMUTABLE;
...
+    data jsonb NOT NULL DEFAULT '{}'::jsonb,
+    deleted boolean NOT NULL DEFAULT false,
+    CONSTRAINT objects_pkey PRIMARY KEY (id, parent_id, resource_name)
+) PARTITION BY LIST(resource_name);
+
+CREATE TABLE objects_default PARTITION OF objects DEFAULT;
+CREATE TABLE objects_record PARTITION OF objects
+    FOR VALUES IN ('record');
+CREATE TABLE objects_history PARTITION OF objects
+    FOR VALUES IN ('history');
```
leplatrem, line 40: `kinto.core` is not "supposed" to be aware of notions of bucket, collections, groups, records, etc. We only manipulate generic resources.

Ideally this would be a setting (although I know we don't have templating in migrations) because in some kinto instances there could be many groups but few records. Or the history plugin not installed.

leplatrem (commented):
Thank you, Alex, for jumping on this so quickly!

The performance improvements are definitely appealing, but there are still a few details I'd like to iron out before moving forward.

> With the current data for remote-settings, this cuts the execution cost of our common queries in half.

We could split the index creation first and see where we land before bringing the heavy artillery :) See https://github.com/Kinto/kinto/pull/3516

kinto/core/storage/postgresql/migrations/migration_022_023.sql
leplatrem: Done in #3519

kinto/core/storage/postgresql/schema.sql
alexcottner: I could have this only create the default partition and allow implementations to create their own.

leplatrem: that's a good idea

alexcottner:
Closing for now as we've reduced our table size by purging old records.

PR #3524, opened 2025-04-08
Add command to purge tombstones

leplatrem:
Since our server is configured with `paginate_by=None` and `storage.max_fetch_limit=10000`, it does not make sense to keep more than 10K tombstones for each collection.

Some of our collection in production (after 10 years of sync!) have more than 22K tombstones 🤣 

See https://github.com/Kinto/kinto/pull/990 for a jump in the past 😅 

* [ ] Basic rebase
* [ ] Import code to add more parameters (collection, buckets)
* [ ] Add docs, etc.

### DIFF
--- a/kinto/__main__.py
+++ b/kinto/__main__.py
@@ -33,6 +33,7 @@ def main(args=None):
         "flush-cache",
         "version",
         "create-user",
+        "purge-deleted",
     )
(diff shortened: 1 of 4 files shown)


alexcottner (approved):
10k tombstones even seems like a lot

PR #3525, opened 2025-04-10
Enable Prometheus multiprocess collector

leplatrem:
While talking to Sven about metrics this morning, he pointed out the fact that we should use a multiprocess collector, because we run `uwsgi` with `workers = X` and `master = true`  ([config](https://github.com/mozilla-it/webservices-infra/blob/429ecad27508f3d0ccd05a31c0eec1adbfe2263b/remote-settings/k8s/remote-settings/templates/configmap.yaml#L406C1-L434C1))

If I'm not mistaken, this should be enough

### DIFF
--- a/kinto/plugins/prometheus.py
+++ b/kinto/plugins/prometheus.py
@@ -187,8 +187,8 @@ def metrics_view(request):
     return resp
 
 
-def _reset_multiproc_folder_content():
-    if os.path.exists(PROMETHEUS_MULTIPROC_DIR):  # pragma: no cover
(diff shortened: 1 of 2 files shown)


kinto/plugins/prometheus.py
```diff
@@ -170,6 +184,12 @@ def metrics_view(request):
     return resp
 
 
+def _reset_multiproc_folder_content(path):
```
alexcottner, line 187: Probably need to add `# pragma: no cover`

alexcottner (approved):
Code coverage dropped, but should be good after adding a `no cover` comment

PR #3534, opened 2025-04-23
Prevent random values to become metrics labels values

leplatrem:
The internet is wild... and bloat our metrics labels values:

https://github.com/user-attachments/assets/47c0aaa3-16a8-421c-a2f1-7e12cf31b80d

### DIFF
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -472,10 +472,18 @@ def setup_metrics(config):
             auth, user_id = user_id.split(":")
             metrics_service.count("users", unique=[("auth", auth), ("userid", user_id)])
 
+        status = event.response.status_code
+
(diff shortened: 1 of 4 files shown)


kinto/core/initialization.py
```diff
@@ -472,10 +472,18 @@ def on_new_response(event):
             auth, user_id = user_id.split(":")
             metrics_service.count("users", unique=[("auth", auth), ("userid", user_id)])
 
+        status = event.response.status_code
+
+        if status >= 400:
+            # Prevent random values of 404 responses to become label values.
+            request_matchdict = {}
```
alexcottner, line 479: An empty match seems reasonable to me. We could also specify `invalid` for the bucket and collection.

PR #3541, opened 2025-05-07
Remove `status` from request_size and request_duration_seconds labels

leplatrem:
This does not seem to be reasonable:

```
$ curl -s https://remote-settings.allizom.org/v1/__metrics__ | wc -l
   26661
$ curl -s https://remote-settings.allizom.org/v1/__metrics__ | grep remotesettings_request_duration | wc -l
   20761
```

This PR adds the `_seconds` suffix to the request duration metric, and removes the status label from size and duration metrics. This will divide the cardinality by a factor of at least 2 (200, 201)

### DIFF
--- a/kinto/core/initialization.py
+++ b/kinto/core/initialization.py
@@ -510,17 +510,16 @@ def setup_metrics(config):
         request_labels = [
             ("method", request.method.lower()),
             ("endpoint", endpoint),
-            ("status", str(status)),
         ] + metrics_matchdict_labels
 
         # Count served requests.
-        metrics_service.count("request_summary", unique=request_labels)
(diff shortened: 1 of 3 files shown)


alexcottner (approved):
This should reduce the counts. We can also pass in some custom buckets [here](https://github.com/Kinto/kinto/blob/main/kinto/plugins/prometheus.py#L121-L126), we probably don't need much precision.

Ex:
```
 _METRICS[key] = prometheus_module.Histogram(
    _fix_metric_name(key),
    f"Histogram of {key}",
    registry=get_registry(),
    labelnames=[label_name for label_name, _ in labels],
    buckets=[0.1, 1, 3, 6, +Inf]
)
```

Issue #872, opened 2016-10-20
Switch from colander to jsonschema?


leplatrem:
Now that Cornice 2.0 supports any kind of validation, we could switch from colander to jsonschema, since it's already a dependency.

Pros: 
- remove a dependency
- schemas would be simple dicts, easy to manipulate
- may simplify #631 

Cons:
- jsonschema are verbose
- Lib may not be as solid as colander?
- Validation errors may not be as clear?
- Would break API of `kinto.core`

Related #790

Natim:
I am not against using JSONSchema for that but rather a bit reluctant.
How would it looks like in terms of code separation? Will we create JSON files with the JSONSchema for each kind? (bucket, collection, group, permission, record, history, webpush) and load the file from the Python code?

As you said the biggest impact would be the error messages that would be hard to configure.

I am not sure it worth it to be honest.

leplatrem:
> How would it looks like in terms of code separation? 

Instead of having Colander classes, we would have simple dicts in the source code (not JSON files)

> I am not sure it worth it to be honest.

+1

PR #3546, opened 2025-05-08
Fixing duplicated prometheus metrics. Resolves remote-settings #872

alexcottner:
Fixes Remote-Settings [#872](https://github.com/mozilla/remote-settings/issues/872).

According to the prometheus python multi-process [documentation](https://prometheus.github.io/client_python/multiprocess/), we shouldn't register our metrics with the multi-proc metrics collector.

> The application must initialize a new CollectorRegistry, and store the multi-process collector inside. It is a best practice to create this registry inside the context of a request to avoid metrics registering themselves to a collector used by a MultiProcessCollector. **If a registry with metrics registered is used by a MultiProcessCollector duplicate metrics may be exported, one for multiprocess, and one for the process serving the request.**

:question: I'm not sure what a good test to add for this is. It's in an optional plugin and this seems like a prometheus client config issue to me.

### DIFF
--- a/kinto/plugins/prometheus.py
+++ b/kinto/plugins/prometheus.py
@@ -121,7 +121,6 @@ class PrometheusService:
             _METRICS[key] = prometheus_module.Histogram(
                 _fix_metric_name(key),
                 f"Histogram of {key}",
-                registry=get_registry(),
                 labelnames=[label_name for label_name, _ in labels],
             )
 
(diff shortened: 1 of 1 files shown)


leplatrem:
Oh excellent Alex 🤩 
I don't read docs carefully enough.
I'll merge this, and see if #3542 turns green after a merge of origin/main :)

leplatrem (approved):

PR #3550, opened 2025-05-20
Fix #3386: set appropriate CSP for attachment previews in images

### LINKED ISSUE #3386: Preview thumbnail of attached image file is not shown in UI
Screenshot 2024-02-07 at 14 04 16

The URL is correct but it is blocked by `Content-Security-Policy: The page’s settings blocked the loading of a resource `

See https://github.com/Kinto/kinto/blob/e8e55d19616d67ee1c9f7d7baa6a67986fcaf058/kinto/plugins/admin/views.py#L30-L40

(could have opened this issue but used remote-settiungs

leplatrem:
Fix #3386

### DIFF
--- a/kinto/plugins/admin/views.py
+++ b/kinto/plugins/admin/views.py
@@ -32,7 +32,7 @@ def admin_home_view(request):
     allow_local_only = "; ".join(
         (
             "default-src 'self'",
-            "img-src data: 'self'",
+            "img-src data: *",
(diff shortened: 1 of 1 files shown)


say-yawn (approved):
Using `*` instead of listing domains because the attachments come from different URLs than the URL for Remote Settings. We discussed potential issues arising from removing the content security policy for images, but we agreed that the risk that the image could be injected is extremely.

PR #3595, opened 2025-10-31
Fix performance of history trimming

leplatrem:
Before (`list_all()` + `delete_all()` > 1sec):

```
2025-10-31 18:20:05.365 CET [67005] LOG:  duration: 482.272 ms  statement:
	            SELECT id, as_epoch(last_modified) AS last_modified, data
	            FROM objects
	            WHERE parent_id = '/buckets/main-workspace'
	            AND resource_name = 'history'
	            AND NOT deleted
	            AND (data->'resource_name' IS NOT NULL AND data->'resource_name' = '"record"') AND (data->'user_id' IS NOT NULL AND data->'user_id' = '"account:admin"')

	            ORDER BY last_modified DESC
	            LIMIT 1001;
```

After, 1 atomic operation < 2msec:
```
2025-10-31 18:17:14.461 CET [64622] LOG:  duration: 1.253 ms  statement:
	            WITH to_delete AS (
	                SELECT id
	                FROM objects
	                WHERE parent_id = '/buckets/main-workspace'
	                  AND resource_name = 'history'
	                  AND (data->'resource_name' IS NOT NULL AND data->'resource_name' = '"record"') AND (data->'user_id' IS NOT NULL AND data->'user_id' = '"account:admin"')
	                ORDER BY last_modified DESC
	                OFFSET 1000
	            )
	            DELETE FROM objects o
	            USING to_delete d
	            WHERE o.id = d.id
	            RETURNING 1;
```

### DIFF
--- a/tests/core/test_storage.py
+++ b/tests/core/test_storage.py
@@ -75,6 +75,7 @@ class StorageBaseTest(unittest.TestCase):
             (self.storage.purge_deleted, "", ""),
             (self.storage.list_all, "", ""),
             (self.storage.count_all, "", ""),
+            (self.storage.trim_objects, "", "", [], 0),
         ]
         for call in calls:
             self.assertRaises(NotImplementedError, *call)


alexcottner (changes requested):
Excellent. Just need to fix the schema check in unit tests.

PR #3597, opened 2025-11-04
Optimize `all_resource_timestamps()`

leplatrem:
Image

Cost divided by 3, by adding a partial index on `resource_name = 'record'`.
The rest of changes bring some improvements but not crucial.

Before

Screenshot 2025-11-04 at 17 40 25

https://explain.dalibo.com/plan/2542259f3947cc83#plan/node/6

After 
Screenshot 2025-11-04 at 17 40 41

https://explain.dalibo.com/plan/hd7f0f51e0db8646#plan/node/5

### DIFF
--- a/kinto/core/storage/postgresql/__init__.py
+++ b/kinto/core/storage/postgresql/__init__.py
@@ -79,7 +79,7 @@ class Storage(StorageBase, MigratorMixin):
 
     # MigratorMixin attributes.
     name = "storage"
-    schema_version = 24
+    schema_version = 25
     schema_file = os.path.join(HERE, "schema.sql")
     migrations_directory = os.path.join(HERE, "migrations")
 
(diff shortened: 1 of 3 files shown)


kinto/core/storage/postgresql/__init__.py
```diff
@@ -252,20 +252,20 @@ def all_resources_timestamps(self, resource_name):
         WITH existing_timestamps AS (
           -- Timestamp of latest object by parent_id.
           (
-            SELECT parent_id, MAX(last_modified) AS last_modified
+            SELECT DISTINCT ON (parent_id) parent_id, last_modified
```
alexcottner: I'm surprised this first CTE change helps performance.

kinto/core/storage/postgresql/__init__.py
```diff
@@ -252,20 +252,20 @@ def all_resources_timestamps(self, resource_name):
...
-          UNION
+          UNION ALL
           (
             SELECT parent_id, last_modified
             FROM timestamps
             WHERE resource_name = :resource_name
           )
         )
-        SELECT parent_id, MAX(as_epoch(last_modified)) AS last_modified
+        SELECT parent_id, as_epoch(MAX(last_modified)) AS last_modified
```
alexcottner: This is a good flip

alexcottner (approved):

kinto/core/storage/postgresql/__init__.py
leplatrem: It's minor here I think, because where this is executed we've already done the selection of objects rows, and the number is very small.

It relies on index last modified desc instead of group by + max

PR #3602, opened 2025-11-14
feat: Add offset-based pagination support

sherlock2215:
# Add Offset-Based Pagination Support

## Problem
Currently, Kinto only supports token-based pagination which requires sequential access to records. Users cannot jump directly to specific record positions (e.g., records 100-120) without paging through all previous records.

## Solution
This PR adds support for `_offset` parameter alongside the existing token-based pagination, allowing direct access to any record position.

## Changes
- **New `_offset` parameter**: Allows specifying starting position for records
- **Backward compatible**: Existing token-based pagination continues to work unchanged
- **Enhanced response format**: Includes pagination metadata for offset requests
- **Manual offset handling**: Implements offset slicing for memory backend

## Key Features
1. **Direct Access**: `GET /records?_limit=20&_offset=100` returns records 100-119 directly
2. **Pagination Metadata**: Response includes `pagination` object with limit, offset, total, has_more
3. **Next Page URLs**: Automatically generated with correct offset values
4. **Priority System**: Offset takes precedence over token when both are provided

## API Examples
```bash
# Offset pagination
GET /records?_limit=20&_offset=100

# Response
{
  "data": [...],
  "pagination": {
    "limit": 20,
    "offset": 100,
    "total": null,
    "has_more": true
  },
  "next": "/records?_limit=20&_offset=120"
}

# Backward compatibility maintained
GET /records?_limit=20&_token=abc123  # Still works

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
--- a/kinto/core/resource/__init__.py
+++ b/kinto/core/resource/__init__.py
@@ -16,6 +16,7 @@ from pyramid.httpexceptions import (
     HTTPServiceUnavailable,
 )
 from pyramid.settings import asbool
+from urllib.parse import urlparse, parse_qs, urlencode, urlunparse
 
 from kinto.core import Service
 from kinto.core.errors import ERRORS, http_error, raise_invalid, request_GET, send_alert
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: .idea/vcs.xml [hunk 1]
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
```
leplatrem: Do we want these `.idea/` files? I'd say add it to `.gitiignore`

kinto/core/resource/__init__.py
```diff
@@ -394,20 +431,45 @@ def _plural_get(self, head_request):
...
 
         # See bigger explanation above about the use of limits. The need for slicing
         # here is because we might have asked for 1 more object just to see if there's
         # a next page. But we have to honor the limit in our returned response.
-        return self.postprocess(objects[:limit])
+        result_objects = objects[:limit]
+        if used_offset_pagination:
+            response_data = {
+                'data': result_objects,
+                'pagination': {
```
leplatrem: I'm not sure about this new `pagination` field. If we add it, we must check that it is consistent with everything else, and well documented. There is also the option to use response headers. Kinto responses haven't changed in 10 years 😅

PR #3718, opened 2026-05-29
Fix #2169: Add docstrings checker

### LINKED ISSUE #2169: Question:  Docstring Format

leplatrem:
Fix #2169

Does not make docstrings mandatory, but at least makes sure the ones present are valid.

### DIFF
--- a/kinto/core/cornice/service.py
+++ b/kinto/core/cornice/service.py
@@ -102,7 +102,7 @@ class Service(object):
 
     There are also a number of parameters that are related to the support of
     CORS (Cross Origin Resource Sharing). You can read the
-    :ref:`CORS specification <http://www.w3.org/TR/cors/>`_.
+    `CORS specification <http://www.w3.org/TR/cors/>`_.
 
     :param cors_enabled:
         To use if you especially want to disable CORS support for a particular
--- a/kinto/core/cornice/service.py
+++ b/kinto/core/cornice/service.py
@@ -101,8 +101,8 @@ class Service(object):
         will be used as the traversal path.
 
     There are also a number of parameters that are related to the support of
-    CORS (Cross Origin Resource Sharing). You can read the CORS specification
-    at http://www.w3.org/TR/cors/
+    CORS (Cross Origin Resource Sharing). You can read the
+    :ref:`CORS specification <http://www.w3.org/TR/cors/>`_.
 
     :param cors_enabled:
         To use if you especially want to disable CORS support for a particular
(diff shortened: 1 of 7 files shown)


kinto/core/cornice/service.py
```diff
@@ -139,8 +139,8 @@ class Service(object):
         here.
 
     See
```
alexcottner:
```suggestion
    See the
```

kinto/core/openapi.py
```diff
@@ -26,9 +26,9 @@ class OpenAPI(CorniceSwagger):
     def expose_authentication_method(cls, method_name, definition):
         """Allow security extensions to expose authentication methods on the
         OpenAPI documentation. The definition field should correspond to a
-        valid OpenAPI security definition. Refer the OpenAPI 2.0 specification
+        valid OpenAPI security definition. Refer the
```
alexcottner:
```suggestion
        valid OpenAPI security definition. Refer to the
```

alexcottner (approved):

PR #3721, opened 2026-06-03
Check typos in `make lint`

leplatrem:


kinto/core/storage/postgresql/__init__.py
```diff
@@ -383,8 +383,8 @@ def create(
             # id already exists.
             # Even if this check doesn't find one, be robust against
             # conflicts because we could race with another thread.
-            # Still, this reduces write load because SELECTs are
-            # cheaper than INSERTs.
+            # Still, this reduces write load because SELECTTs are
+            # cheaper than INSERTTs.
```
leplatrem: Is that an english things form plural forms?

alexcottner: No. It's just hilariously wrong. But we could rewrite it like this if that feels better.
```
            # Still, this reduces write load because SELECT 
            # statements are cheaper than INSERT statements.
```

alexcottner (approved):
Looks good except for the weird sql comment suggestion. Left an alternative suggestion, or we can tell typos-cli to ignore it.

PR #3725, opened 2026-06-09
Fix #1789: add more type annotations

### LINKED ISSUE #1789: Type definitions
Hi,

so since I started working on the project very recently I found it a little difficult to follow the flow of the code at times and I think this could largely be alleviated if we included type definitions (at least for the function definitions), that way we should still remain compatible with python3.5 but it makes it a little easier.

We could of course at the same time switch over to f-Strings as mentioned in #1659 and lose support for python 3.5 but also include type definitions for non function parameters

What are your thoughts?

leplatrem:
Fixes #1789

80% done with claude

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -10,3 +10,4 @@ pyramid==2.1
 python-rapidjson==1.23
 SQLAlchemy==2.0.50
 setuptools<82.0
+redis==8.0.0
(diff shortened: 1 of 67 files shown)


docs/requirements.txt
```diff
@@ -10,3 +10,4 @@ pyramid==2.1
 python-rapidjson==1.23
 SQLAlchemy==2.0.50
 setuptools<82.0
+redis==8.0.0
```
alexcottner, line 13: we need redis to build docs?

alexcottner (approved):

docs/requirements.txt
leplatrem: Because of a link to the Client class

PR #3759, opened 2026-08-02
OpenID: invalid auth payload not turned into proper 401

### LINKED ISSUE #3719: OpenID: invalid auth payload not turned into proper 401
When `audience` is passed, the auth server redirects to:

https://server/v1/openid/ldap/token?error=access_denied&error_description=Service%20not%20found%3A%20kinto-nonprod&state=bf9e12556874220....0740ce4

And kinto does not treat this as an error but tries to parse it and returns 400 instead:

```
code	400
errno	107
error	"Invalid parameters"
message	"code in querystring: Required"
```

Kanman345:
Fixes #3719 

- The `Example of login redirections` section in `docs/api/1.x/openid.rst` only documents the happy path. This fix is self-explanatory from the code, when the IdP returns `error=access_denied`, Kinto now forwards it to the callback instead of returning 400.

-  Added `TokenViewErrorTest` in `tests/plugins/test_openid.py` covering:
  Auth server error redirects to the stored callback
  `error_description` is forwarded alongside `error`
  Invalid state with an error still returns 400
  State is consumed (cannot be reused) even on error
  Missing `code` with no `error` still returns 400

### DIFF
--- a/kinto/plugins/openid/views.py
+++ b/kinto/plugins/openid/views.py
@@ -132,14 +132,10 @@ def validate_token_querystring(node: colander.SchemaNode, value: dict) -> None:
     has_code = "code" in value
     has_error = "error" in value
 
-    success_response = has_code
-    error_response = has_error
-
-    if success_response == error_response:
+    if has_code == has_error:
         raise colander.Invalid(
             node,
-            "Provide either 'state' and 'code', "
-            "or 'error' and 'error_description', but not both.",
+            "Provide either 'state' and 'code', or 'error' and 'error_description', but not both.",
         )
 
 
@@ -153,7 +149,7 @@ class TokenQuerystringSchema(colander.MappingSchema):
     error = colander.SchemaNode(colander.String(), missing=colander.drop)
     error_description = colander.SchemaNode(colander.String(), missing=colander.drop)
 
-    validator = validate_token_querystring
+    validator = staticmethod(validate_token_querystring)
 
 
 class TokenSchema(colander.MappingSchema):
@@ -203,14 +199,7 @@ def get_token(request: Request) -> None:
     oid_config = fetch_openid_config(issuer)
     token_endpoint = oid_config["token_endpoint"]
 
-    code = request.GET.get("code")
-    if not code:
-        # No error param and no code — malformed request.
-        error_details = {
-            "name": "code",
-            "description": "code in querystring: Required",
-        }
-        raise_invalid(request, **error_details)
+    code = request.GET["code"]
 
     # State can be used only once.
     callback = request.registry.cache.delete("openid:state:" + state)
--- a/kinto/plugins/openid/views.py
+++ b/kinto/plugins/openid/views.py
@@ -127,7 +127,7 @@ def get_login(request: Request) -> None:
     raise httpexceptions.HTTPTemporaryRedirect(redirect)
 
 
-def validate_token_querystring(node: colander.SchemaNode, value: dict) -> None:
+def validate_token_querystring(node, value):
     """Enforce that exactly one of (code, error) is present, never both."""
     has_code = "code" in value
     has_error = "error" in value
--- a/kinto/plugins/openid/views.py
+++ b/kinto/plugins/openid/views.py
@@ -127,6 +127,22 @@ def get_login(request: Request) -> None:
     raise httpexceptions.HTTPTemporaryRedirect(redirect)
 
 
+def validate_token_querystring(node: colander.SchemaNode, value: dict) -> None:
+    """Enforce that exactly one of (code, error) is present, never both."""
+    has_code = "code" in value
+    has_error = "error" in value
+
+    success_response = has_code
+    error_response = has_error
+
+    if success_response == error_response:
+        raise colander.Invalid(
+            node,
+            "Provide either 'state' and 'code', "
+            "or 'error' and 'error_description', but not both.",
+        )
+
+
 class TokenQuerystringSchema(colander.MappingSchema):
     """
     Querystring schema for the token endpoint.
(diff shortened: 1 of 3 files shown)


kinto/plugins/openid/views.py
```diff
@@ -157,12 +157,40 @@ def get_token(request: Request) -> None:
     client_id = settings[settings_prefix + "client_id"]
     client_secret = settings[settings_prefix + "client_secret"]
 
+    state = request.GET["state"]
+
+    # If the auth server returned an error (e.g. access_denied when audience
+    # is not found), forward it back to the callback URI instead of trying
+    # to trade a code that was never issued.
+    error = request.GET.get("error")
```
leplatrem, line 181: You could add `error` with `missing:drop` to the query string schema.

kinto/plugins/openid/views.py
```diff
@@ -157,12 +157,40 @@ def get_token(request: Request) -> None:
...
+        callback = request.registry.cache.delete("openid:state:" + state)
+        if callback is None:
+            error_details = {
+                "name": "state",
+                "description": "Invalid state",
+                "errno": ERRORS.INVALID_AUTH_TOKEN.value,
+            }
+            raise_invalid(request, **error_details)
+        error_params = {"error": error}
+        error_description = request.GET.get("error_description")
```
leplatrem, line 192: This field too maybe?

kinto/plugins/openid/views.py
```diff
@@ -157,12 +157,40 @@ def get_token(request: Request) -> None:
...
-    code = request.GET["code"]
-    state = request.GET["state"]
+    code = request.GET.get("code")
+    if not code:
+        # No error param and no code — malformed request.
+        error_details = {
+            "name": "code",
+            "description": "code in querystring: Required",
+        }
+        raise_invalid(request, **error_details)
```
leplatrem: Maybe it would be clearer to move this logic to the schema itself? What do you think?

```py

def validate_token_querystring(node, value):
    has_state = "state" in value
    has_code = "code" in value
    has_error = "error" in value

    success_response = has_state and has_code
    error_response = has_error

    if success_response == error_response:
        raise colander.Invalid(
            node,
            "Provide either 'state' and 'code', "
            "or 'error' and 'error_description', but not both.",
        )


class TokenQuerystringSchema(colander.MappingSchema):
     code = colander.SchemaNode(colander.String(), missing=colander.drop)
     state = colander.SchemaNode(colander.String(), missing=colander.drop) 
     error = colander.SchemaNode(colander.String(), missing=colander.drop) 
     error_description = colander.SchemaNode(colander.String(), missing=colander.drop)

     validator = validate_token_querystring
```

leplatrem (approved):
Excellent ! Welcome the project :) 

Great contribution. I have just one comment about moving some of the logics to the schema to keep the view code shorter/clearer, what do you think? 

Thank you for diving into this

kinto/plugins/openid/views.py
Kanman345: All 3 suggestions look good yeah, would be better this way

Kanman345: Regarding this, I decided to keep `state` as **required**, since `state` is consumed in both error and success paths in `get_token`, making it optional would push a `KeyError` into the view

leplatrem:
You need to run `make format` 😉

kinto/plugins/openid/views.py
leplatrem: Yes, that's perfect!

kinto/plugins/openid/views.py
```diff
@@ -157,12 +177,40 @@ def get_token(request: Request) -> None:
...
+        raise httpexceptions.HTTPTemporaryRedirect(redirect)
+
     # Read OpenID configuration (cached by issuer)
     oid_config = fetch_openid_config(issuer)
     token_endpoint = oid_config["token_endpoint"]
 
-    code = request.GET["code"]
-    state = request.GET["state"]
+    code = request.GET.get("code")
+    if not code:
```
leplatrem: I think this piece of code is not necessary.

If we're here, then `error` is falsy. So `code` cannot be falsy, since the schema validation prevents it.

kinto/plugins/openid/views.py
```diff
@@ -127,13 +127,33 @@ def get_login(request: Request) -> None:
     raise httpexceptions.HTTPTemporaryRedirect(redirect)
 
 
+def validate_token_querystring(node: colander.SchemaNode, value: dict) -> None:
+    """Enforce that exactly one of (code, error) is present, never both."""
+    has_code = "code" in value
+    has_error = "error" in value
+
+    success_response = has_code
+    error_response = has_error
```
leplatrem: maybe having this rename is not fully necessary in the end... it would have been if the condition would have been more complex

Kanman345: Sure, can you have a look now and let me know your thoughts?

leplatrem: This looks great! 

But type annotations fail on `make lint`

```
error[invalid-type-form]: Variable of type `_SchemaMeta` is not allowed in a parameter annotation
   --> kinto/plugins/openid/views.py:130:38
    |
130 | def validate_token_querystring(node: colander.SchemaNode, value: dict) -> None:
    |                                      ^^^^^^^^^^^^^^^^^^^
```

leplatrem: We could maybe remove them

Kanman345: Fixed them!

leplatrem:
Yeah 👏 🎉 

Thank you for your patience, the result is great!

PR #3645, opened 2026-02-15
Enable index range scans for last_modified queries (100x speedup in some cases)

sambhav:
## Problem

Kinto has an excellent composite index `(parent_id, resource_name, last_modified DESC)` that can satisfy filtered listings, pagination, and sorting from a single B-tree scan. However, `last_modified` is wrapped in `as_epoch()` in WHERE and ORDER BY clauses, which prevents PostgreSQL from using this index for range scans and sort elimination.

### Current Behavior

**`_format_conditions` (line 883):** When a filter targets `last_modified`, the generated SQL is `as_epoch(last_modified) >= :value`. This wraps the indexed column in a function, which means PostgreSQL cannot use the B-tree on `last_modified` for range scans — it must evaluate `as_epoch()` on every candidate row.

**`resource_timestamp` (line 204):** The ORDER BY is `as_epoch(last_modified) DESC`. Same problem — the index stores `last_modified` in DESC order, but the query asks for `as_epoch(last_modified)` in DESC order. PostgreSQL can't use the index to avoid a sort.

**`bump_timestamp` trigger (schema.sql line 87):** `ORDER BY as_epoch(last_modified) DESC LIMIT 1`. This fires on **every INSERT and UPDATE**. Instead of being a single index point lookup (get the first entry for this parent_id + resource_name from the index), it's doing a filter+sort on the function output.

**`purge_deleted` (line 661):** `as_epoch(last_modified) < :before`. Same index-defeating pattern.

## Solution

The insight is that `as_epoch()` and `from_epoch()` are inverses. Every timestamp in the `objects` table was set via `from_epoch()` (in the trigger and in create/update). So instead of:

```sql
WHERE as_epoch(last_modified) >= :epoch_value   -- function on column: index can't help
```

We write:

```sql
WHERE last_modified >= from_epoch(:epoch_value)  -- raw column: index range scan
```

The `from_epoch(:epoch_value)` is evaluated once (it's a constant for the query), and then PostgreSQL does a standard B-tree range scan on the raw `last_modified` column.

For ORDER BY, it's even simpler — just remove the `as_epoch()` wrapper:

```sql
ORDER BY last_modified DESC LIMIT 1   -- index provides this order directly
```

The SELECT projections (`as_epoch(last_modified) AS last_modified`) stay unchanged — those are output formatting and don't affect index usage.

### `_format_conditions` (kinto/core/storage/postgresql/__init__.py)

[...]

PR #3790, opened 2026-09-18
Fix concurrent timestamp collision in bump_timestamp() (fixes #3673)

### LINKED ISSUE #602: [cliquet] Timestamp uniqueness by parent_id under heavy load

marouan-boumchahate:
- Serialize concurrent writes on the same (parent_id, resource_name) using `pg_advisory_xact_lock` to prevent duplicate key collisions in `MAX(last_modified)` arithmetic.
- Preserves millisecond-precision timestamps, maintaining full compatibility with index range scans from PR #3645.
- Add `migration_026_027.sql` and bump storage schema version to 27.
- Add migration test for schema version 27.
- Remove `@skip_if_ci` from `test_timestamps_are_unique` and increase concurrency to 5 parallel threads to thoroughly stress-test timestamp serialization.
  
Fixes #3673, fixes #602
  
- [ ] Add documentation.
- [X] Add tests.
- [ ] If you changed the HTTP API, update the [API_VERSION](https://github.com/Kinto/kinto/blob/main/kinto/__init__.py#L15) constant and add an API changelog entry [in the docs](https://github.com/Kinto/kinto/blob/main/docs/api/index.rst)
- [ ] If you added a new configuration setting, update the [`kinto.tpl`](https://github.com/Kinto/kinto/blob/main/kinto/config/kinto.tpl) file with it.


marouan-boumchahate:
Could you please add a label to this PR so the Force pull-requests label check passes? Thank you!

leplatrem (approved):
This looks sound from here 👌 

Thank you for your contribution 👏 🙏 

I think we could link / close this 10 years issue with your changes https://github.com/Kinto/kinto/issues/602 🎉 

But first I would suggest that we find a way to add some tests (failing on main, passing on your branch).
 
For example, we should now be able to remove this decorator at least:
https://github.com/Kinto/kinto/blob/6f7ca70b61ff84b571baaaf0e5fb76142bdbdf16/kinto/core/storage/testing.py#L911
And even increase a bit the number of parallel threads 🤷 
This test was passing fine locally. It would be great to find a way to reproduce the conflict deterministically on local too (we sometimes observe it in production).

I remember back in the days, that switching fsync=on/off was affecting our unit tests. I don't know whether the image https://hub.docker.com/_/postgres used in Github Action has anything special that could affect our results.

If you have some ideas, that would be great. 
Did you run a load test? Just to double check that performance is not affected with normal trafic.

If not, we'd have to take some time to do all this before we can merge 😊 

Thanks a lot again for your valuable contribution! It's not every day that we have the opportunity to close a 10 years old issue 😉

marouan-boumchahate:
Thanks a lot for the feedback and the encouragement. It's super exciting that I was able to help close #602 as well

I've updated the PR just now:
1. Updated the description to `Fixes #3673, fixes #602`.
2. Removed `@skip_if_ci` from `test_timestamps_are_unique` in `kinto/core/storage/testing.py`
and bumped it to run 5 concurrent threads.

About the performance of `pg_advisory_xact_lock`. This latter operates entirely in RAM (PostgreSQL's shared memory) hash table and doesn't perform any I/O on the disk. A lock acquisition takes less than a microsecond. Also, because the lock is meant only for that specific `(parent_id, resource_name)`, writes to diff. collections or buckets will not block each other for sure. For details about proper use of these Locks, see [Section 9.26.8. Advisory Lock Functions](https://www.postgresql.org/docs/9.3/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS).

Lemme know if there's anything else you  wanna me to tweak!

